PR #70, opened 2024-02-01
[USER DATA] centralise directory for user data (eqsl, qsl)

abarrau:
explication done on #63


application/views/interface_assets/footer.php
```diff
@@ -2383,9 +2383,14 @@ className: 'buttons-clear',
 
 <script>
 function viewQsl(picture, callsign) {
+            <?php 
+            $CI =& get_instance();
+            $CI->load->model('Qsl_model');
```
HB9HIL, line 2388: There are several places in the `views` you've changed where you load a `model` in a `view`. We would like to stay in the classic MCV pattern which means you should load the models in the controller and give `$data` to the view 😊

HB9HIL (changes requested):

application/views/interface_assets/footer.php
abarrau: i will change this on next commit, but not on controller.

abarrau:
NOTE : 
i create a generic function file, this functions can be from different place (like the core function base_url())
i chose this way, because qsl path is used on the qso popup view, and this popup is called on different place (map, statistic, log, ...) so added this variable path on controller was very big word and not generic ... 
hope is ok for team

HB9HIL (approved):
works

int2001 (approved):
Checked code for security-issues - can't find one. good work.

Didn't check for function!

phl0 (approved):

PR #73, opened 2024-02-01
[eQSL] add stations_id to eqsl functions

dg0tm:
It seems that all eQSL functions are not aware of the station ID and therefore check all QSOs regardless of their
accounts and locations.

In a multi-user environment, if the label is the same, QSOs may be updated out of scope (club stations).


phl0 (approved):
Tested and works ok imho.

int2001 (approved):
works as designed

PR #74, opened 2024-02-01
[eQSL] only locations with a eQSL nickname should be displayed

dg0tm:
only locations with a eQSL nickname should be displayed in the dropdown under eQSL Import

### DIFF
--- a/application/controllers/Eqsl.php
+++ b/application/controllers/Eqsl.php
@@ -40,8 +40,7 @@ class eqsl extends CI_Controller {
 		$this->load->model('eqslmethods_model');
 		$eqsl_locations = $this->eqslmethods_model->all_of_user_with_eqsl_nick_defined();
 		if($eqsl_locations->num_rows() == 0) {
-			show_error("eQSL Nicknames in Station Profiles aren't defined");
-			exit;
+			$this->session->set_flashdata('error', 'eQSL Nicknames in Station Profiles aren\'t defined!');
 		}
 
 		ini_set('memory_limit', '-1');
--- a/application/views/eqsl/import.php
+++ b/application/views/eqsl/import.php
@@ -43,9 +43,7 @@
                     <?php } } ?>
                     </select>
 			</div>
-
-			<br><br>
-			
+			<hr class="divider">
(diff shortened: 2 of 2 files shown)


phl0:
Not sure about this one tbh. The download link to the eqsl.cc inbox obviously downloads the entire eQSL inbox anyway. So it probably contains more than just the QSOs on the profiles linked to eQSL nicknames?!

int2001:
Agree with @phl0 - it doesn't have any effect since eQSL isn't taking care of that.
Would only affect the UI, but not the logic. 
perhaps we should remove the station-chooser from eQSL completly and show the import only if there's at least one Station-location with eQSL-Nicknames set?

dg0tm:
please see #73

and I believe the downloaded inbox only contains QSLs for or from the selected nickname. We were always confused when QSOs updated out of scope...

int2001:
> and I believe the downloaded inbox only contains QSLs for or from the selected nickname. We were always confused when QSOs updated out of scope...

i just tried that. created a station_location with the same CALL but a different Nickname (fun fact: one that doesn't exist at eQSL). eQSL-Endpoint is downloading/fetching ALL eQSLs and doesn't take care of the Nick when fetching the eQSLs. The "Nickname" is passed to eQSL, but eQSL ignores it

HB9HIL:
error

There is missing an error handler if no station has an eQSL Nickname

HB9HIL (changes requested):
please add an error handler if no station has an eQSL Nickname

int2001 (approved):
works

HB9HIL:
> Okay, I took a first look at it and it is not triggered by this patch.
> 
> Should I patch this in another PR or with this one?

You cann add the error handler in this PR

dg0tm:
quick & dirty patch via flash message

i also add 2 lines for better knowledge off witch button belongs to which function at all

dg0tm:
eQSL-Nickname_not_defined

dg0tm:
eQSL-usability

by switching the order, it is much more intuitive

HB9HIL (approved):

PR #80, opened 2024-02-02
[Debug] Added information about when files were last updated, and links to update

AndreasK79:
image


phl0 (changes requested):
Works. Maybe we can make the date be displayed with custom date format?

AndreasK79:
@phl0 give it a spin now.

int2001 (approved):
nice.
perhaps in a further iteration return to debug-page?

PR #84, opened 2024-02-02
show flashdata if no station has eqsl nick

HB9HIL:
#74 was not complete. this adds the same error message to the Upload section.

eqsl


int2001 (approved):
works

PR #91, opened 2024-02-04
[Contesting] Fix for table qso count

AndreasK79:
This should take care of the count in the QSO table.


int2001 (approved):
working. nice teaming :)

PR #96, opened 2024-02-05
[LBA] Fixed mapping of selected QSOs

AndreasK79:
I forgot to fix mapping of selected QSOs when adding location to search.

@phl0 give this one a try.


phl0 (approved):
Fix confirmed.

PR #100, opened 2024-02-05
Make consequent use of SIG-References from station_profile

int2001:


phl0 (changes requested):
Please ingerate https://github.com/int2001/wavelog/pull/1. Rest is fine with me and works as expected.

phl0 (approved):
Now fine to merge.

PR #102, opened 2024-02-06
[LBA] Checkbox to turn on/off path lines in map

AndreasK79:
Added checkbox to disable path lines in map:
image
Right now, it does not look nice.


AndreasK79:
Thanks to @int2001, the checkbox is much better.
image

int2001 (approved):
works as designed. much easier to look at, when many qso's for map were selected

Issue #105, opened 2024-02-06
Inconsistency in handling of station locations 


205er:
Hi all,

I use station locations to separate my ham radio activities at different location. I have a station location for my home QTH, one for the club station and several locations for outdoor activities.
![Station location](https://github.com/wavelog/wavelog/assets/14996169/891270f9-ed33-4a8c-b6ac-26c66c57b0be)

I always activate the station location, whenever I change my physical location to benefit from the prefilled values in QSO logging tab "station".
Unfortunately, the activation of station locations does not have an impact on the QSO overview (_Logbook_ --> _Overview_)
The QSO that are listed in this overview do NOT belong to the active station location, but to location with ID 1. This is misleading (and in my eyes also wrong), because the active station is indicated at the top left, but it does not relate to the below shown information (QSO).
![Inconsistency](https://github.com/wavelog/wavelog/assets/14996169/654d6557-7c49-4373-bcb9-89f5db9e1cbf)

I would propose to change this behavior: Always output the QSO in the overview that are related to the active station.

 Thanks for your amazing work
 br
 Falko

phl0:
Well it is showing the QSOs belonging to the logbook not the location. So it displays all QSOs of all locations that are linked to the active station logbook.

phl0:
I just did some tests and cannot reproduce the error. Did you link the OE location to the station logbook named "Standard Log"?

phl0:
See my test case here. I created a station profile named "test", linked it to my active station logbook called "All" and logged a QSO with DJ7NT:

![Screenshot from 2024-02-06 17-21-14](https://github.com/wavelog/wavelog/assets/7112907/382e8f15-dd1e-4bdb-9b90-f8008b43d73c)

The advanced logbook shows the QSO of the active station location. And in Logbook overview I can see the QSO:

![Screenshot from 2024-02-06 17-21-24](https://github.com/wavelog/wavelog/assets/7112907/d9609259-9118-4bbf-ad96-2e091e367657)

Because the "test" location is linked to station logbook "All". Now if I remove the link the QSO disappears from the active logbook:

![Screenshot from 2024-02-06 17-24-10](https://github.com/wavelog/wavelog/assets/7112907/fd8271b4-dcb5-4e2d-a86e-8b25dc4f174d)

So I still think that your station location "OE" is not linked to station logbook "Standard Log".

phl0:
Added a change PR to remove station location from the listing as it is not what it is filtered for.

[...]

PR #107, opened 2024-02-06
Make clearer that overview filters on logbook rather than locations

phl0:
Addresses: https://github.com/wavelog/wavelog/issues/105 to make clear(er) that logbook overview only filters on station logbooks rather than station locations.


HB9HIL (changes requested):
[PR#1 against your branch](https://github.com/phl0/wavelog/pull/1)

telegram-cloud-photo-size-4-5913376836912529849-y

PR #109, opened 2024-02-07
[LBA] Added custom mapping icons

AndreasK79:
image

image


phl0 (approved):
Works and looks very nice. Good code to replace custom maps.

PR #110, opened 2024-02-07
Header refactoring

HB9HIL:
Refactored the header as it should look a bit nicer and will get an kind of "quickselect" for locations and logbooks later. 
Also there was an issue with the behaviour on mobile devices (tabs weren't closable if clicked agn).

There are still some issues but this is a first draft.


phl0 (approved):
Tested and working fine so far.

PR #112, opened 2024-02-07
[WIP] Migration for copying QSL images to userdata dir

phl0:
This only triggers if the new config option is enabled before the migration is run.

### DIFF
--- a/application/config/migration.php
+++ b/application/config/migration.php
@@ -22,7 +22,7 @@ $config['migration_enabled'] = TRUE;
 |
 */
 
-$config['migration_version'] = 179;
+$config['migration_version'] = 180;
 
 /*
 |--------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


abarrau:
Hello Florian, 

for this, i read your code and I would suggest a model where we "rename" the directory rather than copying all the files (which is more cumbersome in terms of processing). 

for example, 
```
$src = 'images/eqsl_card_images';
foreach (scandir($src) as $file) {
	if (!is_readable($src . '/' . $file)) continue;
	if ($file != '.' && $file != '..') {
		copy($src.'/'.$file, $userdata_dir.'/'.$eqsl_dir.'/'. $file);
	}
}
```
replace to (like something that) : 
```
if (!rename($src, $userdata_dir.'/'.$eqsl_dir)) { 
     log_message('error','ERROR during moving files from "'.$src.'" to "'.$userdata_dir.'/'.$eqsl_dir.'"');
 }
```

it's just a suggestion ;)
73

phl0:
> for this, i read your code and I would suggest a model where we "rename" the directory rather than copying all the files (which is more cumbersome in terms of processing).

Tbh I did "only" a copy intentionally to have a backup on FS in case something fails. :)

phl0:
Does not really make sense with userdata being disabled by default. Closing here for now.

PR #122, opened 2024-02-10
Tidy up a bit

HB9HIL:
Tried to fix as much as possible

`$CI` to `$this`
instead `load->model` in view serve the `$data` from controller

### DIFF
--- a/assets/js/sections/mode.js
+++ b/assets/js/sections/mode.js
@@ -10,6 +10,7 @@ $('.modetable').DataTable({
 		url: getDataTablesLanguageUrl(),
 	},
 	initComplete: function () {
+		var table = this.api();
 		this.api()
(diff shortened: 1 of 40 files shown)


assets/js/sections/mode.js
```diff
@@ -30,9 +31,9 @@ $('.modetable').DataTable({
 						select.append('<option value="' + d + '">' + d + '</option>');
 					});
 			});
+		table.columns.adjust().draw();
 	},
 });
-$($.fn.dataTable.tables(true)).DataTable().columns.adjust();
```
HB9HIL, line 35: error

This error appeared. Seems like a kind of race conditions problem, Nothing serious. `adjust().draw()` fixes it.

HB9HIL:
too much at once

PR #126, opened 2024-02-12
Label fence

int2001:
This one adds some Form-Validation for the Label-Menu.
All fields have to populated properly when adding or saving a label or paper-type.


HB9HIL (approved):
works

PR #131, opened 2024-02-13
[MENU] Locations Quickswitch

HB9HIL:
Introducing an option to show a "Quickswitch" for Station Locations in the Menu.

Can be activated in the User menu located in "Wavelog Preferences" (only for existing users)

quickswitch

Please test and check how I solved it. Was quite tricky to get the data in the header. Maybe also suggestions for improvement?


int2001:
Tested this one:

- i was able to chose a location even if the location doesn't belong to the active logbook (perhaps that bug was already there before - but should be fixed)
- Not really great UX. During test i "lost" the Menue several times, very "fiddly" with that dropdown/dropleft. just a personal opinion.
- I like: One is able to dis/enable the Feature via options.

phl0 (approved):
Works as expected. Really nice addition for grid activators :)

Screenshot from 2024-02-14 09-19-09

PR #132, opened 2024-02-14
[STATISTICS] Timeplotter Multilanguage Support

HB9HIL:


int2001 (approved):
works as expected.
One more View with MultiLang-Support

PR #133, opened 2024-02-14
Fix space in adif

int2001:
If you tried to upload an ADIF with a space (blank) in Filename it failed.
This fixes the issue.

thx to @dg9vh for poking me.


AndreasK79:
Working here too.

AndreasK79:
Tried again after finding a bug here. All working here now.

AndreasK79 (approved):

PR #134, opened 2024-02-14
Fix: Print SUBMODE if submode present, else MAINMODE at labels

int2001:


phl0 (approved):
Works as expected.

Screenshot from 2024-02-14 09-06-36

PR #135, opened 2024-02-14
Only use CQ zones <= 40

AndreasK79:


phl0 (approved):
Fix works as expected.

PR #138, opened 2024-02-14
Migrate hams.at to its native API

phl0:
Use hams.at's native API instead of the page scraper.


int2001 (approved):
works. noticed a few more fields. i like

PR #140, opened 2024-02-15
[QRA] mutualize function echoQrbCalcLink() on 1 file : qra librairy

abarrau:
Hi, 
propose this PR after speak with @HB9HIL 
- add parameter for remove icon/link on visitor page (because spawnQrbCalculator() not exist for visitor, js error) 
- tested on 4 places : logbook, visitor, previous page, popup with logbook

IMPORTANT INFO : the structure file of qra librairy was not good. 
4 functions was on the file, but not on the class, and they function have the same name of function on the qra class (but not the same utility ...)  ... very strange !

### DIFF
--- a/application/config/autoload.php
+++ b/application/config/autoload.php
@@ -52,7 +52,7 @@ $autoload['packages'] = array(APPPATH.'third_party');
 |	$autoload['libraries'] = array('database', 'session', 'xmlrpc');
 */
 
-$autoload['libraries'] = array('database', 'session', 'curl', 'OptionsLib', 'Paths', 'Frequency');
+$autoload['libraries'] = array('database', 'session', 'curl', 'OptionsLib', 'Paths', 'Frequency', 'Qra');
 
 
 /*
--- a/application/config/autoload.php
+++ b/application/config/autoload.php
@@ -52,7 +52,7 @@ $autoload['packages'] = array(APPPATH.'third_party');
 |	$autoload['libraries'] = array('database', 'session', 'xmlrpc');
 */
 
-$autoload['libraries'] = array('database', 'session', 'curl', 'OptionsLib', 'Paths', 'Frequency', 'Qra');
+$autoload['libraries'] = array('database', 'session', 'curl', 'OptionsLib', 'Paths', 'Frequency');
 
 
 /*
--- a/application/libraries/Qra.php
+++ b/application/libraries/Qra.php
@@ -113,6 +113,15 @@ class Qra {
 		
 		return array($lat * 180 / pi(), $lon * 180 / pi());
 	}
+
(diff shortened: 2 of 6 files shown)


phl0 (approved):
Tested and all locations where the lib is needed obviously work. Good idea!

AndreasK79 (approved):

AndreasK79:
Test working here as far as I can see :) Good work!

application/config/autoload.php
```diff
@@ -52,7 +52,7 @@
 |	$autoload['libraries'] = array('database', 'session', 'xmlrpc');
 */
 
-$autoload['libraries'] = array('database', 'session', 'curl', 'OptionsLib', 'Paths', 'Frequency');
+$autoload['libraries'] = array('database', 'session', 'curl', 'OptionsLib', 'Paths', 'Frequency', 'Qra');
```
int2001, line 55: Can u explain the reason for autoload?

int2001: All in all great work. i like everything which simplifies the codebase.

abarrau: i can load Qra library on each echo_table_col() function
but on this function (echo_table_col()), frequency is on autoload (ex : echo $ci->frequency->hz_to_mhz($row->COL_FREQ); for frequency case). 
so i thinks i better to have the same method; but if team prefer, i can load library on each function

int2001: there are 4 views, where the function is needed. A Solution can be: load it at every of the 4 ones.
you're absolutely right with frequency - it is also autoloaded (but not nescessary in my opinion).

i'm remarking this, because autoload means it is loaded everywhere at any time (on every page call). Yes it isn't huge, but where should we set the threshold?!

just thinking loud. leave it up to you, as the others already approved it ;)

abarrau: remove autoload and change with a load on each function, no pblm .. tell me and i modify , perhaps it's a better way

int2001: thank u so much.
perhaps you find a way to eliminate frequencies, too?
Didn't look where it is used, yet.

abarrau:
@int2001 i removed qra on autoload and add load on each function for qra and frequency
but not deleted frequency on autoload, because lot of place must be change for that ... on other PR ok ? 

and about my IMPORTANT NOTE on 1st msg ?

int2001:
> @int2001 i removed qra on autoload and add load on each function for qra and frequency but not deleted frequency on autoload, because lot of place must be change for that ... on other PR ok ?

yep. Thank u!

> and about my IMPORTANT NOTE on 1st msg ?

indeed strange. great for discovering that. i think the reason is: Historically grown. There are many things like this in the Codebase. Always happy if sbdy. picks those things up and make it more clean ;)

int2001 (approved):

PR #141, opened 2024-02-15
[LBA] Added user options for map layers

AndreasK79:
By popular demand:
image

### DIFF
--- a/application/controllers/Logbookadvanced.php
+++ b/application/controllers/Logbookadvanced.php
@@ -36,6 +36,14 @@ class Logbookadvanced extends CI_Controller {
 			$data['options'] = $userOptions[0]->option_value;
 		}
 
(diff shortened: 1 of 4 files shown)


abarrau:
Hello, 
i thinks that too yesterday, 
but can i propose an other way : save user option when user clic on checkbox map menu ?

phl0 (commented):
There is an issue with the gridlines at least. I checked that box but after mapping the gridlines are missing. Deactivating and activating them on the map works however.

Screenshot from 2024-02-15 14-08-40

Screenshot from 2024-02-15 14-08-50

phl0 (approved):
Working fine now.

AndreasK79:
> Hello, I think that too yesterday, but can i propose an other way : save user option when user click on checkbox map menu ?

Hi, That could be done, but I want to keep it under options. I also play with these layers, and I don't want to save the state after clicking on them.

PR #144, opened 2024-02-16
[LBA] Batch QSO edit

AndreasK79:
Mark qso, click edit. Only added support for a few columns so far:
image


int2001 (approved):
Tested. Works premium!

PR #145, opened 2024-02-16
[QSO View/Edit] States from Database

HB9HIL:
Thx to @phl0 who introduced the subdivisions table

With this PR we fetch available states from the database. 

Made Changes:

- QSO Edit shows states dependent on qso dxcc (only if available in database, only on load - not live if dxcc is changed)
- QuickLookup WAS Check fetches from DB
- QSO Logging US State selection fetches from DB*
- Station Location has now all States aswell, means you can set the location station fix in database for all available DXCC (not only US)


int2001 (approved):
nice nice. very nice ;)
smaller codebase, more flexibility. i like

PR #149, opened 2024-02-17
fix callbook state data

HB9HIL:
in addition to #145 

this fixes the callbook state data for all available data


int2001 (approved):
Tested this. Works great. Even updating out of LBA is now working!
Not 100% sure about the new Logic at qrz.php L100/L103.

HB9HIL:
@phl0 please check. just removed US dependency for state parameter

we need to merge this to test racecondition

HB9HIL:
> Tested this. Works great. Even updating out of LBA is now working!
> 
> Not 100% sure about the new Logic at qrz.php L100/L103.

I added in Line 95

phl0 (approved):
Works as expected now.

PR #151, opened 2024-02-18
Cache externals

int2001:


AndreasK79:
Tested working. Not sure if 10 minutes for SAT is good though. Perhaps @phl0 can pitch in.

int2001:
> Tested working. Not sure if 10 minutes for SAT is good though. Perhaps @phl0 can pitch in.

turning it down to 1min. and merging it. Only affects huge installations

PR #154, opened 2024-02-18
Slim base

int2001:
Outsourcing of often used functions to a lib.
Smaller CodeBase, more recycleable, more maintainable


AndreasK79 (approved):

phl0 (commented):
All award still working here.

PR #155, opened 2024-02-18
[Leaflet] Updated Leaflet library

AndreasK79:
Updated to latest Leaflet version.
Haven't seen any problems yet.

V1.3.4 -> V.1.9.4


int2001 (approved):
Tested a handful pages, where leaflet is used (not all).
Works without error

phl0:
Found no issues on the map pages.

PR #157, opened 2024-02-18
[LBA] Now deletes QSOs in batch, instead of looping through each

AndreasK79:
Before: looping through each id and deletes one QSO at a time. It is slow.

Now: deletes all QSOs in one query. It is very fast.


HB9HIL (approved):
works great :)

phl0:
Working fine here. Confirmed faster execution time.

PR #158, opened 2024-02-19
Fix plotter

HB9HIL:
fixes an issue where the map is broken if a qso does not contain a DXCC AND a Gridsquare


int2001 (approved):
a small change with huge (positive) impact. good find

PR #159, opened 2024-02-19
Deal with lower case IOTA references

phl0:
Fixes the screw up of the IOTA awards list if a lower case IOTA reference is in the log.


int2001 (approved):
good find. merge

PR #160, opened 2024-02-19
Restore distance tab layout

phl0:
Was broken in commit 604c5ba174cb76937cf2a6d80f266b38736bd796


int2001 (approved):
tested. does what is expected ;)

PR #167, opened 2024-02-23
[MAP] show qrz confirmation in green

abarrau:
Hi, 
propose this PR for show QRZ confirmation with green plot on map (dashboard, logbook and map lba). 

(this "multi" confirmation information was in 2 places on code, a unique function (on logbook model) must be better ?)


int2001 (changes requested):
GM Aurelien,
QRZ wasn't in there for a reason. The only counting methods are "LotW" and "QSL-Cards".

at some other views we're selecting the methods the user has chosen at his options. so if you really like to add QRZ to LBA, the thing should be refactored to make use of the user-options (Account -> Default-Values -> Default-QSL-Methods).

perhaps @AndreasK79 has an idea here.

abarrau:
Hello Joerg,

I don't understand : "The only counting methods are "LotW" and "QSL-Cards"." ; 
- QRZ is not use for defined if a qso is valid/confirmed ? 
- eqsl is used on lba, dashboard and logbook, so it should be exclu (if only lotw and qsl card are accepted)

thanks

AndreasK79:
Hi Aurelien.

For most awards, only QSL and LoTW counts. I agree that this should then be a user option to choose which confirmations should be counted for confirmed in the maps. To implement your solution for custom icons in LBA, I took the easy way, and just hardcoded the same way that was for the markers in the map on the dashboard.

abarrau:
ok, thanks all
i will change "count" (show on map) method for dashboard, logbook and lba.
i change the name on this PR ;)

int2001:
> ok, thanks all i will change "count" (show on map) method for dashboard, logbook and lba. i change the name on this PR ;)

it should be already implemented at dashboard and most of the awards.
The counting methods are living in the user-options. See "Account -> Default-Values -> Default-QSL-Methods"

abarrau:
so, i create a qso_is_confirmed() function, now confirmation plot on map (dashbord, logbook, lba) use the "qsl method" from user option.

int2001 (approved):

int2001:
Great. i like the new function

AndreasK79:
Tested working :)

AndreasK79 (approved):

HB9HIL (approved):

PR #170, opened 2024-02-23
fix lotw cert upload flashdata messages

HB9HIL:
the folder permission warning was shown after a cert upload, even if the folder is writable and the cert upload was successful


phl0 (approved):
Works as expected. Nice find!

PR #171, opened 2024-02-24
Qsl foldersize

HB9HIL:
The info banner on QSL/eQSL Card page shows the same foldersize for all users, no matter how many QSL cards they've uploaded. 

With this change we only show this info banner if the config option "userdata" is enabled. 

Also I moved the functions `folderSize($dir)` and `sizeFormat($dir)` to the `Genfunctions.php` Library (@phl0)

qsl


int2001 (approved):
Tested this. works.

PR #173, opened 2024-02-25
Maintenance mode

HB9HIL:
Added a "maintenance mode"

enabling is the same workflow as the developement mode. just adding a file `.maintenance` enables the maintenance mode.

Special thing is here that only Admins are allowed to login. This helps to test features and administrate an instance and avoid users to continue logging qso's

If a normal user is already logged in and the maintenance mode is enabled, the user won't get kicked out. If an admin want's to do that he has to restart the webserver.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,4 +20,5 @@ sync.sh
 *.p12
 *.swp
 .debug
+.maintenance
 .htaccess
--- a/application/controllers/User.php
+++ b/application/controllers/User.php
@@ -707,8 +707,13 @@ class User extends CI_Controller {
 				$this->input->set_cookie($cookie);
 				redirect('dashboard');
 			} else {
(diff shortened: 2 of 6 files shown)


int2001 (changes requested):
if a user is still logged in, while activating that new mode it has no effect to him.

I like the idea of this mode, but since many users leave their browser-window open overnight, it won't help here.

HB9HIL:
> if a user is still logged in, while activating that new mode it has no effect to him.
> 
> I like the idea of this mode, but since many users leave their browser-window open overnight, it won't help here.

Yes thats true. I'm not sure if it's a good way to do a "hard kickout". The actual mode only prevents relogin. As mentioned above if the admin want's to do a hard kickout of all logged in users a `systemctl restart apache2` does that trick aswell.

int2001 (approved):
works

PR #174, opened 2024-02-25
QSL Files Migration

HB9HIL:
Changes:

- moved the "Maintenance" view to the "Debug" view

Additions:

In the Debug view you now have the possibility to migrate the eQSL and QSL images from the old paths to the new userdata path fully automatic. It checks the corresponding user_id to each file and sort's all files in the `userdata` directory. 

Drafted at the moment to add some more "cases". For example if you already run a migration it should check if there are any files left which where are not migrated yet. 

migration


int2001 (approved):
**Didn't review the whole code** but tested some attacking scenarios. No one found during my tests.

PR #175, opened 2024-02-26
Station setup

AndreasK79:
First step at simplifying the setup for logbooks and location.

One page for setting up both logbook and location. The page is now more dynamic, so that every change does not reload page.


phl0 (approved):
Looks fine and works so far :)

int2001 (approved):
working here, too.
only thing missing (but that was a from before): Copy StationLocation.

phl0 (approved):
Tested and works flawlessly here.

PR #176, opened 2024-02-26
Refactor jcc

phl0:
6-digit JCC codes for Tokyo start with 10. So narrowing down the SQL argument.
Also remove redundant data (JCC array) from the code and generate ids from name array.


AndreasK79 (approved):

AndreasK79:
Tested working as before, both WAJA and JCC.

PR #181, opened 2024-02-28
Qso Reset

int2001:


HB9HIL (approved):

phl0:
Band RX is not (re)set to empty when switching from logging a SAT QSO to logging an HF QSO. Needs some investigation here ...

phl0 (approved):
Tested and it looks working ok on my end.

phl0:
Merged on behalf of @int2001

PR #183, opened 2024-02-29
Edit_Ajax Background

HB9HIL:
Just a small Frontend adjustment

Added another background layer to make it easier for the eyes to differe between the qso details modal and the edit ajax dialog

Without my change:

without

with the change

with


int2001 (approved):
-.-

PR #184, opened 2024-02-29
Harmonize keys and change to innoDB for QSO-Table

int2001:


AndreasK79 (approved):

AndreasK79:
Tested again with the new stuff. No errors, and worked as it should here.

PR #185, opened 2024-03-01
[STATION] fix default qslmsg empty when duplicate station

abarrau:
hello, 
fix copy eqsl_default_qslmsg when duplicate station

tested : 
- duplicate station with empty default qslmsg = ok
- duplicate station with value on default qslmsg = ok
- duplicate station with value on default qslmsg and change with new value = ok

### DIFF
--- a/application/controllers/Station.php
+++ b/application/controllers/Station.php
@@ -61,7 +61,6 @@ class Station extends CI_Controller
 				// [eQSL default msg] ADD to user options (option_type='eqsl_default_qslmsg'; option_name='key_station_id'; option_key=station_id; option_value=value) //
 				$eqsl_default_qslmsg = xss_clean($this->input->post('eqsl_default_qslmsg', true));
 				if (!empty(trim($eqsl_default_qslmsg))) {
-					$this->load->model('user_options_model');
 					$this->user_options_model->set_option('eqsl_default_qslmsg', 'key_station_id', array($station_id => $eqsl_default_qslmsg));
 				}
 			}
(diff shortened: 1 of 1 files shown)


application/controllers/Station.php
```diff
@@ -119,12 +119,23 @@ public function copy($id)
 			$data['my_station_profile']->station_profile_name = '';
 
 			if ($this->form_validation->run() == FALSE) {
+				// [eQSL default msg] GET from user options (same as edit()) //
+				$this->load->model('user_options_model');
```
HB9HIL, line 123: you don't need to load the model, as it is in the autoloader
```suggestion
```

application/controllers/Station.php
```diff
@@ -119,12 +119,23 @@ public function copy($id)
...
 				$this->load->view('station_profile/edit');
 				$this->load->view('interface_assets/footer');
 			} else {
-				$this->stations->add();
-
+				if (($station_id = $this->stations->add()) !== false) {
+					// [eQSL default msg] ADD to user options (same as create()) //
+					$eqsl_default_qslmsg = xss_clean($this->input->post('eqsl_default_qslmsg', true));
+					if (!empty(trim($eqsl_default_qslmsg))) {
+						$this->load->model('user_options_model');
```
HB9HIL, line 135: no need to load the model here

```suggestion
```

HB9HIL (commented):
imho we don't need the last function in this file. 
```
public function get_options()
	{
```
You wrote this 3 months ago, but as far I can see it is not used and just a redundant function of the main user_options function.

int2001:
Thank you, @abarrau - your Fix is working.
But we need that within the station_model, since it belongs to the station_profile.
I am already trying it to implement there.

application/controllers/Station.php
abarrau: i delete this load model on create, update and copy

abarrau:
> But we need that within the station_model, since it belongs to the station_profile.
> I am already trying it to implement there.

this function is used by js call on qso page for refresh qslmsg texteara, so this function is used

abarrau:
Tested : 
- create, update, copy station profil
- update msgqsl from qso page

abarrau:
> But we need that within the station_model, since it belongs to the station_profile.

@int2001, can i help about something ? 
you prefered i put the get_options() on station model ?

int2001:
i show you when finished. it's a little tricky and to complex to describe.

int2001:
you're right. whoever putted that into autoload (tbh: I am not a friend of autoload). Thanks for the finding, will fix

abarrau:
close this (reported on 175)

PR #186, opened 2024-03-01
[QSO] Multi POTA in QSO Logging

HB9HIL:
You can now choose multiple POTA's for the other Station when logging a QSO. 

will be undrafted after #181  is merged 

Own PR because has nothing to do with resetting QSO fields but is based on the branch of @int2001 

pota


phl0 (approved):
Code functionality works fine. Also tested ADIF exports. PR follows to adapt wording and to handle refs separately in Awards.

PR #188, opened 2024-03-03
QSO and footer stuff

HB9HIL:
Moved some functions from the footer to common.js to clean up the footer a little bit.

(renamed branch which closed #187)


HB9HIL:
Even it doesn't seem like there will be a merge conflict, it's probably better to merge #181 first

phl0 (approved):
Tested and could not find an error on the first glimpse.

PR #189, opened 2024-03-03
[Bands] Added RAC and JCC. And some future proofing

AndreasK79:
Added so that RAC and JCC can be checked/unchecked in the bands.

Since default will now be set in bandxuser, we do not need to specify awards. That way, we will hopefully avoid mistakes when new awards are added, as long as db column has a default value.

Needs to be merged after #184


int2001 (approved):
tested as addon to #184 - works and simplifies things. i like

PR #190, opened 2024-03-03
Fixed a PHP8.3 issue and removed BOM at Mig 139 (causes header_send)

int2001:
This one fixes:

- "Headers already sent" at a fresh install.
- Issue with "database_check" at install, when pressing install on PHP8.3


AndreasK79:
Installed without any trouble locally.

PR #191, opened 2024-03-03
Use default dxcache url if not set

AndreasK79:
Another step in making Wavelog easier to use for users.
If no dxcache url i set, use jo30.


int2001 (approved):
tested with fresh install and existing installation. works great

PR #193, opened 2024-03-03
[User] Insert default for map icons when creating a new user

AndreasK79:
Yet another thing making life easier when creating new users.
Setting default icons for map when a new user is created.

### DIFF
--- a/application/models/User_model.php
+++ b/application/models/User_model.php
@@ -203,6 +203,9 @@ class User_Model extends CI_Model {
 
 			$this->db->query("insert into bandxuser (bandid, userid) select bands.id, " . $insert_id . " from bands;");
 			$this->db->query("insert into paper_types (user_id,paper_name,metric,width,orientation,height) SELECT ".$insert_id.", paper_name, metric, width, orientation,height FROM paper_types where user_id = 0;");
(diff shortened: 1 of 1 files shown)


int2001:
got an error on a fresh user within Account-Settings, Map-Icon-Setup (within dropdown). PHP tells: `Undefined variable $user_map_gridsquare_show` at User-edit-View.

PS: Should consider setting "Default-QSL-Method" as well, what do you think?

abarrau:
Hello all, 
can i propose an other way ? 
we can allow icon map options directly on create user page. 
i speak about that with @HB9HIL few time ago. 
perhaps is better ? 
i can propose something this week if you are ok
73

AndreasK79:
Hi @abarrau. Sure, it was just to make it easier when creating a new user :)

You can add to "Create user" page. And when you do, set default values, so that the user do not have to choose anything in the page if the user does not want to.

phl0:
But it doesn't harm to have default values even if it can be chosen upon user creation. Just in case ...

abarrau:
hi, 
i this it's my previous proposal need lot of changes. 
the user_model add (and edit) function has a strange code  ... so a simple action, need to add lot of change ...
so lot of work (with risk) for "little" profit...

i propose we change nothing about that, and @AndreasK79 you can finish this PR ... sorry for delay

AndreasK79:
@abarrau thanks, I understand. Will update and merge later.

phl0 (approved):
Works as expected. Good stuff.

PR #195, opened 2024-03-04
[LBA] Now shows comment

AndreasK79:


int2001 (approved):
tested:
- LBA.
- LBA with comment col enabled.
- LBA with comment col disabled.

works great.

PR #199, opened 2024-03-05
Update lotw date

phl0:


int2001 (approved):
tested. can cnfm working

phl0:
@AndreasK79 is also fine with it. So merge time. Lets see where the fire ignites :D

PR #200, opened 2024-03-05
[Installer] Added DXCC. Installer now created logbook and location.

AndreasK79:
I updated install.sql to include DXCC.
I added city and DXCC to installer form.
Installer now inserts rows into logbook and location, then connects them.


int2001 (approved):
Whatever i try, i was not able to reproduce the 500 that was mentioned.
Everything works great. Huge improvement for fresh installations

int2001 (approved):

HB9HIL (approved):
Had no error in a new try. Seems okay now but we should definitly keep an eye on that.

PR #202, opened 2024-03-06
Catch Clublogerror if log was already sent

int2001:
In some special constellations it COULD be that the log was already sent to CL.
in that case clublog answers with a string like "checksum duplicate". This patch catches that message and marks the uploaded QSOs as sent


int2001:
Need to add for qrz.com, too.
This one is a good failsafe, but rootcause can be found in #199 
We need to make sure that **M**odified isn't set, if the same qsl-date (as already at DB) arrives from lotw.

Maybe you can add that here @phl0 as my guest?

phl0 (approved):
works as expected. Added a newline for browser output to prevent character mishmash.

PR #204, opened 2024-03-07
[MAP] remove map/custom reference

abarrau:
Hello, 
with LBA, the custom map was removed. 
so clean code with function used only with old custom map. 

Tested (non-regression) : 
- dashboard map : ok
- logbook map : ok
- visitor map : ok
- lba map (without influence) : ok

QUESTION : 
- the url : "https://%IP%/map" return a white page, perhaps change index() for returning to dashboard or return a 404 error ?


phl0 (approved):

AndreasK79:
@abarrau thanks for doing cleaning. I removed some stuff, but not all.
Add a return to dashboard, since we still have the controller.

abarrau:
> @abarrau thanks for doing cleaning. I removed some stuff, but not all (when I removed the custom map). Add a return to dashboard, since we still have the controller.

@AndreasK79 : it's done.

phl0 (approved):
Fine with me.

PR #206, opened 2024-03-07
Add user options to allow for showing workable passes only

phl0:
This allows that hams.at table to show which passes are workable or even only show the workable passes based on the grid which is set in the hams.at profile.

Screenshot from 2024-03-07 16-12-45

Screenshot from 2024-03-07 16-12-56


int2001 (approved):
i like. Helps a lot when working LEO/MEO

HB9HIL (approved):
works great

PR #207, opened 2024-03-08
Show (E)QSL recevice date rather than filename

phl0:
Replaced column filename with EQSL RX date. Would apply same to QSL if consent.


phl0:
Patch applied for QSL as well.

int2001 (approved):
tested both (eQSL and QSL). Much better than b4

PR #208, opened 2024-03-08
eQSL bug fix

HB9HIL:
fixes a bug where it can happen that there are multiple eqsl card files for one qso. Result is a buggy QSO details view and doubled data in database


int2001 (approved):
Checks (successful) done:
- Migration up and down without eQSL-Dupes
- Migration up with eQSL-Dupes
- Tried to enforce new dupes - no chance (as it should be)

phl0 (changes requested):
Corrected two minor things in https://github.com/HB9HIL/wavelog/pull/13. Rest ist working ok for me.

phl0 (approved):
Good to go now imho.

PR #210, opened 2024-03-08
Contest bearing

int2001:
Added Bearing-Lookup to contesting.

image

### DIFF
--- a/application/views/contesting/index.php
+++ b/application/views/contesting/index.php
@@ -103,7 +103,8 @@
                             <div class="mb-3 col-md-3">
                                 <label for="callsign"><?php echo lang('gen_hamradio_callsign'); ?></label>
                                 <input type="text" class="form-control form-control-sm" id="callsign" name="callsign" required>
-                                <small id="callsign_info" class="badge text-bg-danger"></small>
+                                <small id="callsign_info" class="badge text-bg-danger"></small><br/>
+                                <small id="bearing_info" class="form-text text-muted"></small><br/>
                             </div>
 
                             <div class="mb-3 col-md-1">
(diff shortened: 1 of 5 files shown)


phl0 (changes requested):
Bearing is not cleared upon reset / logging the QSO. Also not updated when logging a new station.

application/views/contesting/index.php
```diff
@@ -103,7 +103,8 @@
                             <div class="mb-3 col-md-3">
                                 <label for="callsign"><?php echo lang('gen_hamradio_callsign'); ?></label>
                                 <input type="text" class="form-control form-control-sm" id="callsign" name="callsign" required>
-                                <small id="callsign_info" class="badge text-bg-danger"></small>
+                                <small id="callsign_info" class="badge text-bg-danger"></small><br/>
```
HB9HIL, line 106: just remote the `<br>`. everything else is great :) thx for that one! 

```suggestion
                                <small id="callsign_info" class="badge text-bg-danger"></small>
```

HB9HIL (approved):

application/views/contesting/index.php
HB9HIL: rr was able to reproduce that. sorry

phl0 (approved):
Works.

application/views/contesting/index.php
int2001: as chatted, @HB9HIL - if you have an idea how to fix that (and if its important enough), pse open new PR.
I tested with Safari, FF and Chrome (wothout BR) and the result was always bearing direct next to Warning.

As best-effort i'll merge this.

Issue #211, opened 2024-03-09
Problem with database password with single quote


reinerh:
**Describe the bug**
I have just installed wavelog and used a generated password for the database.
After finishing the installation, wavelog was only returning a blank page and I saw an
500 HTTP error in the access log.
After some debugging I noticed that `application/config/database.php` has an invalid string, as the generated password contained a single quote character, which broke the `password` setting.

**Version Information**
We need some information about your versions. You can get all Information in the Admin > Debug Menu
- OS:               Debian 12
- PHP:              8.2
- MySQL/MariaDB:    10.11
- Wavelog:          1.2 / git

**Steps to Reproduce**
Installed wavelog with a database password that contains a single quote.

**Expected Behavior**
Either a password with known problematic characters is rejected during installation, or problematic characters like `'`, `\` etc. are escaped.

HB9HIL:
Issue with single quotes will be fixed. other special chars shouldn't be a problem. But we can't escape them because the DB Password is written as plain text to the database.php.

PR #212, opened 2024-03-09
[Installer] DB Password

HB9HIL:
This is a fix for #211  

Changed in config.php and database.php (and samples) all single quotes to double quotes (for consistency). Also added a js warning which does not allow double quotes in db_password. 

We cannot escape it since the db password is written plain to the database.php.


int2001 (approved):
tested. works

PR #220, opened 2024-03-17
[LABELS] Add option to print reference

HB9HIL:
You have the option to print available references on the labels.


int2001 (changes requested):
Works fine out of label-menu,
but pse add checkbox for Reference, when printing labels out of LBA too.

AndreasK79 (approved):

PR #221, opened 2024-03-18
Stationsetup update

AndreasK79:
Ajaxified logbook setup too.


int2001 (approved):
i like. Much more "straight" than the old one.

Maybe disabling the buttons, while AJAX is working, can be a further improvement

int2001 (approved):
Yeah. Disabling of Buttons (while ajax is working) works fine

phl0 (approved):
Tested and works as designed :)

PR #222, opened 2024-03-20
Added Option to add ADIF-Contest to Contest

int2001:
* If ADIF already contains "contest_id" and "No Contest" is selected, the QSOs will be treated as defined in ADIF.
* If (an active) contest is selected, the QSOs will be added to that contest, regardless what contest is setted or not setted at ADIF

### DIFF
--- a/application/controllers/Adif.php
+++ b/application/controllers/Adif.php
@@ -139,6 +139,9 @@ class adif extends CI_Controller {
 	}
 
 	public function index() {
+		$this->load->model('contesting_model');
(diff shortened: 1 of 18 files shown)


AndreasK79:
Approved. 
1. Tested with setting contest when importing -> ok
2. Tested with contest set in adif, then importing -> override ok
3. Tested with contest set in adif, selecting nothing -> fetches from file.
Go for merge.

AndreasK79 (approved):

PR #228, opened 2024-03-22
Move completly to InnoDB

int2001:
Let's create everything with innodb instead of myISAM.

Was less than i expected.


AndreasK79:
No errors detected when trying this pr.

phl0 (approved):
Upgrading existing installation worked.

phl0:
Fresh install also worked. So all fine. Go for merge.

PR #229, opened 2024-03-22
References to QSO tab

phl0:
We had several requests to bring some of the fields in general tab to the front tab (i.e. QSO tab). So this brings a little flexibility by adding switches to choose which refs should be shown on QSO tab:

Screenshot from 2024-03-22 08-35-27

### DIFF
--- a/application/controllers/Qso.php
+++ b/application/controllers/Qso.php
@@ -40,6 +40,13 @@ class QSO extends CI_Controller {
 			$data['user_iota_to_qso_tab'] = 0;
 		}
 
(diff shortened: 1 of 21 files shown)


int2001 (approved):
works. great to let the user customize his QSO-Pane!

Tested:
* Fresh setting (without setting anything)
* Fresh Logging (without setting anything)
* Setting of IOTA / POTA and DOK
* Logging with that setting

phl0:
Language also ok? I was not sure about the wording of the setting.

int2001:
oh. language in english is okay. it's missing for german (and i think all other langs)

PR #230, opened 2024-03-22
Added a new Type of Map

int2001:
tnx to @AndreasK79 - that was great teamwork.


phl0 (approved):
It works :)

image

Very nice feature.

PR #232, opened 2024-03-22
Selfupdate

int2001:
Super-Simple Selfupdater. PSE chk carefully!

can be invoked by pressing "Update" at git-section within debug-menue.

stashes everything, pulls HEAD of current branch and pops things afterwards.

diff best viewed with "ignore whitespaces"


AndreasK79 (approved):

AndreasK79:
Tested working in my test environment, and with cloning and adding commits!

HB9HIL (approved):
I want to add some more frontend and a public function for a version check cron. This is a great base for that :) thx

Hint: If there are already saved stashes but in the first `git stash push` there were no local changes, it will pop an old stash. This can cause issues. We need to find a solution for this (rare) case. Even with that I think we can merge this. Good job :)

PR #236, opened 2024-03-24
Updater frontend

HB9HIL:
a bit more fancy frontend for the selfupdater

image


int2001 (changes requested):
As already mentioned:
Great work - much more fancy than before.
I like the UI.

But pse enable the update-button even if the instance-admin is NOT on master. Currently it only works if there's a new tag at master-release.

second one (minor): pse add a http-header to the output of wavelog_version.

application/views/debug/index.php
```diff
@@ -461,3 +466,7 @@
     </div>
 
 </div>
+
+<script>
+    var local_branch = '<?php echo $branch; ?>';
```
int2001, line 471: this one fails / is empty when using a version which wasn't installed via git.

int2001 (changes requested):

application/views/debug/index.php
HB9HIL: catched

int2001 (approved):

PR #240, opened 2024-03-27
Fix missing columns in logsearch

DJ3CE:
(See also Cloudlog PR 3019)

This PR/commit adds (user-selected) columns to the search_result_ajax, which were previously missing and thus led to errors if the user had selected any of them in her/his user-profile.

Additionally, it structures the code a bit more in the fashion used at other places in the code (functions for column headers and column rows), to make maintenance a bit easier.

### DIFF
--- a/application/views/search/search_result_ajax.php
+++ b/application/views/search/search_result_ajax.php
@@ -1,3 +1,86 @@
+<?php
+function echo_table_header_col($name) {
+	switch($name) {
+		case 'Mode':      echo lang('gen_hamradio_mode'); break;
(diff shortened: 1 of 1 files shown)


int2001 (approved):
woah!
i like it. tested it, it's working fine here. @AndreasK79 - may be you can check, too, before we merge?

AndreasK79:
As far as I can tell, it's working! @DJ3CE thanks for the contribution. and it's much easier to maintain the file now.

AndreasK79 (approved):

PR #241, opened 2024-03-27
Contest copy to any

DJ3CE:
In contest-logging, Cloudlog featured copying the received exchange into the (rather specific) DARC DOK field.

This PR brings this valuable approach to a broader set of contests by implementing 'copy exchange to'

* age,
* name,
* US-State,
* power used by the other station ('RX_PWR'),
* locator

which could easily be enhanced for other needs (I just included those, I know about).

Some notes:

*    in models/Logbook_model.php, an example for a more sophisticated copy is included
    some fields never appear in Cloudlog, like 'age'. But why not fill it - it's in the database. :)
* The select-field gets disabled, if no (applicable) exchange-type is selected

See Cloudlog PR as well.


int2001:
Works so far. But:
Should we implement a minimal check for the fields? If one accidently sets "copy to grid" and enters a number (or not a grid), some other things will break.

The QSO-Page for example checks the Grid for validity and throws an error to the user if grid isn't properly provided. This is important, because distance-calculations and so on rely on the value.

Same _could_ be at State. as far as i remember @phl0 created an enumeration for valid states.

DJ3CE:
I see two options:

* Leave out locator, it's kind of duplicate anyway...
* Server-side check before populating locator-field

I'd abstain from client-side s.th. to stay out of the way - but that's just mho. The feature's main intention is second-level qso-data enrichment...

int2001:
i really like the idea of the enrichment.
i'm with you - should be done on serverside.

That check will complicate saving of QSOs out of Contest a bit, since there's currently no messaging/rejection implemented for that.

"Normal QSO" uses the index-function of QSO-Controller, where rudimentary-form-checks are already implemented. The Contest makes use of "saveqso". The checks are missing there.

Maybe you can add them there.

PS: While reading the source i recognized you put the handling for those things into the model. That's ok, but could you pse add some sanitation (xss_clean and so on) to it, when collecting the post-params?

DJ3CE:
I'm happy to help out improving the structure of the code, so if you have a long term plan on the purpose of controller and model, let me know and I'll implement it that way. As you can see, I just plugged in the code where the DOK handling was done.

Is a basic regex-check for US-state (like `[A-Z]{2}`) and locator (like `[A-R]{2}[0-9]{2}([A-Xa-x]{2}([0-9]{2}?))?` (4-8 letters/digits)) sufficient for you?

(*) Regex in the correct slang then of course... ;)

int2001:
it is. have a look at L237 of libraries/Qra.php for Grid.
Regarding State... @phl0 is the ultimative expert here. Not sure if a single `[A-Z]{2}` is enough.

HB9HIL:
Regarding US state a Regex `[A-Z]{2}` could be enough for the moment. On the other hand you could check it against `primary_subdivisions` table.

application/models/Logbook_model.php
```diff
@@ -157,6 +163,49 @@ function create_qso() {
       $darc_dok = $this->input->post('darc_dok');
     }
 
+    $qso_locator = strtoupper(trim(xss_clean($this->input->post('locator')) ?? ''));
+    $qso_name = trim(xss_clean($this->input->post('name')));
+    $qso_age = null;
+    $qso_usa_state = $this->input->post('input_state_edit') == null ? '' : trim(xss_clean($this->input->post('input_state_edit'));
```
AndreasK79, line 169: @DJ3CE Missing a ) at the end of the line.

AndreasK79:
I would prefer not to limit state. If a user logs anything longer, it will log nothing in the database, and no message to the user.

Should this

>    if($this->input->post('copyexchangetodok')) {
>       $darc_dok = $this->input->post('exch_rcvd');
>     } else {
>       $darc_dok = $this->input->post('darc_dok');
>     }

be changed to

> $darc_dok = $this->input->post('darc_dok');

in logbook_model?

Since copyexchangetodok is not used, should we get rid of it all places?

And this in contesting_model

> 'copytodok'             => $this->input->post('copyexchangetodok', true) == "" ? 0 : xss_clean($this->input->post('copyexchangetodok', true)),


should probably be changed to copyexchangeto.

@int2001 would like your input here too.

int2001:
exchangeto is more generic, indeed @AndreasK79 

What about putting that logic to the contest-MVC, since it is really only for contesting.
You'll be able to give the user instantly feedback with the CI-Validator-Rules, and you can prepare the data directly for this case.

So moving the whole contest-thingy to contest_controller and contest_model is a good idea. helps to clean up things from history ;)

application/models/Logbook_model.php
DJ3CE: :( Sorry for that.

DJ3CE:
Thanks for the discussion. Trying to summarize:

1.) `copytodok`: The db-field in the session-table is named `copytodok`. As I'd like to avoid db-changes, there's a break in naming things. Other places yes, getting rid of `...dok`.

2.) 'To validate or not to validate': As already stated, this is thought to be 'second-level convenience data-handling'. I wouldn't even raise a warning for data not being transferred to additional fields. Would a note to the user be helpful here? Like "Copy exchange to ... if applicable"? "Don't distract the op" would be my standpoint. I like locator-validating, I'm neutral to state-validating and don't care otherwise. Thinking about states it would be good to exclude the (non-)'state' "DX" (and maybe "NM" as well?) maybe - opinions?

3.) structural decisions: Had a look into that and I guess while I'm totally into the idea of restructuring code, I'd prefer to have a seperate PR for that. Would moving the contest specific code to the `saveQso()`-function (in Qso-controller) be a (good) compromise at this point?

AndreasK79:
@DJ3CE looks better now :)

1. Could always be solved with a migration, but it's ok for me.
2. Yeah, I agree with you that the focus should be on operating and logging QSOs. Could just have a notice of some kind... As for state, I see you excluded DX. That's ok. NM could be New Mexico, so it should not be filtered. Also the regex should be more relaxed. Something like this ^[a-zA-Z]*. Other countries could have more than 2 characters in state.
3. Refactor in a separate PR is totally ok.

DJ3CE:
> looks better now :)

That's always the case, with more than two eyes and one brain looking at code and offering insights and thoughts - I appreciate that :+1: .

Regarding 2.: I added a title to the select-field and would be grateful for a thought on the chosen wording.

I have added the changes as another commit but would prefer to squash all into one commit for merging.

application/models/Logbook_model.php
```diff
@@ -151,12 +157,50 @@ function create_qso() {
       $clean_county_input = null;
     }
 
-    if($this->input->post('copyexchangetodok')) {
-      $darc_dok = $this->input->post('exch_rcvd');
-    } else {
-      $darc_dok = $this->input->post('darc_dok');
+    $darc_dok = trim(xss_clean($this->input->post('darc_dok')))
```
phl0, line 160: The trailing semicolon is missing. Leading to a PHP error.
Fix in: https://github.com/DJ3CE/wavelog/pull/1

phl0 (changes requested):

application/models/Logbook_model.php
DJ3CE: Please accept my apologies - both the missing bracket and this 'incident' happened, because I did retype the changes between my 'pushing machine' and the 'testing machine'...

phl0: No worries. Tnx for fix. Now on with testing here.

phl0:
The basic code works. So go for merge from me.
But maybe we can make the UI a little more tidy like this for example?

Screenshot from 2024-04-03 14-19-29

DJ3CE:
I like the idea. Label and select-box on one line or like your screenshot label-above-select-box?

Would the `general_word`- and `gen_hamradio`-translations (in `general_words_lang.php`) be appropriate for use of the DB-fields here? I see not translations for 'RX Power' (Power, the other station was using) and 'age' currently. Suggestions for good keys besides `general_word_age` and `gen_hamradio_receive_power`(*)?

(*) I think the wording could be misleading, but that's what the adif-standard uses...

HB9HIL:
> I like the idea. Label and select-box on one line or like your screenshot label-above-select-box?
> 
> Would the `general_word`- and `gen_hamradio`-translations (in `general_words_lang.php`) be appropriate for use of the DB-fields here? I see not translations for 'RX Power' (Power, the other station was using) and 'age' currently. Suggestions for good keys besides `general_word_age` and `gen_hamradio_receive_power`(*)?
> 
> (*) I think the wording could be misleading, but that's what the adif-standard uses...

@DJ3CE  to add translations more easy and you work on a Linux machine you can use some `sed` magic

For example:
```
sed -i "78i\$lang['options_bands'] = \"Bands\";" /var/www/wavelog/application/language/*/*options_lang.php
```
`78` is the line where to insert.

DJ3CE:
Nice command-line, thanks. My goto-command usually is sth like `find . -iname 'options_lang.php' -exec sed -i -e "/'options_bands'/ {p; s/_bands/_power/; s/= \".*\"/= \"Power\"/; }" {} \;`. Didn't know, that globs work with sed that well :).

phl0:
> I like the idea. Label and select-box on one line or like your screenshot label-above-select-box?

One line is even better. Saves some vertical space for the table below.

> Would the `general_word`- and `gen_hamradio`-translations (in `general_words_lang.php`) be appropriate for use of the DB-fields here? I see not translations for 'RX Power' (Power, the other station was using) and 'age' currently. Suggestions for good keys besides `general_word_age` and `gen_hamradio_receive_power`(*)?
> 
> (*) I think the wording could be misleading, but that's what the adif-standard uses...

Yepp use the existing translations where applicable and create new for the ones you need.

phl0 (approved):
Layout looks fine to me now.

PR #246, opened 2024-03-28
Tweaks to gui and spinner on update dxcc button

AndreasK79:
Added spinner to update button.
image

Used hx-get so that result will be placed below buttons on 
* Check QSOs missing DXCC data
* Re-check all QSOs in logbook - I removed printing of every QSO that is updated. The id from the table is listed, and is of very little use for the user. Only total at the end is displayed. I also limited this so that only the logged in users QSOs will be updated. Not 100% sure of this one though.
* Check QSOs missing continent data

### DIFF
--- a/application/controllers/Update.php
+++ b/application/controllers/Update.php
@@ -11,7 +11,7 @@ class Update extends CI_Controller {
     function __construct()
 	{
 		parent::__construct();
-		
+
 		if (ENVIRONMENT == 'maintenance' && $this->session->userdata('user_id') == '') {
             echo "Maintenance Mode is active. Try again later.\n";
 			redirect('user/login');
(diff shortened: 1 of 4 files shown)


HB9HIL (approved):

int2001 (changes requested):
i like the UI, but if have questions on "Check QSOs for missing DXCC-Data"

The "old" check_missing_dxcc was available to anyone (not really good) and updated every QSOs with missing DXCC instance-wide. It was (and is) reachable via direct link or the Admin-pages.

The new one wants a session (which is good, because the function takes its time) but it only updates the QSOs of the Admin and not longer of any other users. Other Users can't find this item, because it isn't linked. Cron wont work also, because there's no session.

maybe i'm wrong, but this is a bit confusing.

imho there are two options to solve:
a) Check for (user-)session at the controller and show the item to the normal user somewhere on the page, so users can update their own QSOs
b) Check for (admin-)session at the controller, remove the session-join at Logbook_model, so only admin can update all QSOs instance-wide

int2001 (approved):

PR #250, opened 2024-04-06
[Maps] Moved legend/info css from view to general.css

AndreasK79:
The same css was duplicated in all maps for awards. Removed the css from the views and added to general.css instead.

Added darkmode for legend/info too:
image


phl0 (approved):
Works as expected. Like the dark legend!

PR #253, opened 2024-04-07
[Maps] Tweaked legend map some more

AndreasK79:
Further tweaked legend and popup for map:
image
And for award maps:
image


int2001 (approved):

phl0 (approved):
Like the updated dark layout.

PR #254, opened 2024-04-08
Fix activated grids re orbit

phl0:
Add filter for satellite and orbit type to activated gridsquares.


int2001 (approved):
tnx for fixing. works

PR #257, opened 2024-04-09
[BUGFIX] Fixed Advanced Search

HB9HIL:
fixed a bug in Advanced Search if Frequency is chosen as column in the results. 

We need to use `get_instance()` in this case as the class is not defined in a view. We would need to rebuild the search in the MVC pattern to use `$this`.


AndreasK79:
Confirmed working now :)

AndreasK79 (approved):

PR #259, opened 2024-04-09
Fixes css for links when popup is initiated from award map

AndreasK79:
This only happens when you click on the award map to show QSOs.
The leaflet css overrides, but by changing to div, it will show correctly.

Before:
image
After:
image


phl0 (approved):
Fix confirmed.

PR #260, opened 2024-04-09
Add orbit type to LBA

phl0:
This adds a filter option for orbit type to LBA. Orbit is also printed on map popups and result table.


AndreasK79:
Works in my testes here. Nice work.

AndreasK79 (approved):

PR #261, opened 2024-04-10
Public export map

AndreasK79:
"Wavelog-url"/index.php/visitor/exportmap/?slug=slugname&showgrid=false&qsocount=250&showlines=true&showcq=true

image


phl0 (approved):
Works as expected :) Very nice feature.

int2001 (approved):
works.

PR #264, opened 2024-04-12
[Exportmap] Added options

AndreasK79:
wavelogurl/visitor/exportmap/slugname

Options exist per slug basis, and can be found in stationsetup.


int2001 (approved):
From the technical aspect this works and looks very nice.

PR #265, opened 2024-04-12
[Distances worked] Added orbit as a selection

AndreasK79:


phl0 (approved):
Works as expected. Nice addition.

PR #268, opened 2024-04-12
[DXCC Award] Added orbit and sat to the selection

AndreasK79:
Added sat and orbit to selection in the GUI.


int2001 (changes requested):
All in All great feature.

Things to be completed (so far):
* When clicking to SAT, SAT and orbit appears. When going back to "All" (which is my default) they won't disappear (Notice: At first loading with "All" there's no Orbit/SAT)
* displayContactsOnMap is also called from waja, helvetia, iota, cq and rac - since u added sat and orbit in the middle of the params. those are breaking
* same for displayContacts

phl0:
>     * When clicking to SAT, SAT and orbit appears. When going back to "All" (which is my default) they won't disappear (Notice: At first loading with "All" there's no Orbit/SAT)

Fixed.

HB9HIL (approved):
good from my side

PR #273, opened 2024-04-14
some installer fixes...

dg0tm:


int2001 (approved):
Good find. Tnx

@HB9HIL since the installer is ur playground pse approve as well

HB9HIL (approved):
thx for that :)

PR #276, opened 2024-04-14
[Exportmap] Added lastqso

AndreasK79:
@int2001 now added exportmap/#slug#/lastqso


int2001 (approved):
works as designed.
well done. Reduces Traffic

PR #277, opened 2024-04-14
Docker readiness

int2001:
This one provides Docker-Readiness.

In Detail:
- Store config-Files somewhere else, if Environment is docker.
- Store things there where they belong if classic LAMP


HB9HIL (approved):
does not break anything in LAMP stack :)

PR #279, opened 2024-04-14
[Hotfix] Fix displayContacts on dashboard

phl0:


int2001:
Should do the job now. Can't find any more issues

PR #284, opened 2024-04-15
API grid cnfm

phl0:
Upgrades the API to also ask for confirmation state for worked grids. Example:

```
curl -X POST -H "Content-Type: application/json" -d '{"logbook_public_slug":"YOURSLUGHERE", "key":"YOURAPIKEYHERE", "grid":"EM13" , "band":"SAT", "cnfm":"lotw"}' https://URL_TO_CLOUDLOG/index.php/api/logbook_check_grid
```

### DIFF
--- a/application/controllers/Api.php
+++ b/application/controllers/Api.php
@@ -364,6 +364,13 @@ class API extends CI_Controller {
 				$band = null;
 			}
 
+			// If $obj['cnfm'] exists
+			if(isset($obj['cnfm'])) {
+				$cnfm = $obj['cnfm'];
+			} else {
+				$cnfm = null;
+			}
+
 			$this->load->model('logbooks_model');
 
 			if($this->logbooks_model->public_slug_exists($logbook_slug)) {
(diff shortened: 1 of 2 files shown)


application/controllers/Api.php
```diff
@@ -388,15 +395,25 @@ function logbook_check_grid() {
...
+				} else if ($cnfm == null) {
 					echo json_encode(['gridsquare' => strtoupper($grid), 'result' => 'Found']);
 				} else {
-					echo json_encode(['gridsquare' => strtoupper($grid), 'result' => 'Not Found']);
+					$arr = [];
+					foreach($query->result() as $line) {
+						$arr[] = $line->gridorcnfm;
+					}
+					if (in_array('Y', $arr)) {
+						print_r($arr);
```
int2001, line 410: Qu'est que c'est?

int2001 (changes requested):
Pse remove debugging(?) code-line.
Rest works

application/controllers/Api.php
phl0: f*ckup f*xed

PR #286, opened 2024-04-16
Fix gridchecker to be compliant with recent changes

phl0:
After changes in https://github.com/wavelog/wavelog/pull/284 we need to fix some internal function calls.


int2001 (approved):
well done. works

PR #287, opened 2024-04-16
Ci pipeline

int2001:
This one adds:
* Dockerfile (for building from scratch)
* Github-Action which fires on new release or on demand

Since there are small issues, this one can only be properly tested if merged to dev (and master afterwards).

Risks:
* Pipeline won't work (doesn't hurt)


phl0 (approved):
Just by reading the code looks ok.

PR #288, opened 2024-04-16
Use current band to check if grid worked

phl0:
Lookups of grid for grid and partial are considering different bands.
gridsquare input uses current band whereas partial uses the user's default band. This is an attempt to harmonize this.


int2001 (approved):
resolves an old bug. tnx

PR #299, opened 2024-04-19
Fix broken QSO-Detail-Links

int2001:
DisplayContactsOnMap were broken, since Orbit was added.
This one should fix it.


AndreasK79 (approved):

AndreasK79:
Nice catch. It's in use in too many places ;)

PR #300, opened 2024-04-19
Make DXCC Award also show a map of not-worked DXCC

phl0:
The DXCC Award map only calculated the not worked DXCC right if either confirmed or worked was chosen as well. Generating a map of only the not-worked DXCC was right but legend didn't count number of not-worked DXCC correctly and shows all (i.e. 340) instead:

Screenshot from 2024-04-19 23-37-27

JS code was corrected so that also only number of not-worked DXCC is calculated correctly:

Screenshot from 2024-04-19 23-37-38

### DIFF
--- a/assets/js/sections/dxccmap.js
+++ b/assets/js/sections/dxccmap.js
@@ -81,7 +81,7 @@ function load_dxcc_map2(data, worked, confirmed, notworked) {
         }
     ).addTo(map);
 
-    var notworkedcount = data.length;
+    var notworkedcount = 0;
     var confirmedcount = 0;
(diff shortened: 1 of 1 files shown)


AndreasK79 (approved):

AndreasK79:
@phl0 nice fix. Think the other maps need this fix too :)

phl0:
> @phl0 nice fix. Think the other maps need this fix too :)

Yepp. But those habe another issue as well. If you choose worked only for CQ for example I get no results at all. Will look into that in another PR.

DK9JC:
tested and works. 
thank you for the fix.

PR #301, opened 2024-04-20
[Awards] Added WAB

AndreasK79:


phl0 (approved):
Tested and works fine. If I only had the data ... :D

HB9HIL (approved):
ressource heavy but works :)  nice job

int2001 (approved):
works

PR #302, opened 2024-04-20
Markdown

HB9HIL:
Markdown support and conversion in version dialog


int2001 (approved):
looks much better now

PR #304, opened 2024-04-21
Clublog Ignore Switch

HB9HIL:
Due to the possibility to add a callsign, you do not own on Clublog, there needs to be a Ignore switch to turn off Clublog Upload for this particular location/callsign


AndreasK79:
Looks good. Works for creating new/editing locations. Everything is set in db. No errors.
Have not tested the Club Log upload, but it should work when looking at the code.

AndreasK79 (approved):

phl0 (changes requested):
Proposed some changes in https://github.com/HB9HIL/wavelog/pull/17.

phl0 (approved):
fine with me now :)

PR #305, opened 2024-04-22
Add tabindex to QSO entry pane

phl0:
Add tabindex for QSO entry and make it a useful order of tabs.


int2001 (approved):
works. good find to skip annoying field which are rarely needed

AndreasK79 (approved):

PR #306, opened 2024-04-22
[CRON] Cron Manager

HB9HIL:
First draft of a Cron Manager.

This Cron approach redefines how cronjobs are handled in Wavelog. The idea is to replace all "single" jobs as we describe them currently in the [Wiki](https://github.com/wavelog/wavelog/wiki/Recommended-Cron-Jobs) through a single cronjobs which runs EVERY MINUTE (This is necessary for the included library to work properly. 

I created a database table which contains various information about the cronjobs and which is used by the [cron-expression library ](https://github.com/poliander/cron) to analyse the cron expression and runs the jobs based on this expression. 

Why should we do that:

1. This gives us developers more flexibility to work on the functions which are currently used by the cronjobs. We can edit, add or remove such functions, without killing a remote instance. 
2. Admin can manage the intervalls of the "cronjobs" by theirself directly in the WebUI without necessary SSH access to the  server. 
3. Especially for Docker Installations this is an much easier approach then bashing into a docker machine or mounting a crontab file in a volume. 

This is still under construction. There is some work to do in the frontend and the possibility to edit the expressions. But the cronmanager is already working if you replace all existing cronjobs with the cronjob described in the view:

```
* * * * * curl --silent <URL to Wavelog>/index.php/cron/run &>/dev/null
```


HB9HIL:
Ready 2 Review in a first manner. Excited to get some feedback

PR #307, opened 2024-04-23
Removed some unused files and refactored some functions

AndreasK79:
Removed thirdpartylogin controller and model.
Removed migrationversion model. We have a debug_ model, so moved the single function there.
Moved count_all_qso and calls_without_station_id from logbook_model to debug_model.

The reason for moving function from logbook_model is that it is too big, so better place them in debug_model in this case. The function are only used there.


HB9HIL (approved):

HB9HIL:
Thanks :)

int2001 (approved):
good finding

PR #311, opened 2024-04-25
Added the tuned importer (from ADIF) to QSO-API as well

int2001:
API is now capable of Bulk-Import

### DIFF
--- a/application/controllers/Api.php
+++ b/application/controllers/Api.php
@@ -160,7 +160,10 @@ class API extends CI_Controller {
 	*/
 	function qso($dryrun = false) {
 		header('Content-type: application/json');
+		set_time_limit(0);
+		ini_set('memory_limit', '-1');
 
+		session_write_close();
 		$this->load->model('api_model');
 
 		$this->load->model('stations');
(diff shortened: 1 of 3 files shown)


application/controllers/Api.php
```diff
@@ -198,49 +204,33 @@ function qso($dryrun = false) {
...
 					}
+					if(count($record) == 0) {
+						break;
+					};
+					array_push($alladif,$record);
+					$return_count++;
+				};
+				$record='';	// free memory
+				gc_collect_cycles();
+				log_message("Error","Start");
```
phl0, line 226: Debug leftover?

phl0 (changes requested):
Debug code forgotten to remove?

application/controllers/Api.php
int2001: tnx. removed

assets/json/dok.txt
phl0, line 1: Should this and the updates of clublog and lotw be included here?!

phl0 (changes requested):
Changes in dok.txt, clublog_scp.txt and lotw_users.csv accidentially slipped into commit 1104f30bd0187bc1cd1739932abb9ffc6daf7aab. I think this should be reverted / committed in a separate PR.

assets/json/dok.txt
int2001: Sri - removed from tracking now

phl0 (approved):
Importing multiple QSOs with a single API call works.

PR #313, opened 2024-04-25
[LBA] Now uses profiles home grid

AndreasK79:
Before: used col_my_gridsquare or col_my_vucc_grids for plotting on the map.
Now: uses gridsquare set in station location.


HB9HIL (approved):

HB9HIL (approved):
works :)

phl0 (approved):
Tested ok. Makes sense to use station profile here as well.

PR #316, opened 2024-04-27
Added ITU Zone award

AndreasK79:


phl0 (approved):
Tested and works. New baby to play with :D But will be hard on SAT I guess ;-)

int2001 (approved):
tested. works!

PR #318, opened 2024-04-28
Added ITU support in LBA

AndreasK79:


int2001 (approved):
Nice one. Helps keeping track of Zones

PR #324, opened 2024-04-30
Only overwrite incoming LoTW grid if at least of the same precision

phl0:
The current code always overwrites existing gridsquare in logbook. This may be inconvenient if the user logged a more specific gridsquare than the QSO partner configured in his LoTW profile. As a first fix we only overwrite the grid now if it is at least of the same precision as the one in the logbook.

### DIFF
--- a/application/controllers/Lotw.php
+++ b/application/controllers/Lotw.php
@@ -525,7 +525,10 @@ class Lotw extends CI_Controller {
 					}
 					// Present only if the QSLing station specified a single valid grid square value in its station location uploaded to LoTW.
 					if (isset($record['gridsquare'])) {
-						$qsl_gridsquare = $record['gridsquare'];
+						// Only overwrite if incoming grid is at least of the same length/precision than existing in the QSO record
+						if (strlen($record['gridsquare']) >= strlen($status[2])) {
+							$qsl_gridsquare = $record['gridsquare'];
+						}
 					} else {
 						$qsl_gridsquare = "";
 					}
--- a/application/controllers/Lotw.php
+++ b/application/controllers/Lotw.php
@@ -524,15 +524,11 @@ class Lotw extends CI_Controller {
 						$state = "";
 					}
 					// Present only if the QSLing station specified a single valid grid square value in its station location uploaded to LoTW.
+					$qsl_gridsquare = "";
 					if (isset($record['gridsquare'])) {
-						// Only overwrite if incoming grid is at least of the same length/precision than existing in the QSO record
-						if (strlen($record['gridsquare']) >= strlen($status[2])) {
+						if (strlen($record['gridsquare']) > strlen($status[2]) || substr(strtoupper($status[2]), 0, 4) != substr(strtoupper($record['gridsquare']), 0, 4)) {
 							$qsl_gridsquare = $record['gridsquare'];
-						} else {
-							$qsl_gridsquare = "";
 						}
-					} else {
-						$qsl_gridsquare = "";
 					}
 
 					if (isset($record['vucc_grids'])) {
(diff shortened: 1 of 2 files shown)


application/controllers/Lotw.php
```diff
@@ -525,7 +525,10 @@ private function loadFromFile($filepath, $display_view = "TRUE")
 					}
 					// Present only if the QSLing station specified a single valid grid square value in its station location uploaded to LoTW.
 					if (isset($record['gridsquare'])) {
-						$qsl_gridsquare = $record['gridsquare'];
+						// Only overwrite if incoming grid is at least of the same length/precision than existing in the QSO record
+						if (strlen($record['gridsquare']) >= strlen($status[2])) {
```
HB9HIL, line 529: here is missing something. if the $record['gridsquare'] is set but has not the same precision as the origin record it does not define $qsl_gridsquare. I would suggest add another else $qsl_gridsquare = ""; Thx @int2001

HB9HIL (approved):
thx for this one :)

int2001 (changes requested):
as far as i read/understand this code it doesn't look into the grid itself, it only checks length.

e.g.: 

i log JN47gl
LotW confirms with JN47
JN47gl stays in WL

2nd example
i log JO48
LotW cnfms with JO49
LotW JO49 overwrites the entry JO48 in WL

3rd example (Dangerous!)
i log JO48gl
LotW cnfms with JO49
No overwrite will happen - inconsistence!

phl0:
I cannot cover case #3. How do you decide which grid is correct? Also LoTW can be incorrect as it was @HB9HIL's use case ...

HB9HIL:
> I cannot cover case #3. How do you decide which grid is correct? Also LoTW can be incorrect as it was @HB9HIL's use case ...

perhaps..

```
if (wavelog.grid.length() >= lotw.grid.length() && substr($wavelog.grid, 0, 4) != substr($lotw.grid, 0, 4)) { 
        do_nothing 
} else { 
         update 
}
```

int2001:
i'd say: LotW is the truth (because we're not able to complain about a wrong confirmed grid there - the grid which is cnfmd by LotW counts).

Proposal from @HB9HIL reads good

application/controllers/Lotw.php
```diff
@@ -524,10 +524,11 @@ private function loadFromFile($filepath, $display_view = "TRUE")
 						$state = "";
 					}
 					// Present only if the QSLing station specified a single valid grid square value in its station location uploaded to LoTW.
+					$qsl_gridsquare = "";
 					if (isset($record['gridsquare'])) {
-						$qsl_gridsquare = $record['gridsquare'];
-					} else {
-						$qsl_gridsquare = "";
+						if (strlen($record['gridsquare']) > strlen($status[2]) || substr(strtoupper($status[2]), 0, 4) != substr(strtoupper($record['gridsquare']), 0, 4)) {
```
int2001, line 529: Sri. but this would mean: We take OUR grid (from WL) if we're more precise than LotW.

e.g.: LotW cnfms with JO30, Op entered JO31oo. --> JO31oo is taken because it's more precise
that's wrong imho

phl0: > e.g.: LotW cnfms with JO30, Op entered JO31oo. --> JO31oo is taken because it's more precise that's wrong imho

Why is that wrong? For VUCC / LoTW only 4-char grids count AFAIK. So why not keep our 2-char more specific suffix?

phl0: > Why is that wrong? For VUCC / LoTW only 4-char grids count AFAIK. So why not keep our 2-char more specific suffix?

Just disregard my comment... too blind to read.

phl0: > Why is that wrong? For VUCC / LoTW only 4-char grids count AFAIK. So why not keep our 2-char more specific suffix?

Different 4-char grid is covered by the 2nd part or the OR stuff.

int2001: Sorry - reversed boolean logic is confusing sometimes.
This one works correct.

PR #325, opened 2024-04-30
Autodisable QRZ-Upload if AUTH-Error // Logic-part

int2001:
This one disables the QRZ-Upload for a station which has entered the wrong API-Key, and QRZ answers with "NO AUTH" temporary.

No new column is used, instead qrzrealtime is used. currently it is enumed as followed:
0 = no realtime (but cron)
1 = realtime (and cron)

new on Top:
-1 = temporary disabled (will be setable by user via UI / Stationsetup, or automaticly if auth fails)


HB9HIL:
added a test button with a bit of UI chichi :D

https://github.com/int2001/wavelog/pull/8

HB9HIL (approved):
good improvement. brings stability

PR #330, opened 2024-05-02
Menu sorted

HB9HIL:
Sorted the Award Menu to make the liste less long.

image


int2001:
Nice Flags, but on Windoze is looks like the following. And yes, the same ""Bug" is at other places in wavelog, but never at the main-menu.

image

phl0 (approved):
Works. Tested with:

- Linux Firefox and Chromium
- Windows Firefox and Edge browser.

int2001 (approved):
Finally: Flagproblem with problematic OSes solved. tnx

PR #332, opened 2024-05-03
Make hams.at switchable

phl0:
This refactors the hams.at page to make use of a JSON request. Also a button was added to quickly switch between all passes and workable passes without editing the user profile.


HB9HIL (approved):

int2001 (approved):
Like it!

PR #334, opened 2024-05-03
Fix QSL arrows

HB9HIL:
fixes a small frontend bug where the qsl arrows perform a nasty line wrap

before
image  
  
after
image


phl0 (approved):
Fix w0rx!

PR #336, opened 2024-05-04
[CRON] Option to restrict IP range for crons ('cron/run')

HB9HIL:


HB9HIL:
PR closed

Can be solved in .htaccess file

PR #338, opened 2024-05-05
[C to M] Part 1 - Update_Model

HB9HIL:
This is part 1 of a big movement of functions from the controllers to models.


HB9HIL:
Todo for the last part of this series: 

Debug View File updates change to direct call with info dialog for result

HB9HIL:
Missing: 

update_dxcc because it's bit more complex because of the dynamic html in the web UI

int2001 (approved):
Tested this on my production. Works

Untested (but i think it'll work, since it is "only shifting" of functions): Call from cron.

Will bring more clean code in future.

PR #340, opened 2024-05-05
Cron special expressions (@...)

HB9HIL:
Cron Manager now supports @ expressions in PHP and JS

JS is handled by the cronstrue.min.js which comes from https://github.com/bradymholt/cRonstrue and the source is adjusted by me. ([See PR at external Repo](https://github.com/bradymholt/cRonstrue/pull/318))

image


int2001 (commented):
tbh: i don't know how to test it. it's an edge-case for wavelog.
Should be part of the cron-project for PHP.

Sorry. Perhaps someone else can bring up the effort for testing it.

HB9HIL:
> tbh: i don't know how to test it. it's an edge-case for wavelog. Should be part of the cron-project for PHP.
> 
> Sorry. Perhaps someone else can bring up the effort for testing it.

Edit a Cronjob and type in `@weekly` 

it now should recognize it in JS and PHP and is able to calculate the next run.

phl0 (approved):
Tested the stuff on PHP8. All expressions work so far. I will setup a PHP7 machine to test the backwards compatibility tomo.

PR #341, opened 2024-05-05
Option to show UTC time in header menu

HB9HIL:
user option

same logic as quickswitcher

image


phl0 (approved):
I like the idea. Code works.

int2001 (changes requested):
it's within the "quickswitcher".
so if quickswitcher is disabled the time won't show up (even if enabled in options)

PR #343, opened 2024-05-05
Fixes UTC at header

int2001:
show when enabled, even if Quickswitch was disabled


HB9HIL:
I already fixed in a direct commit

https://github.com/wavelog/wavelog/commit/fad182e3f8ac79acb688c154b653af319c3503b3

PR #347, opened 2024-05-07
Adif export can now use filters, no need to mark QSOs

AndreasK79:
The adif export in the advanced logbook can now use the filter for ADIF export, so that you do not need to mark QSOs you want to export.


phl0 (approved):
Works as expected :)

phl0 (approved):
Still working as expected ;-)

PR #348, opened 2024-05-07
Responsive GridMap height

HB9HIL:
use "ignore whitespace" for review!

test: 
all gridmasters
ffma award
gridmap 
visitor gridmap

wait for #347 to avoid conflicts


phl0 (approved):
Nice solution. Works as expected.

phl0 (approved):
Still works after solving merge conflicts.

PR #351, opened 2024-05-09
Added the ability to check for LoTW-user in Quick Lookup

AndreasK79:
To test: ALT-L, choose LoTW user and enter a call.
image


int2001 (approved):
Works. Nice one

PR #352, opened 2024-05-09
UI passwords

HB9HIL:
Requested by @int2001 

For usability we return `**********` in the password fields in user/edit
also we can allow to clear the password from database if the field is empty on submit

### DIFF
--- a/application/controllers/User.php
+++ b/application/controllers/User.php
@@ -227,6 +227,7 @@ class User extends CI_Controller {
 		$query = $this->user_model->get_by_id($this->uri->segment(3));
 
 		$data['existing_languages'] = $this->find();
+		$pwd_placeholder = '**********';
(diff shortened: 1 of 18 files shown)


HB9HIL:
found som other bugs. fixed in this PR

HB9HIL:
> did i miss something?
> 
> when not changing the PW it is set to multiple '*' with in the database (Overwritten)

solved

int2001 (approved):
Works.

Good finding with the options.
Things tested:
- Change Password at own user.
- Trying to set them to "NULL" (Works as expected. UserPW wasn't set to NULL - thats right)
- Tried the same with other users (as admin) - works
- tried with index.php visible/enabled as well as disabled - works.

Well done and thank you Fabian

PR #354, opened 2024-05-09
Download Clublog-"QSLs" to Wavelog

int2001:
This one brings "QSLs" from Clublog to wavelog and adds a cronjob for it to the new Cron-Manager.
Wiki-Entry (for classic users) will follow after merging it.

Includes:
- Taking care of "Clublog-Ignore" for Stations
- Setting Clublog-Ignore to true, if someone tries to download QSLs for Calls he's not granted at Clublog
- Aggregates the Download by "Call" - since Clublog doesn't handle our station_profiles
- Shows QSL-State at QSO-Detail-View

Does not include (to reduce PR, overview, etc.):
- Arrows for Clublog-status (Will do that in a later PR)
- UI for triggering the DL (There's also no UI for manually triggering the upload)

### DIFF
--- a/application/migrations/197_clublog_recvd.php
+++ b/application/migrations/197_clublog_recvd.php
@@ -3,13 +3,25 @@ defined('BASEPATH') or exit('No direct script access allowed');
 
 class Migration_clublog_recvd extends CI_Migration {
 	public function up() {
-                if (!$this->db->field_exists('COL_CLUBLOG_QSO_DOWNLOAD_STATUS', $this->config->item('table_name'))) {
-                        $fields = array(
-                                'COLUMN COL_CLUBLOG_QSO_DOWNLOAD_DATE DATETIME NULL DEFAULT NULL',
-                                'COLUMN COL_CLUBLOG_QSO_DOWNLOAD_STATUS VARCHAR(10) DEFAULT NULL',
-                        );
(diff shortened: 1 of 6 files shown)


application/migrations/197_clublog_recvd.php
```diff
@@ -0,0 +1,35 @@
...
+				'id' => 'clublog_download',
+				'enabled' => '0',
+				'status' => 'pending',
+				'description' => 'Download QSOs from Clublog',
+				'function' => 'index.php/clublog/download',
+				'expression' => '7 00 * * *',
+				'last_run' => null,
+				'next_run' => null
+			));
+			$this->db->insert_batch('cron', $data);
```
HB9HIL, line 24: please make a check if the entry `clublog_download` already exists.

HB9HIL (approved):
Tested: 
[x] migration (but please add a check as mentioned above
[x] Cron in Cronmanager
[x] Cron in CLI with Curl
[x] Clublog Ignore
[x] Download with real QSO's
[x] Last Rcvd Date before download

phl0 (approved):
W0rx!

application/migrations/197_clublog_recvd.php
int2001: Done. Good find. should be implemented as well at many other points where data is added via migration.

PR #355, opened 2024-05-09
Contestcalendar

HB9HIL:
@AndreasK79 

I tweaked your contestcalendar a bit. Wrote a parser which creates proper DateTime from the "description" in the RSS data.


HB9HIL:
Comment: 

I used `curl` instead `file_get_contents` to get the rss data

AndreasK79:
Tested working. Much nicer presentation! Good work.

PR #357, opened 2024-05-13
[WIP] OpenSSL signatures need a newline after base64 string

phl0:
As per definition the OpenSSL signatures should have a newline at the end.

In addition we strip the complete upload path for LoTW uploads.

Before:
Screenshot from 2024-05-13 08-14-15

After:
Screenshot from 2024-05-13 09-26-49


int2001 (approved):
Works and is now:
- more compatible with SSL
- less data-leasking (due to stripping of path)

PR #358, opened 2024-05-13
Clublog UI

int2001:
Bring Clublog-QSLs to Frontend:
- Overview (if chosen as "Default QSL-Method" in settings)
- Last-QSOs (if chosen as "Last-QSL" in Settings)

Affected views:
- User-Settings
- Overview
- log_partial
- QSO-View

Caution: Arrows at Overview are depending on "Default QSL-Methods" (User-settings) from now on, to avoid "Arrow-Inflation". 

Keeping things small: Checkboxes at Awards will be another PR

### DIFF
--- a/application/config/migration.php
+++ b/application/config/migration.php
@@ -22,7 +22,7 @@ $config['migration_enabled'] = TRUE;
 |
 */
 
-$config['migration_version'] = 197;
+$config['migration_version'] = 198;
 
 /*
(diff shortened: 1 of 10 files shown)


phl0 (approved):
Works for me. Tested:

- Logbook Overview
- Last QSOs (upon logging)
- User Settings
- QSO view

PR #359, opened 2024-05-14
Reload with same settings re workable passes

phl0:
The JS timer would always reload the page with the predefined value for showing workable passes. This is changed so that workable/all are still shown once the button is pressed and the preset is overriden.


int2001 (approved):
works!

PR #361, opened 2024-05-15
some tweaks

dg0tm:


int2001 (approved):
Tested. Works and is much more intuitive for the "non-docker-user" as well.

Pse adjust wiki

PR #364, opened 2024-05-16
Show error message if LoTW upload fails

phl0:
The current code throws an exception which causes the script to terminate without any output. This is to change so that the user sees an error message that upload had failed:

Screenshot from 2024-05-16 08-26-11


HB9HIL (approved):
works

int2001 (approved):
Only "dry-run" (Eye-Integral) tested.
looks good

phl0:
Then testing in dev commences ;D

PR #365, opened 2024-05-16
3rd dis

int2001:
This one disables Syncing to/from 3rd party-services at the GUI for Users.

Helpful if u have a huge instance or active-users, because of preventing race-conditions between UI and Cron


phl0 (changes requested):
Tested following stuff:

- LoTW: works
- eQSL: works
- HRDlog: works, but UI has wrong variable in code (looks for eqsl config option instead of hrdlog)
- qrz.com: works, but does not hide the warning like hrdlog does.

See PR.

phl0 (approved):
Issues fixed. All good to go now.

PR #366, opened 2024-05-16
Only show and evaluate 4-char grids for hams.at

phl0:
We should only evaluate (and show) 4-char grids for hams.at display. 6-char grids lead to false information:

Screenshot from 2024-05-16 12-59-35


int2001 (approved):
Thankfully there were some stations on hams.at with an overprecise grid to test ;)
Works

PR #367, opened 2024-05-16
Show prop mode in eQSL pages

phl0:


int2001 (approved):
More flexible :)
Works

PR #368, opened 2024-05-16
Use cURL for LoTW download to be able to catch errors

phl0:
Use cURL for LoTW downloads so that we are able to catch errors. WIP as this needs to be verified once LoTW is back online.

### DIFF
--- a/application/controllers/Lotw.php
+++ b/application/controllers/Lotw.php
@@ -253,7 +253,7 @@ class Lotw extends CI_Controller {
 				// Nothing to upload
 				if(empty($data['qsos']->result())){
 					if ($this->user_model->authorize(2)) {	// Only be verbose if we have a session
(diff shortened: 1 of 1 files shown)


phl0:
In case of errors during download it should look like this:

Screenshot from 2024-05-16 18-30-36

phl0:
Just tested with a locally provided LoTW report file. Working fine so far. So undrafting here.

int2001 (changes requested):
it still leaks lotw-credentials to the log (perhaps a try/catch can solve it)
But: Good improvement. One more piece of code, where we get rid of fopen.

Needs to be tested again when lotw is back online

phl0:
I will have a look if I can get it to *not* spit credentials to log or GUI. Tnx

int2001 (approved):

PR #369, opened 2024-05-16
[QSO] Max input lenght for name and location

HB9HIL:
Our DB allows 128 chars for COL_NAME and 64 chars for COL_QTH. 

At the moment you can enter longer strings as they can be in the database. This causes the qso save call to fail. This change limits the max input lenght for these two fields to the max chars set in database.


phl0 (approved):
Tested. Works.

PR #370, opened 2024-05-16
Show slashed zero in LBA and hams.at

phl0:
Works in hams.at. Maybe @AndreasK79 can help testing LBA. I checked bulk update and ADIF export which are still working.


AndreasK79 (approved):

AndreasK79:
Haven't found any problems so far :)

phl0:
So then merge time here. Will look at maps separately.

PR #372, opened 2024-05-17
Changed worked/confirmed to badges instead

AndreasK79:
Before:
image

After:
image


phl0 (approved):
Looks better and works as designed.

PR #373, opened 2024-05-18
Refactor QSO entries on dashboard map

phl0:
Refactored QSO entries on dashboard map a little:
- Zeros replaced by slashed zero
- Separate 2nd entry for callsign removed
- Name displayed instead if present

**Before**:

Screenshot from 2024-05-18 07-02-52

**After**:

Screenshot from 2024-05-18 07-00-48


int2001 (approved):
Works

PR #374, opened 2024-05-18
Added sat indexes to speedup LBA search and map

AndreasK79:
Before: 3-3.5s to load the Advanced logbook.

After: 1s to load the Advanced logbook.


phl0 (approved):
Confirmed that it is at least 50% of loading time now.

int2001 (approved):
Works great.
Testeed with mariadb 11.3.2 down to mariadb 10.3.12

PR #375, opened 2024-05-18
Cleaned up the hardware table so that it looks like other tables

AndreasK79:
Before:
image

After:
image


phl0 (approved):
Looks much cleaner.

int2001 (approved):
like that.
it's the small things which are gaining more comfort ;)

PR #376, opened 2024-05-19
Added propagation in the table columns

AndreasK79:
This is for the Advanced logbook. Since we have the filter, I added it to be displayed in the columns.


int2001 (approved):
Works. Still fast ;)
i like

phl0 (approved):
You read my mind. 👌🥳

PR #381, opened 2024-05-20
Catch frequency being null / empty

phl0:
If a radio is selected which has not been updated via CAT API yet the frequency being asked for is set to 'null':

Screenshot from 2024-05-20 17-48-01

That results in a PHP error being inserted into the frequency input field:
Screenshot from 2024-05-20 17-36-36

The error message reads:

```
<div style="border:1px solid #990000;padding-left:20px;margin:0 0 10px 0;"><h4>A PHP Error was encountered</h4><p>Severity: Warning</p><p>Message:  Attempt to read property "data" on null</p><p>Filename: libraries/Frequency.php</p><p>Line Number: 138</p></div>
```

So we need to catch this case and return an empty result for the default frequency if the band was null or 'null'. This PR achieves that.

### DIFF
--- a/application/controllers/Qso.php
+++ b/application/controllers/Qso.php
@@ -431,9 +431,11 @@ class QSO extends CI_Controller {
 
 	function band_to_freq($band, $mode) {
 
-		$this->load->library('frequency');
+		if ($band != null and $band != 'null') {
+			$this->load->library('frequency');
+			echo $this->frequency->convert_band($band, $mode);
+		}
 
-		echo $this->frequency->convert_band($band, $mode);
 	}
 
 	/*


application/controllers/Qso.php
```diff
@@ -431,9 +431,11 @@ function delete_ajax() {
 
 	function band_to_freq($band, $mode) {
 
-		$this->load->library('frequency');
+		if ($band != null and $band != 'null') {
```
int2001, line 434: works with PHP8.3. but: wouldn't

`if (($band ?? '') != '') {`

be better, and more failsafe? or is the string "null" coming from JS?

phl0: > is the string "null" coming from JS?

Yepp. See the screenshot of the request above.

int2001: roger. in that case: approve

int2001 (approved):

PR #388, opened 2024-05-22
Check and hide double announced grids in hams.at

phl0:
Check and hide double or multiple entered grids in hams.at announcements.

Before:

Screenshot from 2024-05-22 08-15-17

After:

Screenshot from 2024-05-22 08-14-54


int2001 (approved):
Thanks god there were some eligible Spots with double-Grids for testing.
Works!

PR #389, opened 2024-05-22
[LBA] Added "Not received" Action Button

HB9HIL:
- Added an action button to LBA where a user can set QSL to not received
- Made "not" buttons red to seperate them and make it visually more intuitive
- Fixed a small bug where `$data['orbit']` is not set in $data array. Maybe @phl0 can check that.


phl0 (approved):
Confirm the orbit fix.

phl0:
Rest also works fine for me.

AndreasK79 (approved):

PR #390, opened 2024-05-22
Fix empty DXCC on LBA update functions

phl0:
The JS update / QSL functions of LBA return an empty DXCC. This fixes the issue by using the correct lookup from DXCC table for dxccname.


HB9HIL (approved):
seems this fixes the issue

AndreasK79 (approved):

AndreasK79:
@phl0 seems I missed this one. Thanks for fix. I also updated so that contest is updated too.

PR #392, opened 2024-05-22
Catch empty grid and inform user to correct config

phl0:
Inform the user if not station location is configured (i.e. no grid set):

Screenshot from 2024-05-22 17-38-44

Alternative to https://github.com/wavelog/wavelog/pull/391.


int2001 (approved):
i like

PR #394, opened 2024-05-23
Fence: Strip length of Lon/Lat so it fits DB

int2001:
Prevent DB-Errors if LAT/LON are to long:

```
Severity: error --> Exception: Data too long for column 'COL_LAT'
Severity: error --> Exception: Data too long for column 'COL_LON'
```


phl0:
How do you reproduce this error? Just checked the code and there is a limit to 11 chars in line 3610 further up the code. Tested with an ADIF import and a few debug outputs before and after that line:

```
DEBUG - 2024-05-23 06:26:11 --> 1st: 51.12345678901234567 - length: 20
DEBUG - 2024-05-23 06:26:11 --> 2nd: 51.12345678 - length: 11
DEBUG - 2024-05-23 06:26:11 --> 3rd: 51.12345678 - length: 11
```

At least during ADIF import lat is trimmed to 11 chars ...

int2001:
Absolutely right.
Source must be a different one than ADIF-Import

PR #397, opened 2024-05-23
Loose Weight

HB9HIL:
I noticed we have a lot of files in assets, which are neither loaded, nor somewhere else needed. So we can get rid of them.


HB9HIL:
Explaination of the changes: 

`controllers/Dashboard.php` and `setup/check_list.php`

- Removed the setup checklist as it's not in use.



`awards/was/map.php`

- Removed unused demo map file



`interface_assets/footer.php header.php mini_header.php`, `qso/edit_ajax.php`,  
`visitor/exportmap/header.php` and `visitor/layout/header.php`

- Adjusted the source loadings for bootstrap and fontawesome


`assets/css/bootstrap-*`

- Removed not needed bootstrap files (grid/reboot)


`assets/css/[themes]`

- Added the full bootstrap library files. They don't get loaded unless the browser devtools are open but they help debugging CSS



`assets/fontawesome/*`

- We only use the CSS/Webfonts way to implement fontawesome icons. So we can get rid of all the other stuff



`assets/js`

- we use the bootstrap.bundle stuff. So I added the min and map files for it and removed the "non-bundle" js files
- removed sourcing for easyprint.js as we don't have the source `bundle.js` in the repo and also the map is not necessary
- other unused files like jquery.usmap.js are removed
- removed leaflet esm modules as we don't use them, also moved the leaflet-src.js to a `dist` subfolder for correct mapping
- removed other unused files like raphael.js or color.jquery.js (they belonged to awards/was/map.php)
- added showdown.js so the mapping works

PR #403, opened 2024-05-24
Decrease db queries for dashboard rendering

AndreasK79:
Decreased several DB calls for the dashboard by making small changes to VUCC model.
Also redid not calling options twice in header.
Reduced queries in LBA too.
Fixed VUCC when you click on grids worked. Now only needs one query per grid, not two.


phl0 (approved):
My dashboard still works :)

int2001 (approved):
tested it.
great performance. i like!

PR #405, opened 2024-05-24
Activators Refactoring

HB9HIL:
There was a frontend bug in gridsquare activators

image

I took this opportunity and did a small code refactoring. 

- we now call `qso_detail_ajax` instead a dedicated modal just for this category
- moved javascript from footer to a dedicated js file. Loading with timestamp caching
- also the datatable gets loaded on page load with some default values


phl0 (approved):
Still working :)

PR #406, opened 2024-05-25
Calculation issue if lat/lon or some grid-information is missing

int2001:
i some situations users are uploading / logging QSOs without any (or only partial) Location-Informations.
In that case WL tries to calculate distance and so on out of the given Felds.

This fails (if there's not enough information). PR addresses that.

@AndreasK79 is going to enrich data out of the given DXCC.


AndreasK79 (approved):

AndreasK79:
Nice find, and it helped uncover another bug :)

PR #407, opened 2024-05-25
Fixed a few errors that popped up while clicking around

AndreasK79:
Fixed this when trying to add qsos to a qsl:
image
Fixed this when entering mode, and the line doesn't seem to do anything:
image


int2001 (approved):
works.
always good to fix those small ones ;)

PR #409, opened 2024-05-25
[PUBLIC VIEW] Global option to disable public maps

HB9HIL:
Now it's possible to disable the maps (main map and gridsquares map) in the public view.


AndreasK79 (approved):
Tested working!

phl0 (commented):
Works. But hides the map by default. Maybe we should add a migration to leave the map enabled by default and only have the users disable it who wanted this setting instead of the other way round?

PR #410, opened 2024-05-25
[PUBLIC VIEW] Hide QSO Time in Public Map

HB9HIL:
We want to hide the QSO time in the public map. 

qso details on map

Additionally we now show the QSO date and time on the dashboard map in the user_date_format


phl0 (changes requested):
IMHO it needs a little more code. Showing/hiding of time should be based on config setting as for the time column in the table below. Visitor's date format should also be derived from config instead of hard coded. See PR https://github.com/HB9HIL/wavelog/pull/20.

phl0 (approved):
Fine with me now :)

PR #414, opened 2024-05-26
Fixes QSL-Columns

### LINKED ISSUE #412: Enabled QSL-Methods not shown correctly in Logbook Overview
**Describe the bug**
With having one or multiple QSL-Methods enabled in the account, disabling QSL results in hiding the LoTW column instead of QSL column in Logbook overview.
In this example, only enabling LoTW is described.

**Version Information**
- OS:               FreeBSD
- PHP:              8.1
- MySQL/MariaDB:    10.6.16
- Wavelog:          1.6 

**Steps to Reproduce**
1. Go to 'Account'
2. Click on 'Default Values'
3. Enable "LoTW" only
4. Got to Logbook -> Overview
5. See the QSL column instead of LoTW

**Expected Behavior**
Only show the enabled QSL-Methodes columns

**Screenshots**
Settings
Bildschirmfoto 2024-05-26 um 12 36 04

Logbook Overview
Bildschirmfoto 2024-05-26 um 12 36 17

**Desktop Information:**
- OS:               MacOS 14.5
- Browser:         Chrome / Safari
- Version:          17.5 / 125.0.6422.76

int2001:
Should Fix #412

### DIFF
--- a/application/views/view_log/partial/log_ajax.php
+++ b/application/views/view_log/partial/log_ajax.php
@@ -69,19 +69,19 @@ function echo_table_col($row, $name) {
                 echo_table_header_col($this, $this->session->userdata('user_column5'));
 
                     if(($this->config->item('use_auth')) && ($this->session->userdata('user_type') >= 2)) { 
-    		    	if ( substr_count($this->session->userdata('user_default_confirmation'),'Q') >0  ) { ?>
+    		    	if ( strpos($this->session->userdata('user_default_confirmation'),'Q') !== false  ) { ?>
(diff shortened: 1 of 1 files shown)


phl0 (approved):
Works. But any reason for switching to subsr_count instead of strpos?

int2001:
two ways to solve:
a) work with strpos, but it needs the more "lax" operator `!== false` instead of `!= false` (because it returns false if no match found,. but position if match was given - inconsistent types imho)
b) work with substr_count - which works oob

so i decided myself to use substr_count.

PR #415, opened 2024-05-26
Import speed

AndreasK79:
This PR aims for speed improvement when importing an ADIF file.

It does contain a few tweaks for removing unneeded DB calls here and there as well.


phl0 (approved):
Cannot reliably do speed diff calculations atm. But functions still working fine. So approving here.

AndreasK79:
Finally managed to test it myself. Results are the same as far as I can tell :)

PR #421, opened 2024-05-29
Fix last_seen (Compare Timestamp instead of seconds after full minute)

int2001:
Tnx for finding this nice bug, @HB9HIL 
Should fix it

### DIFF
--- a/application/models/User_model.php
+++ b/application/models/User_model.php
@@ -538,7 +538,7 @@ class User_Model extends CI_Model {
 		if(($this->validate_session($u)) && ($u->row()->user_type >= $level) || $this->config->item('use_auth') == FALSE || $level == 0) {
 			$ls=new DateTime($u->row()->last_seen ?? '1971-01-01');
 			$n=new DateTime("now");
-			if ($ls->diff($n)->format('%s') > 60) {	// Reduce load of the Spy-Function. shouldn't be called at anytimne. 60seconds diff is enough
+			if ($n->getTimestamp()-$ls->getTimestamp() > 60) {	// Reduce load of the Spy-Function. shouldn't be called at anytimne. 60seconds diff is enough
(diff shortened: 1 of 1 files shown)


HB9HIL (changes requested):
We don't need to create an object first an then get the Timestamp from that object. It can be done way simpler and faster. 

https://github.com/int2001/wavelog/pull/11

For speed comparison between DateTime Class and getTimestamp() vs strtotime() refer to this benchmark

https://code-bude.net/2012/10/22/benchmark-strtotime-vs-datetime-vs-gettimestamp-in-php/

HB9HIL (approved):

HB9HIL:
Thx

int2001:
Great that we now safe a bit more time every 60seconds. tnx for the better approach to @HB9HIL

application/models/User_model.php
```diff
@@ -536,9 +536,10 @@ function authorize($level) {
 			$level = $this->config->item('auth_mode');
 		}
 		if(($this->validate_session($u)) && ($u->row()->user_type >= $level) || $this->config->item('use_auth') == FALSE || $level == 0) {
-			$ls=new DateTime($u->row()->last_seen ?? '1971-01-01');
-			$n=new DateTime("now");
-			if ($ls->diff($n)->format('%s') > 60) {	// Reduce load of the Spy-Function. shouldn't be called at anytimne. 60seconds diff is enough
+			$ls = strtotime($u->row()->last_seen ?? '1971-01-01');
```
phl0, line 539: Is there a reason we use 1971 instead of 1970 (i.e. Unix timestamp of 0).

phl0: So shall I fix it (directly)?

int2001: yep.

phl0: Fixed.

PR #422, opened 2024-05-29
Fix post dxcclookup

int2001:
This one fixes a bug, when looking up "past" calls at qso-form.

e.g.:
- GB6WW normally belongs to Great Britain
- There was an exception during May (1st till 28th) where GB6WW belongs to Scotland. Well documented at Clublog and our dxcc-exceptions
- if you logged GB6WW during that time everything resolved to Scotland (assumed Countryfiles up2date)

Consider "Post-Logging". Here's the bug.
When logging GB6WW (today) with the QSO-Date 24th of May it'll resolve to "England" which is wrong.

Solution:
added an optional parameter for the date to the json-lookup function. default is - as before - sysdate.

### DIFF
--- a/application/controllers/Logbook.php
+++ b/application/controllers/Logbook.php
@@ -84,13 +84,17 @@ class Logbook extends CI_Controller {
         echo json_encode($return, JSON_PRETTY_PRINT);
     }
 
-	function json($tempcallsign, $tempband, $tempmode, $tempstation_id = null) {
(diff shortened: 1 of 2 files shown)


phl0 (approved):
Yeah works for me. Very good finding and fix (had GB6WW logged via API and with wrong DXCC of course. Forgot to update after email from clublog).

PR #423, opened 2024-05-29
Check LoTW upload/fail correctly

phl0:
We should check LoTW dates correctly (i.e. timestamps instead of comparing alphabetically).


HB9HIL:
tested, works

int2001 (approved):
Tested this.
One more piece of code which hardens the Application. nice :)

PR #427, opened 2024-05-29
Hide hint and button if number of workable passes matches number of all passes

phl0:
We should hide the hint and the button re workable passes if the number of workable and non-workable passes matches (i.e. exactly same passes).


int2001 (approved):
Injected a little fake-data to simulate non-workable-rows. Button appears in that case and does it it has to do.

makes UI more clean.

PR #428, opened 2024-05-29
[SFLE Part 2] Styling Adjustments and Gridsquare Support

HB9HIL:
SimpleFLE got some styling adjustments. QSO Table is larger and responsive to windows size. 

An options-button and the corresponding functions are prepared for the last part 3 which will contain a option for "Auto Callbook Lookup"

Additionally SimpleFLE does now support logging gridsquares (4-10 digit locators). Wiki will be updated after merge. For the Gridsquare support I dropped the operator column in the qso table. The operator still gets logged as defined in the basic config, but we don't need to show the same operator for each QSO (we do not support operator change within a sfle-session).


phl0 (approved):
Works fine so far. Like the new layout. 
Don't forget to reset logging session. Else you will not see the changes in the table if still some older format stored and not cleared.

PR #431, opened 2024-05-30
[QSO] Only overwrite Frequency if no Radio is selected

### LINKED ISSUE #430: QSO data mismatch hardware interface data
### Discussed in https://github.com/wavelog/wavelog/discussions/426

<sup>Originally posted by **R1BLH** May 29, 2024</sup>
Sometimes i found a strange situation. 
I'm using FT-857 controlled by PC via Omnirig. To connect Omnirig with Wavelog I used CloudCat application.
Information about the transceiver frequency is normally displayed on the dashboard, but to the QSO entry form only the band information is transferred correctly, but the frequency field is filled with the default frequency for the band.
Please have a look at the screenshots:
Omnirig (CloudCat):
изображение
Dashboard:
изображение
Station data from new QSO form:
изображение
QSO form data:
изображение

Update:
Before transmit all was OK.
During transmit all is 0:
изображение
After switching from transmit to receive band is correct, frequency is default:
изображение

</div>

HB9HIL:
It happens when choosing a radio in QSO form, that the frequency gets overwritten by `band/mode.change(function()`. 

Solution would be to just overwrite frequency if no radio is selected.

Fixes #430

### DIFF
--- a/assets/js/sections/qso.js
+++ b/assets/js/sections/qso.js
@@ -1029,20 +1029,24 @@ $( document ).ready(function() {
 
 	/* on mode change */
 	$('.mode').change(function() {
-		$.get(base_url + 'index.php/qso/band_to_freq/' + $('#band').val() + '/' + $('.mode').val(), function(result) {
-			$('#frequency').val(result);
-			$('#frequency_rx').val("");
-		});
+		if ($('#radio').val() == 0) {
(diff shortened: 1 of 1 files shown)


int2001 (approved):
Fixes a bug, when CAT isn't providing QRG continuously after a mode or band-change.

looks okay for me.

PR #437, opened 2024-05-30
Show hint and button on table load

phl0:
We should hide the hint initially and show them when table is loaded instead of showing them and hiding later.


HB9HIL (approved):
perfect :)

PR #438, opened 2024-05-31
Map Performance

HB9HIL:
We can further increase the performance: 

The map needs two ajax calls to render. One for the "map_custom" feature and one for the QSOs.
By writing the map_custom stuff into the sessiondata and a JS variable we can reduce the time for loading the maps. 

To accomplish this I moved the ajax targets (controller functions) into the optionslib to make it available. 

On Dashboard, Logbook and LBA we now have one sql query and one ajax call less. You feel that the map is loading a bit faster. 
On Exportmap and Visitor View we just have one ajax call less since we don't have sessiondata here. But you also feel an improvement.

### DIFF
--- a/application/controllers/Dashboard.php
+++ b/application/controllers/Dashboard.php
@@ -53,6 +53,8 @@ class Dashboard extends CI_Controller
 		$data['themesWithoutMode'] = $this->setup_model->checkThemesWithoutMode();
 		$data['dashboard_map'] = $this->optionslib->get_option('dashboard_map');
 
+		$data['user_map_custom'] = $this->optionslib->get_map_custom();
+
 		$this->load->model('cat');
 		$this->load->model('vucc');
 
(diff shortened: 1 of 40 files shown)


int2001 (changes requested):
This one gains speed for Visitor-Map and LBA because the ajax was eliminated.

On the other hand it adds a (unnecessary) Call to nearly EVERY Page/View and wastes sesion_var-Space for data which is only needed at maps.

Proposal: Eliminate those AJAX-Calls (good idea) by fetching the data at controller-level for those pages where it is needed directly in PHP

HB9HIL:
> This one gains speed for Visitor-Map and LBA because the ajax was eliminated.
> 
> On the other hand it adds a (unnecessary) Call to nearly EVERY Page/View and wastes sesion_var-Space for data which is only needed at maps.
> 
> Proposal: Eliminate those AJAX-Calls (good idea) by fetching the data at controller-level for those pages where it is needed directly in PHP

You are right.. Could have thought that myself. Changes made. Still fast :)

int2001 (approved):
Tested:
- MapViews at LBA and Visitor
- ExportMap

Both still working and feeling much faster.

PR #439, opened 2024-05-31
Sorting hams.at table

phl0:


int2001 (approved):
works. makes life easier ;)

PR #441, opened 2024-05-31
minor adjustments

HB9HIL:
two small adjustments:

- Hide the radio_cat_state in QSO form if users switches back from radio to no-radio
- change location in SimpleFLE, if the user changes active location with quickswitcher (same as #335 but for SimpleFLE)


phl0 (approved):
Both changes working fine.

PR #442, opened 2024-05-31
Enable Logging on CR at Callsign-Field when running a no-xchg-contest

int2001:
When running a contest without XCHG, there was no way to log a QSO by pressing Return.
This one enables logging "on Return" in only that case (all other cases untouched)


HB9HIL (approved):
works as expected

PR #444, opened 2024-06-02
Deactivate Clublog-Upload on Reject to prevent being blacklisted at CL

int2001:


phl0:
Reading the code CL upload is them disabled until user enables agn?

int2001:
yep. user must manually re-enable it (after he setted up everything right at Clublog) - it's the same logic as with Qrz.
Otherwise WL is doing requests at every cronjob again.

phl0 (approved):
Works fine for wrong credentials :)

PR #445, opened 2024-06-03
[C -> M] eQSL

HB9HIL:
created eqsl `sync()` function in eqslmethods model aswell. The function in the controller will stay to not break existing classic crons. Related functions were moved to the model. 

Tested
- call URL/eqsl/sync = works
- call different UI functions = works


int2001 (approved):
Reviewed code: Looks good. Can't spot an issue here.
Tested: All Manual parts (Upload / Download / fetching pics, etc.): Working!
Untested (because tricky): Cron.

PR #448, opened 2024-06-04
Lock Account also at download

int2001:


application/models/Clublog_model.php
```diff
@@ -159,11 +159,15 @@ function downloadUser($userid, $username, $password) {
 					$return = curl_error($request);
 				} elseif (preg_match_all('/Login rejected/', $response)) {
 					$return = "Wrong Clublog username and password for Callsign: '" . $station_row->station_callsign . "'. 'LOGIN REJECTED'.";
-					log_message('error', $return);
+					log_message('debug', $return);
+					$sql = 'update station_profile set clublogignore = 1 where station_id = ?';
+					$this->db->query($sql,$station_row->station_id);
 				} elseif (preg_match_all('/Invalid callsign/', $response)) {	// We're trying to download calls for a station we're not granted. Disable Clublog-Transfer for that station(s)
 					$this->disable_sync4call($station_row->station_callsign, $station_row->station_ids);
```
HB9HIL, line 166: why not using 
```
$this->disable_sync4call($station_row->station_callsign, $station_row->station_ids);
``` 
like again as used in the invalid callsign section?
`$station_row->station_id` does not work here.

HB9HIL (changes requested):

application/models/Clublog_model.php
int2001: It's another UseCase. This one targets invalid-loginData, which is set at station_location-level.

The other one references to an invalid call...

But will double-check again

PR #450, opened 2024-06-06
config-switch to disable qsl-image-feature

int2001:
Sometimes you don't want to have QSL-Image-Management at your instance (Various reasons).

This one implements a switch for that.
if absent or setted to false, everything works as before.

if set to true (config.php // `$config['disable_qsl'] = true;` the functions for QSL-Image-Management are disabled for the instance.

The Switch/PR has no real need for standard-installations. But since i built it for another instance i thought: y not sharing?


HB9HIL (approved):
Tested. Works as expected.

if a QSO already has a QSL image this will be shown anyway. (Fine in my eyes)

PR #451, opened 2024-06-06
[LANGUAGE] Introducing gettext-like translations and po files

HB9HIL:
To improve the ability translating Wavelog this PR brings po files and gettext-like functions into the code-base.

This is the first step of switching to the new enginge.

1. This PR. Adding the modified gettext hook, renamed users 'language' column and added basic po files. For a better feeling and see how it works currently only the dashboard or switched to the new function. 

Instead using `<?php echo lang('arraykey'); ?>` we now can directly use `<?= __("Text"); ?>` 
(Note: it's double underline here instead single underline, see below) 
without the need to create a new $lang key in a language file. The po-files will be created fully automatic by a translation project at [crowdin.com](https://crowdin.com/) and the included github workflows (currently only manually, to get an idea).

Next Steps will be: 

2. Replace ALL `echo lang('array_key')` with `__("Text")` and build the pot/po files from it. For this I'm currently building a script which will do that for me. You can find the script [here](https://gist.github.com/HB9HIL/e22d0f93c9dd8b5ea76da3cc8a366d89).

3. Setting up a public translationproject at Crowdin.com (free) to get help from crowd to erasily translate Wavelog. Crowdin.com has an Github Integration which keeps a dedicated translation branch with crodwin synced.

Why "gettext-like":

CodeIgniter and gettext are no real friends. But with this [library](https://github.com/bkader/ci-gettext-hook) hook it works. It emulates gettext and uses a defined local (in our case session-data/cookie). Each gettext function has in this case an underline `_` as prefix. 

instead `_("Text")` we use `__("Text")` (`_` is an alias for `gettext`)
instead `ngettext("Singular", "Plural", integer)` we use `_ngettext("Singular", "Plural", integer)`
insteadt `pgettext("Context Description", "Text")` we use `_pgettext("Context Description", "Text")`

### Bonus: 
We don't want Links in the translations. We can use placeholders instead.

Easy Links: 
 ```php
 sprintf(
      __("You need to update country files! Click %s to do it."),
      "<a href='" . site_url('update') . "'>" . __("here") . "</a>"
);
```

Digits (here in combination with Singular/Plural Detection)
```php
sprintf(
     _ngettext("You have had %d QSO today", "You have had %d QSOs today", intval($todays_qsos)), 
      intval($todays_qsos)
);
```


HB9HIL:
Hint: 

This PR is big. I tried to reduce the commits and made them in a clear structure. So it's easier here to review the 10 commits one by one.

int2001:
just create a "Feature-Branch" out of it.
it's much to huge to review / test (for me - don't know what @phl0 or @AndreasK79 say).

i'll switch to that branch from time to time to see what happens

HB9HIL:
Closed due the fact that there is an `alpha` branch now

PR #463, opened 2024-06-10
[LANGUAGE] Introducing gettext()

HB9HIL:
Details see #451


int2001:
Getting better and better.
I'd like to run it at least until Thursday evening. So far only little problems, which were instantly fixed by @HB9HIL .Good job.

If there isn't any (real) Problem after Thursday, we can possibly merge it?!

phl0 (approved):
Ok with me. Did test alpha branch in prod system without issues. Translations for newly added stuff works. Need to check weblate etc. however. But good to merge imho.

int2001 (approved):
Fine for me as well.
(As told on tuesday)

PR #464, opened 2024-06-11
Fixes showing other stations QSL-Requests

int2001:
Missing bracket at OR-Statement causes other people QSOs showing up in the List of requested QSOs.

e.g.

DL1ABC marked his QSO with DL2ABC as queued for printing
DL1XYZ made and logged (within his own account) a QSO with DL2ABC.

When calling "Labels -> View QSOs -> QSO-List" as DL1ABC the QSO from DL1XYZ was shown as well


phl0 (approved):
Bug reproduced. Fix tested successfully. Nice find :)

PR #465, opened 2024-06-11
add config-switch to disable oqrs-feature

dg0tm:
It does the same as #450 but for the OQRS feature.


HB9HIL:
Hello @dg0tm 

Could you please edit this PR and add a Description? Why do we need this option. What did you change and are there any special cases we should take care off?

application/config/config.sample.php
```diff
@@ -695,3 +695,6 @@
 
 # Disables QSL-Image-Feature if true. if non-existant or false the QSL-Feature is enabled
 $config['disable_qsl'] = false;
+
+# Disables OQRS-Feature if true. if non-existant or false the OQRS-Feature is enabled
```
HB9HIL, line 699: This config option should be in `install/config/config.php` aswell

HB9HIL (changes requested):

application/views/interface_assets/header.php
```diff
@@ -394,10 +394,12 @@ function handleKeyPress(event) {
 								}
 
 								$oqrs_requests = $this->oqrs_model->oqrs_requests($location_list);
+								if ($oqrs_requests > 0) {
```
HB9HIL, line 397: I'm not happy with this one. If OQRS is enabled we should show the menu entry for it. So the user knows where to find the requests. Also this does not work as you wrote it. 

```suggestion
							
```

HB9HIL:
Another one noticed while testing.

If a user already had OQRS enabled for a station location it will show the OQRS menu tab in public view. 
You can also hide this based on the config option. This would be more elegant then just redirect to dashboard/user login

Check `application/views/visitor/layout/header.php` Line `75`

application/views/options/oqrs.php
```diff
@@ -66,4 +67,5 @@
 		</div>
 	</div>
 
-</div>
\ No newline at end of file
+</div>
+<?php } >
```
HB9HIL, line 71: incomplete PHP closing

```suggestion
<?php } ?>
```

application/views/interface_assets/header.php
dg0tm: okay, what's the point of a menu item that shows nothing but a empty page?
and as you can see, it's an optional commit in this PR.

grafik

HB9HIL: We just discussed that in team. 

We don't want to hide this menu entry if OQRS is enabled, even if there are no requests yet. 
You can hide the menu entry if OQRS is disabled.

HB9HIL: > I am against hiding the menu item. It counts for requests with state < 2, that means menu item will only show when you have new requests. What if you want to check something on old requests?

It shows the menu entry anyway but it shows a small number with new requests if there are more then 2 new

image

Imho it should stay like this. Just hide it if OQRS is disabled @dg0tm

HB9HIL: image

HB9HIL (approved):
Looks good to me now

PR #473, opened 2024-06-14
[QRZ] Fix QSL downloads 

HB9HIL:
In a specific case the QRZ QSLs are not correctly downloaded. 

If the QSO time match EXACTLY with the time on QRZ.com the QSL will be downloaded successfully. But if the QSO time differs a few minutes the `import_check()` method in the `Logbook_model.php` Line 3205 says "FOUND" but the QSL does not get updated because in the `qrz_update()` method in Line 3254 we search for the QSO again without the 15 minute intervall as in the `import_check()`. This results in a wrong information. 

Import_check says QSO record FOUND, but the QSO does not get updated in the database because the time does not match exactly. 

I fixed this with the QSO ID. We get this QSO ID back in the import check so we can reuse it in the qrz update function to update the correct QSO without searching for it again.


int2001 (approved):
nice one.

PR #474, opened 2024-06-14
[OQRS] Fix saving Station Locations

HB9HIL:
Due to the changes in #465 saving station locations is broken, if OQRS is disabled with the new config switch. This fixes this.


int2001 (approved):
failsafe for OQRS. Good find

PR #462, opened 2024-06-10
RDA support for ADIF export

R1BLH:
Added export MY_CNTY field to the ADIF file for the RDA data.

PR #485, opened 2024-06-24
Revert "RDA support for ADIF export"

phl0:
Reverts wavelog/wavelog#462

Sri this patch does not help because we currently do not support entering county for DXCCs other than U.S., Hawaii and Alaska. This would have to be implemented first.

### DIFF
--- a/application/libraries/AdifHelper.php
+++ b/application/libraries/AdifHelper.php
@@ -198,30 +198,12 @@ class AdifHelper {
 
         $line .= $this->getAdifFieldLine("MY_STATE", $qso->state);
 
-        // We fill county only if it has a value and it's USA, Alaska, Hawaii or Russia (European or Asiaric). Other countrys are not supported at the moment due complex adif specs
-
-	if ($qso->station_cnty) {
-		switch ($qso->station_dxcc) {
-		    case '291':
-		    case '6':
-		    case '110':
-			$county = trim($qso->state) . "," . trim($qso->station_cnty);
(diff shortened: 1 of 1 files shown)


R1BLH:
This is unfortunate, because in my Wavelog instance I filled the relevant data in the database with a SQL query and everything was exported correctly to the e-QSL web service currently under development (qsl.su).

phl0:
> I realize that I need an interface part for station location where the user can select the CNTY value from a list to store in the database, as well as the ability to enter data for the CNTY field during QSO entry. Unfortunately, this is beyond my skills (php is not a programming language I use every day).

I re-added your PR as https://github.com/wavelog/wavelog/pull/493. But will need to find a solution to make these fields available to fill in the GUI.

phl0:
We were having a discussion about this lately and decided to export the county in general if it is set. So this should fix your issue.
The issue that remains is: Adding GUI support to configure counties for DXCCs other than U.S., Hawaii and Canada. Should be handled separately.

R1BLH:
> The issue that remains is: Adding GUI support to configure counties for DXCCs other than U.S., Hawaii and Canada. Should be handled separately.

Should i open this issue?

phl0:
> Should i open this issue?

Oh yes please :)

R1BLH:
> > Should i open this issue?
> 
> Oh yes please :)

Done
https://github.com/wavelog/wavelog/issues/654

PR #486, opened 2024-06-25
Adding FKs to the Tables

int2001:


HB9HIL (approved):
tested up and down a hundred times while testing #487 :D 

works

HB9HIL (approved):

PR #487, opened 2024-06-25
Migration Lock

HB9HIL:
In big environments or on really slow machines it can happen that a migration is called twice. This causes bad error messages and should not happen. Therefore we create a lockfile `application/cache/.migration_running` to avoid that. Migration gets locked when running and unlocked when done. 

Please test and break. For me this works for `up()` and `down()` functions. 

To test add `sleep(10)` to the latest migration and test up and down by changing `application/config/migration.php`. You need to use two seperate browsers to simulate different sessions.

### DIFF
--- a/system/libraries/Migration.php
+++ b/system/libraries/Migration.php
@@ -321,9 +321,6 @@ class CI_Migration {
 					$this->_update_version($current_version);
 				}
 
-				// After the migrations we can remove the lockfile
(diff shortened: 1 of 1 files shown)


HB9HIL:
tested with #486. works as expected

int2001 (approved):
works.
great improvement for a very hidden bug ;)

PR #492, opened 2024-06-25
"Operator Dialog" for Amateur Radio Club's and Special Callsign Activities

HB9HIL:
Currently it's not possible in Wavelog to share Logbooks between users. This has serveral reasons and we have this in the backlog. To provide a "minimum-solution" for radio clubs and special callsign activities (SCA) just yet, we now can provide the "Operator Dialog" to suit their needs a bit more.

### How
The recommended setup for radio clubs and SCA is to have ONE account which contains the special/club callsign and sharing credentials between the operators. The issue here is that the most operators do not change the operator callsign in the QSO logging. This ends up in "no one knows, who did which QSO" and is imho just bad pratice. To force the operators to provide their callsign I implemented the "Operator Dialog". This dialog, once enabled with a config switch, pops up ONCE after login of a regular operator. Admin's do not get nagged by this dialog. The operator then have to provide his personal callsign which gets written in the `session_data`. Due the fact that this is session data multiple operators can login from different computers at the same time and log QSO's with the special/club callsign -> and with the correct operator callsign. 

In addition there is a second config switch which hides the user menu (and blocks path `user/edit/..` to add a fence for operators to not be able to change account data. This is default enabled but only takes action if the `special callsign` switch is enabled aswell.

TODO: Write a Wiki for amateur radio clubs and SCA for a recommended setup. Link already created [here](https://github.com/wavelog/wavelog/wiki/Recommended-Setup-for-Special-Callsigns-and-Clubs).

PS: This PR does not contain updated po files. I will do this after merge seperately.

image


R1BLH:
This is a really cool feature. I'm thinking about a portable wavelog server for SCA activities in areas without Internet coverage. There could be two or more operator positions there.

phl0 (approved):
Tests successful. Looks to be a good feature for club and special event stations. :)

PR #498, opened 2024-06-26
Slash zeros

phl0:
Slashed some more zeros and fixed a bug re table formatting (though station callsign should always be preset)

Screenshot from 2024-06-26 12-14-11


phl0:
Based off wrong branch

PR #505, opened 2024-06-27
Placed CQ Zone markers in it's own file. Remove unused JS loading

AndreasK79:
This change is for the maps in the advanced logbook, cq award and exportmap.


HB9HIL (approved):
great to have that fixed :)

Issue #339, opened 2024-05-05
[C -> M] Code Cleanup (Progress Tracker)


HB9HIL:
This issue tracks the progress of the 'C -> M' movement. 

Target: 
We want to move all cron functions which are currently located in the controllers move to the/a model. This allows us to call the functions from the 'Cron Manager' directly in PHP with CodeIgniter instead to run a curl, to run a curl, to run a curl....

- [x] Move Function 'clublog/upload'
    - [x] Move Code 
    - [x] Test
    - [x] Check all Functions in WebUI working as before
    - [x] PR: https://github.com/wavelog/wavelog/pull/408
- [x] Move Function 'eqsl/sync'
    - [x] Move Code 
    - [x] Test
    - [x] Check all Functions in WebUI working as before
    - [x] PR: #445 
- [x] Move Function 'hrdlog/upload'
    - [x] Move Code 
    - [x] Test
    - [x] Check all Functions in WebUI working as before
    - [x] PR: #509 
- [ ] Move Function 'lotw/lotw_upload'
    - [ ] Move Code 
    - [ ] Test
    - [ ] Check all Functions in WebUI working as before
    - [x] PR: #536 
- [ ] Move Function 'qrz/download'
    - [ ] Move Code 
    - [ ] Test
    - [ ] Check all Functions in WebUI working as before
    - [ ] PR:
- [ ] Move Function 'qrz/upload'
    - [ ] Move Code 
    - [ ] Test
    - [ ] Check all Functions in WebUI working as before
    - [ ] PR:
- [ ] Move Function 'update/dxcc'
    - [ ] Move Code 
    - [ ] Test
    - [ ] Check all Functions in WebUI working as before
    - [ ] PR:
- [x] Move Function 'update/lotw_users'
    - [x] Move Code 
    - [x] Test
    - [x] Check all Functions in WebUI working as before
    - [x] PR: https://github.com/wavelog/wavelog/pull/338
- [x] Move Function 'update/update_clublog_scp'
    - [x] Move Code 
    - [x] Test
    - [x] Check all Functions in WebUI working as before
    - [x] PR: https://github.com/wavelog/wavelog/pull/338
- [x] Move Function 'update/update_dok'
    - [x] Move Code 
    - [x] Test
    - [x] CCheck all Functions in WebUI working as before
    - [x] PR: https://github.com/wavelog/wavelog/pull/338
- [x] Move Function 'update/update_pota'
    - [x] Move Code 
    - [x] Test
[...]

PR #509, opened 2024-06-27
[C2M] Moved HRDlog Upload functions from Controller to a Model

HB9HIL:
Refering to #339


int2001 (approved):
Did not test, but read carefully. Should work.

Testing was impossible/hard because of real-data and so on. Will check after merging because i found another issue within the hrdlog-controller.

PR #510, opened 2024-06-28
[Advanced Logbook] Multi select is now available in the location dropdown

AndreasK79:
Thanks @HB9HIL for the idea and the initial code.
Got backend working.

I hope I have changed all the places necessary.


HB9HIL (approved):
Tested LBA, Map and played around a little :) I'll take care of translations in JS to a later point. This is just awesome we have multiselect now. Thank you @AndreasK79 :)

PR #512, opened 2024-06-28
Add filter for contests

derFogel:
As the title says. This adds a filter for the advanced logbook that allows you to search by contest.

I know translations are missing, but I don't know how to create them.


HB9HIL:
Hello @derFogel 

For Information how we translate things in Wavelog you can refer to this Wiki:
https://github.com/wavelog/wavelog/wiki/Translations

Use `__("Contest")` 

not `lang('gen_hamradio_contest')`


You don't have to do more. The translation process takes care of the rest.

application/views/logbookadvanced/index.php
```diff
@@ -276,6 +276,10 @@
                         <label class="form-label" for="operator"><?= __("Operator"); ?></label>
                         <input type="text" name="operator" id="operator" class="form-control form-control-sm" value="">
                     </div>
+                    </div>
+                         <div class="mb-3 col-lg-2 col-md-2 col-sm-3 col-xl">
+                         <label class="form-label" for="contest"><?php echo lang('gen_hamradio_contest'); ?></label>
```
HB9HIL, line 281:
```suggestion
                         <label class="form-label" for="contest"><?= __("Contest"); ?></label>
```

HB9HIL (changes requested):
@AndreasK79 should check this aswell. Please change this line as mentioned.

AndreasK79:
I had plans for adding this filter, however with a dropdown with contests from the database, so that you don't have use free text to search. I believe that would be a better approach.

AndreasK79:
Just to add to it, I see you use wildcard and like in the query, so perhaps this is sufficient. 
Any objections to contest filtering this way @int2001 and @phl0 ?

HB9HIL:
> I had plans for adding this filter, however with a dropdown with contests from the database, so that you don't have use free text to search. I believe that would be a better approach.

Maybe just contests which appear at least once in QSO table?

AndreasK79:
I found another problem with this approach. Since the Advanced Logbook uses the existing contest table, a contest from that table needs to be set, to be displayed in the table, If not, the column will be blank.

In other words, if you batch edit qsos and set a contest that is not in this table, it will be a problem.

I think the solution would be to use a dropdown filled from the existing contest table.

int2001:
> I think the solution would be to use a dropdown filled from the existing contest table.

... which will cause a problem if the Op imports ADIFs with contests which aren't at that table.
A full-scan on every LBA-Load (for filling the dropdown with distinct contest-ids) can also not be the solution. it'll slow down everything in extreme (doesn't scale on huge logbooks).

perhaps someone has an idea to solve this problem.

derFogel:
I fixed a bug, were sometimes the map/globe wasn't showing the stations.

I like the idea with the dropdown. To deal the problem of missing contest ids, there could be an option in the admin menu to create missing contests from the qso database.

AndreasK79 (approved):

HB9HIL (approved):
Fine with me now

phl0 (approved):
Works fine for me.

AndreasK79:
@derFogel we did some small modifications, but agreed on it in the end. Merge time :)

PR #515, opened 2024-06-28
QSO Edit - Callbook update for a single QSO

HB9HIL:
Requested in #459 

This has some potential for improvement but it's a good starter. 

#### Why I did not use the function from LBA as it was mentioned in the Feature Request:
The LBA function does update the database directly. I think this is not optimal in this situation. I implemented the button only in the edit_ajax view yet, so only if you open the QSO edit (and not in QSO details aswell). 

I want to see the changes which where made before I click save.


dg9vh:
Tested with result: working :-)

int2001 (approved):
works. well done!

phl0:
Works (also with hamqth).

PR #517, opened 2024-06-29
[QSO] sanitize some input data

### LINKED ISSUE #516: QSO: Hardening callsign and locator input against unnecessary spaces at the beginning or end of input
In Cloudlog and Wavelog, from time to time I'm suffering from QSOs not being saved, because I have entered unnecessary spaces at the beginning or end of the callsign or gridlocator.

This happens frequently when copying for example from QRZ and pasting into the text fields.

If then you try to save the QSO, nothing happens, not even an error message. I had been using Cloudlog for half a year before I found out the spaces are the problem. Before I found out i thought both systems were just buggy or a session had run out of time.

I can imagine other users are also suffering from that problem.

I suggest to just ignore spaces at the beginning of the input string or trailing spaces at the end in callsign and gridlocator field, because both inputs should never include spaces.

HB9HIL:
While waiting for the ferry after the HAMradio expo I fixed #516 

Please test and break


int2001 (approved):
nice one

Issue #418, opened 2024-05-27
Add the ability to set a default radio used for live QSO


zone11:
**What Feature is Missing?**
I would recommend adding the ability to define a default radio (from the hardware interface list) to be used on the "Live QSO" page. Alternative would be storing the last selected radio within the user.

**Why is This Feature Important?**
Using hardware interfaces is an awesome feature, but every time you navigate around Wavelog, coming back to add a live QSO, you need to set the radio first. The feature would save time getting ready, especially in hectic moments.

**Is This Feature Personal or Beneficial to Others as Well?**
It's definitely a benefit for every user of a hardware interfaces in Wavelog

wafer-li:
And maybe adding the "default ANTENNA" as well?

HB9HIL:
@zone11 

Started to work on this. PR #700 is still drafted but feel free to test.

zone11:
@HB9HIL I've done a quick functional test, works fine and the implementation is very user friendly. Tnx!

HB9HIL:
The "default radio" handler is now available in `dev` branch and will be published in the next release.

zone11:
Thank you for the implementation!

PR #526, opened 2024-07-01
[QSO] Skip Radio Cookie

HB9HIL:
Refering also #418 

When logging a QSO in the WebUI with a selected Radio, then this radio ID is written together with some other stuff in a Cookie AND a session. This session data gets written in a cookie which gets written in a session data again to reselect this radio when coming back to the QSO view. However.. This is doubled and produces overhead.. 

Unfortunately this feature of reselecting the radio was broken when using pretty URL's (removing `index.php` from the URL) because the cookie was set with the wrong cookie path. 
cookie path

Instead fixing the cookie path we can skip the whole cookie thing and write the radio_id directly to the session in PHP to manage this info on server side. Now the reselecting of the Radio in QSO view works again, no matter if pretty URL is used or not.


int2001 (approved):
simplify code is always a good idea.
tested it: works

PR #533, opened 2024-07-02
[Station setup] Shows if location belongs to active logbook

AndreasK79:
Added a new column in location table to display if the location is linked to the active logbook:
image


HB9HIL (approved):

int2001 (approved):
works great.
perhaps you can add a reload to the station-page as well?

After linking/unlinking the mainpage (with the checkmarks) isn't refreshed. Nothing big - can be done in another PR later.

int2001 (approved):
Yeah! Now with reload. i like

PR #536, opened 2024-07-02
[C2M] LotW Part 1

HB9HIL:
refering to #339


HB9HIL:
Due the fact that moving the LotW stuff will be a bigger action, I split it up in more then one PR to make it reviewable.

This is Part 1

- Renamed LotwCert to Lotw_model to match our other model names 
- removed deprecated openssl warning (in our min. PHP Version 7.4 is openssl already included)

phl0 (approved):
This works as expected.

PR #538, opened 2024-07-03
[IOTA Award] Fix second table loading, and add PK to speed up

AndreasK79:
I noticed that the IOTA award was terribly slow. Tracing this down to adding a primary key on the iota table.

While investigating, I also noticed that if you load the table a second time, confirmed contacts were not counted.

Thanks to @int2001 for db help.


int2001 (approved):
Works and speeds up IOTA. Great

PR #539, opened 2024-07-03
Fix datatables lang

HB9HIL:
fixes an error in browser console, when user language is english. Now proper language loading for datatables. 

image

also replaces all `\"` with `'` in the po files. We really should not use double quotes in the translations. 

Reason is that we often use js vars enclosured with double quotes. There is no logic yet which escapes the double quotes twice to make that work. To avoid double quotes in translations is easier then avoiding single quotes as they are more often used in different languages.


int2001 (approved):
works. can't spot an issue here.
opposite happens: Less issues in JS :)

Well done.

Issue #452, opened 2024-06-06
Callsign Lookup - No informations if a suffix like /P or a prefix is added to the Call


zone11:
**Describe the bug**
When adding a QSO (Live or Post-QSO) with a callsign using a suffix like /P, no informations will be added to the QSO. The same applies to prefixes for different countries. The fallback to the callsign without is made, but only shown as information.

**Version Information**
- Wavelog:          1.6 

**Steps to Reproduce**
1. Go to QSO -> Live QSO
2. Click in the callsign field and add CALL/P (e.g. HB9HJQ/P and LA/HB9HJQ)
3. Check if the name is added.
4. Check the informations in the box on the right below the map. 

**Expected Behavior**
If no direct match for CALL/SUFFIX or PREFIX/CALL is found, use the informations for CALL

**Desktop Information:**
- OS:               MacOS
- Browser:          Safari
- Version:          17.5

**Screenshot**
![Bildschirmfoto 2024-06-06 um 14 42 53](https://github.com/wavelog/wavelog/assets/1774084/61b94ad5-a813-4ad9-8906-bf9ca481be09)
![Bildschirmfoto 2024-06-06 um 14 47 29](https://github.com/wavelog/wavelog/assets/1774084/4e301c67-f204-4a65-a57d-0c5efa53fbe1)

phl0:
Difficult question. IMHO except for the name none of the normally fetched information is usable if you enter a prefixed or suffixed call (i.e. DXCC suffix). So I guess it would only make sense to pull the name from the logbook?

zone11:
Good point, location data is not valid in both cases and could be ignored. The name would definitely make sense.

DK9JC:
i would also very much welcome this. the name based on the lookup at qrz etc. and as a backup the name you already have in the log with "homecall". QSL route - as already written - would also be good.

m0urs:
Das hätte ich auch sehr gerne. Das Problem hatten wir bei Cloudlog schon mal diskutiert. Und dort wurde es so implementiert:

Wenn bei QRZ.com ein entsprechender Eintrag existiert, dann werden die dort hinterlegte Daten aus QRZ.com übernommen.
[...]

PR #541, opened 2024-07-05
Callsign Lookup - No information if a suffix like /P or a prefix is added to the Call

m0urs:
Should address issue https://github.com/wavelog/wavelog/issues/452

### DIFF
--- a/application/models/Logbook_model.php
+++ b/application/models/Logbook_model.php
@@ -4703,6 +4703,24 @@ function lotw_last_qsl_date($user_id) {
       }
     }
 
+    function get_plaincall($callsign) {
+		$split_callsign=explode('/',$callsign);
+		if (count($split_callsign)==1) {				// case F0ABC --> return cel 0 //
+			$lookupcall = $split_callsign[0];
+		} else if (count($split_callsign)==3) {			// case EA/F0ABC/P --> return cel 1 //
+			$lookupcall = $split_callsign[1];
+		} else {										// case F0ABC/P --> return cel 0 OR  case EA/FOABC --> retunr 1  (normaly not exist) //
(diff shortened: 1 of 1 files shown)


HB9HIL:
Hello @m0urs 

unfortunately this approach won't work in a good way. 

1. Please check `application/controllers/Logbook.php` Line 172. The method `get_plaincall()` already exists there. 
2. The reason why no QRZ comes back is not the plaincall. Reason is that we need to escape `/` in the api call

`application/libraries/Qrz.php`
```php
public function search($callsign, $key, $use_fullname = false) {
		$data = null;
		try {
			// URL to the XML Source
			$xml_feed_url = 'http://xmldata.qrz.com/xml/current/?s=' . $key . ';callsign=' . $callsign . '';
```
```php
$callsign = str_replace("/", "&#47", $callsign);
```

3. But even then we should limit which fields get written with callbook info if the callsign contains a suffix or prefix, because things like grid, dxcc are not valid then. (probably only "name" and "qsl via" is valid in these cases)

m0urs:
Sorry,  I am not really a PHP developer.

I just tried to re-implement what has been done in Cloudlog to solve the same issue some time ago. In my own fork the above approach does work correctly. Even the correct DXCC is kept so if I use e.g. "8S/DG3NAB/M" the DXCC is correctly set to "Sweden".

However, you are right about the wrong QTH and grid in this case. Only name and picture are always valid.

I just did a quick try and added the line 

$callsign = str_replace("/", "&#47", $callsign);

before the "$xml_feed_url = " line in the qrz.php. However this did not work for me. I still do not get any information from QRZ for a compound call sign.

So you could reject my change request and someone who has more knowledge needs to check the correct way to implement this. In the meantime I will work with this solution in my own fork.

PR #543, opened 2024-07-05
[QRZ & HamQTH] Fix Callbook Lookup

HB9HIL:
QRZ returns callbook information in this manner

Search for CALL/P
if /P does not exist return just info for CALL
if /P exist return this data

To escape the slash fixes this issue (referred in #452 )

Problem: We need to discuss if ....

1. We just update all fields even if we know that things like `Location (Grid)` are not valid

or

2. We just update certain fields in QSO logging if the callsign contains a suffix or prefix

In both cases the user needs to check the data by hand. Imho we can use 1. option but we can discuss this and I'm open for other arguments. 

This also affects #459  and #515. For this reason we should implement the decision in a central place (in the QRZ library functions)

### DIFF
--- a/application/controllers/Logbook.php
+++ b/application/controllers/Logbook.php
@@ -123,7 +123,7 @@ class Logbook extends CI_Controller {
 
 		$return['dxcc'] = $this->dxcheck($callsign,$date);
 
-		$lookupcall=$this->get_plaincall($callsign);
+		$lookupcall=$this->logbook_model->get_plaincall($callsign);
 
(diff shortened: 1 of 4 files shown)


HB9HIL:
did the same for HamQTH to be complete

HB9HIL:
there is some more to do here. QRX

int2001 (approved):
tested again with some call i never worked.

does a fallback to maincall if extended Call (additional `/P`) does not exists.

works like expected

Issue #165, opened 2024-02-22
Improve Frequency input


205er:
Hi all,

When displaying a QSO, the frequency will be shown in the following way: 21.300 - Frequency in MHz with dot as decimal separator and three digits as decimal. Unfortunately, you cannot enter a frequency in this format, but it must be entered in Herz. To stick to my example: 21300000 must be entered! If you enter 21.300, the frequency will not be overtaken correctly, but set to the value 21 and displayed as 0.000

Would it be possible to improve this behavior or at least implement a check, so that the input does not get lost?

![Freq](https://github.com/wavelog/wavelog/assets/14996169/75242b87-46ce-45da-be13-53cb0d35a169)
![Freq](https://github.com/wavelog/wavelog/assets/14996169/d62b5053-0f9b-46c0-867e-8d77799fefae)

73

HB9HIL:
Hello @205er 

this is a fair point. We may add a User Option for it so the user can decide which format he wants to use 👍

205er:
Thanks for your quick and positive response.
At least, it would be greate to have the input checked (ideally directly when leaving this field and not only when the QSO will be saved). This would help to prevent input mistakes.
Making the input user defined would be wonderful! So to say the cherry on the cake 😄

dg9vh:
Good idea... it could be checked against the band-value (if selected) or against some bandplans that in ideal world could be customized by the users with it's boundaries.

dg9vh:
We may not need to have them customizable... was a wrong bias in brain :-)

205er:
Today i tried out SFLE, which is an fantastic feature! Frequency input in MHz with decimal separator "." is possible (e.g. 21.304) there.
The standard QSO input will not accept this freq. format. :-( Additionally, the band indicator is not logically linked to the frequency, which means you could enter "21,304 MHz" but the band is still set to "80m". When having some pileup, this would easily lead to inconsistency in the heat of the moment.

HB9HIL:
This will take a bit longer since it's quite complex in the qso logging and I need time to dig into this. Please have some patience

HB9HIL:
Solution is merged to dev branch.

PR #544, opened 2024-07-05
[QRG] Allow QRG unit selection per band

HB9HIL:
This is a first step to improve the frequency input. Ref. #165 

This adds another column to the bands table and allows choosing another frequency unit then the defaults. 

To keep performance up this qrg units are written to the session data and the user_options. When logging in available qrgunits are written to the session data. This makes sure we don't call the database too much. 

next step will be displaying the frequency's based on the qrgunit 

Comment:
Why not starting with the minimum approach here and just improving the frequency input in the QSO logging? Well the frequency displays in Wavelog are either in Hz, which is really uncommon and not really handy or they are inconsistent. 
Goal is to get a clean and intuitive WebUI, which is consistent, stable and fast at the same time. This includes imho a consistent way to display frequencies. Just one unit is not enough as in the hamradio world we have a lot of different bands and spectrum areas.

Hint: Ignore whitespace in this PR


HB9HIL:
Hint:

Place this at the end of `application/views/interface_assets/footer.php` to easy display the session data

```html
  <pre><?php print_r($this->session->userdata()); ?></pre>
```

int2001 (approved):
Does actually nothing (as described) which affects the User.
Can't spot a problem so far.

tbh: scrolled through code and tested the Band-UI in combination with QSO. Not more.

PR #545, opened 2024-07-06
Fixes broken detail-view

int2001:
Detailview was broken, because the header for "Station Callsign" was tried to be setted by a dependency which could never be fullfilled (at the time the dep for `isset($station_callsign)` was asked that var wasn't filled).

This one heals the detail-view and displays the station_call always here


phl0 (approved):
Fixes as designed. Go for merge!

PR #547, opened 2024-07-09
[INSTALLER] Multilanguage Support for the Installer

HB9HIL:
The first impression counts

Therefore it's just logical to support gettext and mutli languages in the installer aswell. As the installer is in the most parts independent from the main application I included the gettext lib in the installer aswell and adjusted it a bit. 

The logic gets the default browser language and sets a cookie from it. So the user will directly see his native language (if everything is translated). He is still able to choose another language with a little button on the first page. 

I will include the installer language as another component to our [weblate platform](https://translate.wavelog.org/engage/wavelog/)


int2001:
Tested with docker.

1st issue: Click on Language (at first page) and chose german: Redirects ultimate to `http://127.0.0.1:8086/lang=de_DE/install/install/install/install/install/install/install/install/install/install/install/install/install/install/install/install/install/install/install/install/install/install/install/install/install/install/install/install/install/install/install/install/install/install/install/install/install/install/install/install/install/install/install/install/install/install/install/install/install/install/install/install/install/install/install/install/install/` and so on.
2nd issue: on first call it shows `Warning: Undefined array key "install_lang" in /var/www/html/install/index.php on line 104`
3rd issue: Continued with "english" as default-lang. When setup/migrations were finsished i see a german login-screen. after logging in my language is setted to Bulgarian

int2001:
Issue 1 and 3 are fixed with your latest commit.
issue 2 still there.

Repro for #2: Delete Cookies or use incognito-mode for calling the installer

int2001 (approved):
works now as expected.

Issue #549, opened 2024-07-09
LoTW upload failed: Signing failed


violarulan:
**Describe the bug**

My LoTW account won't update any new QSOs.

Cron works normally. If I click 'Manual Sync', it shows 'Signing failed.'.

**Version Information**

- OS:               Rocky Linux 9.2 Apache/2.4.57 (Rocky Linux) OpenSSL/3.0.7
- PHP:             8.2.13 
- MySQL/MariaDB:   10.5.22-MariaDB
- Wavelog:          1.6
- tQSL        2.7.3

**Additional Context**

I've uploaded tQSL to latest version (2.7.3), re-exported my callsign certificate and uploaded it to Wavelog. It still shows the same.

After some searching, the message is at 

https://github.com/wavelog/wavelog/blob/d846bc153b90245010700a4ba0c426f2fdc8214f/application/controllers/Lotw.php#L275

If I'm right, Wavelog searches the ADIF file, if it contains `<SIGN_LOTW_V2.0:1:6>`, it stops uploading.

I print the `$adif_to_save` variable in the context and found every QSOs in the file contains `<SIGN_LOTW_V2.0:1:6>`.

```
// $adif_to_save variable
<TQSL_IDENT:54>TQSL V2.5.4 Lib: V2.5 Config: V11.12 AllowDupes: false

<Rec_Type:5>tCERT
<CERT_UID:1>1
<CERTIFICATE:1654>MIIEwDCCA6igAwIBAgIDCse1...............QEBCwUAMIHYMQswCQYDVQQGEwJV
.....
<EOR>

<Rec_Type:8>tSTATION
<STATION_UID:1>1
<CERT_UID:1>1
<CALL:6>BG2ELG
<DXCC:3>318
<GRIDSQUARE:6>PN17wd
<ITUZ:2>33
<CQZ:2>24
<IOTA:0>



<EOR>

<Rec_Type:8>tCONTACT
<STATION_UID:1>1
<CALL:6>YE1...
<BAND:3>20M
<MODE:3>FT8
<FREQ:9>14.076256
<FREQ_RX:1>0



<QSO_DATE:10>2024-07-09
<QSO_TIME:9>09:44:30Z
<SIGN_LOTW_V2.0:1:6>
<SIGNDATA:51>24PN17WD.....YE1...14.0762560FT82024-07-0909:44:30Z
<EOR>

<Rec_Type:8>tCONTACT
<STATION_UID:1>1
<CALL:6>VK7...
[...]

PR #552, opened 2024-07-09
Add else clause in case of openssl sign failure

violarulan:
#549 

In some distro (like RHELs), SHA-1 has been disabled by default. 
To avoid that a NULL is returned and nothing happens, this else clause is added to log the error.


phl0 (approved):
We will take that for now and try to make LoTW use more recent hash algos.

PR #555, opened 2024-07-10
SecFix: Some eqsl-functions could even be called without authentication

int2001:


HB9HIL (approved):
did not test but code looks good to me, can't spot a function which is currently used by a cron.

PR #557, opened 2024-07-10
[DEBUG] Added permission check for cache folder

HB9HIL:
The migrations use a locking file, which needs to be written into the caching folder `application/cache`

This adds a check in debug view


phl0 (approved):
w0rx

PR #561, opened 2024-07-11
Cookie Encryption

HB9HIL:
save encrypting cookies to keep me logged in


int2001 (changes requested):
It's working. Well done.

Some minor (but important) things, as discussed:
- Disable Login via remember_me Cookie, if secret wasn't altered.
- pse set httpOnly as well at the cookie-options.

application/controllers/User.php
```diff
@@ -798,7 +868,23 @@ function login() {
...
+				// Create a keep_login cookie
+				if ($this->input->post('keep_login') == '1') {
+
+					$encrypted_string = $this->user_model->keep_cookie_hash($data['user']->user_id);
+
+					$cookie = array(
+						'name'   => 'keep_login',
+						'value'  => $encrypted_string,
+						'expire' => '2592000',  // 30 days
+						'secure' => TRUE
```
int2001, line 881: This one will perhaps fail, if WL runs in a non-TLS-Environment.

HB9HIL: not sure. the 'keep logged in' checkbox is hidden if the page is called without https

int2001: fine for me as well. So no "remember me" for HTTP-only users :)

int2001 (approved):
Well done. Can't spot any problems.

tbh: a seconds pair of eyes on this won't hurt. it's a very sensitive area of code...

phl0 (approved):
Works as expected.

PR #568, opened 2024-07-11
Demo

HB9HIL:
demo stuff directly in main code

can be enabled by creating a file `.demo` in root folder


int2001 (approved):
Works so far - can't spot an issue while reading code/testing.

But:
if demo/demo doesn't exist at DB-Level it won't work. A bit missleading imho. What about the idea to simply login the user (demo/demo is already hardcoded) in that case?

HB9HIL:
> Works so far - can't spot an issue while reading code/testing.
> 
> But: if demo/demo doesn't exist at DB-Level it won't work. A bit missleading imho. What about the idea to simply login the user (demo/demo is already hardcoded) in that case?

Added a comment to make it more clear how to use this

PR #570, opened 2024-07-11
Files

HB9HIL:
Currently we have the working files in the for sota, pota, wwff and stuff directly in the repository. They are in use and get updated by the user functions. 

Problem is that they are tracked and the should not be tracked in a git environment. I placed those files in a central directory and they get copied to the updates folder when used for the first time. Now they can get updated without version tracking. 
telegram-cloud-photo-size-4-5800856059626243298-x

Important for this PR: 

All users need to do a `git stash` when updating to the next version. We have to do a proper communication of that. Therefore this PR has a major impact.

I also removed some deprecated and dead Lotw stuff.


AndreasK79 (approved):

AndreasK79:
I like these changes, so that we will not have any problems with these files when updating Wavelog.

Issue #567, opened 2024-07-11
add "Mark printed" in QSL-Label-printing-dialog


dg9vh:
**What Feature is Missing?**
To shorten the workflow of printing QSL-Labels a button "mark printed" within the QSL-Label-printing-dialog, where you finaly start the printing of the labels before you chose the printer to use could potentially shorten the workflow when printing labels.

Actually you have to print the labels and then in a second step you have to switch over to the print queue and click there on "mark all as printed" (or similar). This sometimes has potential to be forgotten so that you sometimes have labels left in the queue that are already printed.

**Why is This Feature Important?**
I think the two factors above are the important things:
- saving time by having less mouseclicks
- saving from forgetting to mark them as printed.

**Is This Feature Personal or Beneficial to Others as Well?**
This feature would be a benefit for all those who are using the label printing functionality :-)

int2001:
Good idea. I am going to add a checkbox "mark as printed" to the printing dialog.

dg9vh:
I think internally you could use already exististing functions for this so that only some UI and command-routing has to be implemented :-)

int2001:
@dg9vh can we close this issue?

dg9vh:
Woeks as designed

PR #572, opened 2024-07-12
Label mark

int2001:
Added the ability to mark QSLs as sent/printed within the printing Dialog, to save a few clicks as requested in #567 

image


phl0 (changes requested):
IMHO we need a hint here. I made the mistake to mark as printed and then wanted to hit print. But the QSOs are marked as printed already. Maybe lock the mark button until print is pressed? Or add a text hint?

AndreasK79 (approved):

phl0 (approved):
This looks fine to me now. :)

HB9HIL (approved):

phl0 (approved):
Also fine with me. Tested and works.

PR #573, opened 2024-07-12
Handle propagation modes not suppored by LoTW (i.e. RPT & INTERNET)

phl0:
Second part to https://github.com/wavelog/wavelog/issues/562. This removes the uploaded mark from existing RPT and INTERNET QSOs as they were marked as uploaded in the past though the entries were not processed by LoTW and can never be confirmed.


phl0:
Besides marking the QSOs as invalid/ignore there has been more code added which:

- Marks logged QSOs upon LoTW upload as invalid/ignore
- Shows ignores LoTW state arrows as grey (like QSL)
- Disables the LoTW fields in QSO edit in case an unsupported prop mode is selected
- Shows a hint if an unsupported prop mode is selected in QSO edit

assets/js/sections/common.js
```diff
@@ -265,6 +265,28 @@ function qso_edit(id) {
...
+                       success: function(data) {
+                          unsupported_lotw_prop_modes = $.parseJSON(data);
+                       },
+                    });
+
+                    $('#prop_mode').change(function(){
+                       if (unsupported_lotw_prop_modes.includes($('#prop_mode').val())) {
+                          $('#lotw_sent').prop('disabled', true);
+                          $('#lotw_rcvd').prop('disabled', true);
+                          $('#lotw_propmode_hint').html("This propagation mode is not supported by LoTW. LoTW sent/received fields disabled.").fadeIn("slow");
```
phl0, line 282: @HB9HIL I need a little support with translations here. Can u pls give a hint on how to integrate into JS?

hb9hil-tester: Just as we did before. Just create a JS var in PHP. Thats it

HB9HIL: Just as we did before. Just create a JS var in PHP. Thats it

phl0: Will do

HB9HIL (commented):
code looks good. did not test yet

HB9HIL (approved):
tested. works
probably we have to change the text of the tooltip for grey arrows some day but not necessary atm 
I like the approach of this PR

phl0:
Let me know whats wrong with the text. Still time to change.

HB9HIL:
> Let me know whats wrong with the text. Still time to change.

imho it's hard to see why in the LotW tab the 'sent' and 'rcvd' dropdowns are disabled, since there is no hint. 

Thoughts:  

>I imagine a user see's grey arrows in the logbook overview
image  

>Clicking on the QSO to see whats wrong
image  

>Mhh noting here. But I remember I can set the states.
image  

>Mhh they are disabled but dunno why, all other QSO's are working. 
....

>it will take some time he sees the little hint at prop mode unless he realizes whats wrong

image

---

so my suggestion would be something like this to make it clear from the beginning why both arrows are grey

image

HB9HIL:
But in fact it has nothing to do with the functionality :) so I approved 👍

int2001 (approved):
works as expected. huge effort - well done

PR #576, opened 2024-07-13
[SFLE] Add the ability to add QSO's to a Contest in SimpleFLE

HB9HIL:
and fixed a bug where the OP call was not saved to database


AndreasK79:
Tested working, both with contest and operator.

AndreasK79 (approved):

PR #577, opened 2024-07-14
fix eqsl

HB9HIL:
Made a mistake during #339 

this fixes eqsl


int2001 (approved):
seems legit.
don't know what happened b4, since the same model was loaded!?

HB9HIL:
> seems legit. don't know what happened b4, since the same model was loaded!?

I moved these function from the controller to the model and forgot to adjust these

PR #579, opened 2024-07-15
Added VUCC to accumulated-stats

int2001:


HB9HIL (approved):
tested :) works. Nice addition

PR #585, opened 2024-07-16
Use fast query if db is supported

AndreasK79:
If MySQL >= 8, use fast query in VUCC accumulated stats.


phl0 (approved):
Major impact. Brings accumulated stats queries from 28000ms to 100ms. I like.

PR #587, opened 2024-07-16
[INSTALLER] Small Enhancements

HB9HIL:
Some small enhancements for the installer 

1. Get rid of the default encryption key by generating a safe key and place it in the config.php 
2. renamed `max_upload_file_size` to  `upload_max_filesize` to match the true PHP option
3. Make sure all mig's are done and show success message (there was a race-condition issue)


HB9HIL:
screwed this one up. will propose a new PR

PR #588, opened 2024-07-16
Accumulated VUCC now also included vucc_grids

AndreasK79:
Now counts border or crossgrids in col_vucc_grids, but only for MySQL >= 8


int2001 (approved):
great improvement.
Multigrids are now counted as well here.

phl0 (approved):
Nice addition. Works on my system(s).

PR #590, opened 2024-07-17
Set default language if detection fails

phl0:
BG6HJE reports that the installer fails with following error message:

q8l-UAHR

So we should set a default language if (auto)detection fails. Waiting for Op to confirm the fix so draft for now.


HB9HIL:
a fix for this issue is included in #603

phl0:
Closing here as covered in https://github.com/wavelog/wavelog/pull/603.

PR #592, opened 2024-07-17
[Contesting} Set userdata radio when logging

AndreasK79:
Since we set radio from user session, it's also good to set radio in user session when logging, just like the regular qso entry.


HB9HIL (approved):
works as expected :) 👍

PR #593, opened 2024-07-18
Improve user experience when logging unknown calls

AndreasK79:
I propose to ignore /D and /X in callsigns.

Also return empty array if DXCC is not identified.


phl0:
Should we set lat and lng to '' / null instead of 0 to prevent mapping of those stations here?

AndreasK79:
@phl0 good point. Had to set them to blank. Null errors out in JS when I test.

application/models/Logbook_model.php
```diff
@@ -4418,7 +4431,7 @@ function wpx($testcall, $i) {
       $c = '';
 
       $lidadditions = '/^QRP$|^LGT$/';
-      $csadditions = '/^T$|^P$|^R$|^A$|^M$|^LH$/';
+      $csadditions = '/^X$^D$^T$|^P$|^R$|^A$|^M$|^LH$/';
```
int2001, line 4434: am i wrong, or are we missing an OR `|` here at the RegEx?

AndreasK79: No, you are absolutely right. Thanks for spotting!

int2001 (approved):
works and improves the experience ;)

PR #598, opened 2024-07-20
[eQSL] Error Handling

HB9HIL:
When eQSL Sync was called by a cronjob and the upload for one user failed due a wrong password or another issue, there was a `redirect` in the code which caused the whole sync process to stop. In result of that all users AFTER the failing one weren't synced. This solves this issue by setting a proper status and breaks the foreach only for this user. By the way I added gettext to the flashmessages.


int2001 (approved):
Nice finding.
Read the code twice but didn't test (because of missing test-env for eQSL).

This looks good and functionable.

PR #602, opened 2024-07-21
Fixed DXCC2-Summary at Add QSO-Window (bottom)

int2001:


int2001:
Seems like the Window at bottom of the QSL-Form (behind the "+") was really the only one with problems.
@phl0 would be nice if you can cnfm

AndreasK79:
Tested working. This PR fixes Quick Lookup as well (since it's the same component).

AndreasK79 (approved):

PR #605, opened 2024-07-22
[Maps] Distance is now shown in the unit set in user settings

AndreasK79:
image


int2001 (approved):
works!
Now its consistent and taked the unit from the setting :)

HB9HIL (approved):
works :) great fix

PR #607, opened 2024-07-22
degoogling

HB9HIL:
While testing the installer upgrade @int2001 noticed that we have some googleapi urls in our code. It's seems they live in the bootstrap 5.3 files. As we do not use the font "Source Sans Pro" we really do not need any google stuff in this software. 

Degoogled


int2001 (approved):
looks good.
tested a few pages (especially those with icons / flags and special-characters). can't spot a problem

PR #609, opened 2024-07-22
Language flags

HB9HIL:
In case you do not speak chinese and you want to set it back to your native language,  you now know which language is yours xD 

image


AndreasK79:
Tested working. No Norwegian flag though ;p

AndreasK79 (approved):

phl0:
Like it and works here.

PR #610, opened 2024-07-22
[Gridmap] Added a trim in case some white space would exist in vucc grids

AndreasK79:


HB9HIL (approved):

phl0:
This works in my test env. A good fix as I had some grids causing errors due to whitespace (imported from another log).

PR #612, opened 2024-07-22
handle special chars in passwords

HB9HIL:
Passwords like `@m4cU;\IZp3+AAa;$%r;.x#p}(<*Hk` can't be handled as in html forms chars like `<` get escaped. 

We need to take care of that otherwise a user is no longer able to login with his set password. 

Thanks to HB9EDI for this bug report.


int2001 (approved):
Tested this absolute edgecase with various passwords.
including from dev to #612 

magically it works. tbh: i am not 1000% sure if there are other cases where a user wont be able to login, because the handling of the string is changing here.

for now i'd approve. but let's keep an eye on that.

PR #616, opened 2024-07-23
[OQRS] iFrame Widget für Wavelog OQRS

HB9HIL:
Simple iFrame Widget which can be placed in a QRZ.com Bio or on any other website which allows iframes

image

To use this widget place it somewhere with this iframe code:

```html
<iframe name="iframe" src="[WAVELOG URL]/widgets/oqrs/[CALLSIGN]" height="220" width="670" frameborder="0"></iframe>
```
> [!IMPORTANT]  
> Due the fact that this HTML is not responsive yet I recommend to stick with `height="220" width="670"` in the iframe. You may have to play a little bit with these settings to find the right spot. 
---
### Options 
The Widget is customizable. Currently available options:

- `theme`
  Choose the theme widget with one of the themes in Wavelog
  Example: `theme=darkly`

- `slug`
  The Widgetlogo is a link which shows per default on Wavelog's Github Repo (https://github.com/wavelog/wavelog). You can set your public view page instead by defining the public slug of the particular logbook. 
Example: `slug=hb9hil`
  
 To add options just place them as GET parameter in the URL like this:
```
[WAVELOG URL]/widgets/oqrs/[CALLSIGN]?theme=darkly&slug=hb9hil
```

### DIFF
--- a/application/controllers/Oqrs.php
+++ b/application/controllers/Oqrs.php
@@ -53,8 +53,20 @@ class Oqrs extends CI_Controller {
 		$this->load->model('oqrs_model');
 		$data['result'] = $this->oqrs_model->getQueryDataGrouped($this->input->post('callsign'));
 		$data['callsign'] = $this->security->xss_clean($this->input->post('callsign'));
+		
+		if($this->input->post('widget') != 'true') {
+			$this->load->view('oqrs/request_grouped', $data);
+		} else {
+			$data['stations'] = $this->oqrs_model->get_oqrs_stations();
+			$data['page_title'] = __("Log Search & OQRS");
+			$data['global_oqrs_text'] = $this->optionslib->get_option('global_oqrs_text');
+			$data['groupedSearch'] = 'on';
+			$data['widget_call'] = true;
 
-		$this->load->view('oqrs/request_grouped', $data);
+			$this->load->view('visitor/layout/header', $data);
+			$this->load->view('oqrs/index');
+			$this->load->view('interface_assets/footer');
+		}
 	}
 
 	public function not_in_log() {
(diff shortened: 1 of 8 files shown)


AndreasK79 (approved):

AndreasK79:
Tested working here when I got the correct url :)

phl0:
Tested and working fine (though I don't have any OQRS enabled)

R1BLH:
I tried to use the widget. I found one annoying moment -- when the light theme is turned on, the background of the call sign input field remains dark, as a result I can't see the text entered there (the text is visible only when you select it).


And if you don't have index.php hidden (php-fpm is used on the server), you need to add index.php to the widget url:
`<iframe name="iframe" src="[WAVELOG URL]/index.php/widgets/oqrs/[CALLSIGN]" height="220" width="670" frameborder="0"></iframe>.`

изображение

HB9HIL:
> I tried to use the widget. I found one annoying moment -- when the light theme is turned on, the background of the call sign input field remains dark, as a result I can't see the text entered there (the text is visible only when you select it).
> 
> And if you don't have index.php hidden (php-fpm is used on the server), you need to add index.php to the widget url: `<iframe name="iframe" src="[WAVELOG URL]/index.php/widgets/oqrs/[CALLSIGN]" height="220" width="670" frameborder="0"></iframe>.`
> 
> изображение

fixed that in a direct commit [here](https://github.com/wavelog/wavelog/commit/69f71aaea7ef1aaf8ef064d546682ecea9f52ba2)

R1BLH:
> fixed that in a direct commit [here](https://github.com/wavelog/wavelog/commit/69f71aaea7ef1aaf8ef064d546682ecea9f52ba2)

It works! Thank you!
изображение

Issue #631, opened 2024-07-25
max_upload_file_size should read upload_max_filesize


nobswolf:
**This is for reporting bugs with the code, if this issue is regarding setting up on your local server please use the discussions area**

**Describe the bug**
the php.ini attribute for the file size for uploads is spelled wrong

**Version Information**
We need some information about your versions. You can get all Information in the Admin > Debug Menu
- OS:                Debian 12
- PHP:              8.2
- MySQL/MariaDB:     10.11
- Wavelog:          1.7 git

**Steps to Reproduce**
1. start installer
2. read about max_upload_file_size 
3. don't find it in php.ini
4. find  upload_max_filesize instead

**Expected Behavior**
correct name in the check

phl0:
Fix proposed in https://github.com/wavelog/wavelog/pull/632.

phl0:
Correction: Will be fixed with https://github.com/wavelog/wavelog/pull/603.

HB9HIL:
@nobswolf 

is fixed in debug view and installer. please check and close this issue :)

PR #603, opened 2024-07-21
[INSTALLER] Next generation of the Wavelog Installer

HB9HIL:
First I just wanted to fix some known bugs in the installer after the big Frontend Remake in #37 

But I wanted to improve the backend aswell and give the user a better feedback during the process as the whole database stuff can take a while. So I rebuilt the backend of the installer and added some frontend stuff for better response to the user. 

Improvements:

- Fixed some bugs regarding language
- Save inputs in localStorage to don't loose all the entered data in case of a page reload
- Finally added the checklist on the last page
- Added a second page which shows the current step of the installation for better UX

Work in Progress:

- [x] Readd POST Validation for server side validation
- [x] Better Error Handling if one install step fails
- [x] Fix $browser_language fallback
- [x] Fix #629 
- [x] Fix Warning in user Form (red instead yellow)
- [x] nginx detection by @phl0 
- [x] get rid of flag cdn

image

PR #632, opened 2024-07-25
Fix PHP variable name for  upload_max_filesize

phl0:
Fixes https://github.com/wavelog/wavelog/issues/631.


phl0:
Will be corrected with https://github.com/wavelog/wavelog/pull/603.

PR #636, opened 2024-07-26
[Logbookadvanced] Added LoTW sent/received to batch edit

AndreasK79:
image
image


phl0:
Batch edit basically works. But I guess we need to deal with display of states like R or Q. Those are still shown in red (using other colors in paper QSL for example).

phl0 (changes requested):
Added some code to also color LoTW R states :) See https://github.com/AndreasK79/wavelog/pull/32.

phl0 (approved):
Good now :-)

PR #644, opened 2024-07-27
set rst also on fav recall

### LINKED ISSUE #642: Update RST(S) RST(R) when using favorites
When setting band/mode in Live-QSO via favorites, the RST(S) and RST(R) dont't get updated. If CW was set before and switched to SSV, the default in the fields stays at 599 599. When switching to SSB via dropdown, it turns to 59 59. Would be nice if it also gets update when using favorites

HB9HIL:
fix #642


phl0 (approved):
Tested + works. Nice and helpful improvement.

PR #657, opened 2024-07-29
Added a page for QSL statistics, just numbers

AndreasK79:
Analytics -> QSL Statistics

Raw numbers for each band/sat and mode.


phl0 (approved):
works fine now.

PR #652, opened 2024-07-29
added pre-precheck for mbstring

HB9HIL:
Probably no further description needed

telegram-cloud-photo-size-4-5852941982723326596-y

PR #660, opened 2024-07-29
Add curl to installer required modules

phl0:
PR https://github.com/wavelog/wavelog/pull/652 contained only half of the truth as curl is also required for the nginx checker to work. Without this the same bug as for mbstring occurs for users using nginx as webserver.

So made the function a little more versatile by adding a list for pre-installer required modules so that it is extendable should there be need in the future. Currently it has mbstring and curl.


HB9HIL (approved):
Good addition. I wasn't aware of that

PR #663, opened 2024-07-30
Changed to use contest name in Cabrillo export, if available

AndreasK79:
Use contest name if available (contest id if not) when doing a Cabrillo export:
image


HB9HIL (approved):
works as expected. consistency rocks :)

PR #667, opened 2024-07-30
VUCC grids are not plotted correctly on the dashboard map

phl0:
VUCC grids are not plotted correctly on the dashboard map at least.

Screenshot from 2024-07-30 22-39-20

The reason is that the array holding to grids for a specific QSO is not reset but added to for multiple QSOs. Thus all further multi grid QSOs after the first one get an unplausible midpoint. See screenshot.


AndreasK79 (approved):

AndreasK79:
Fixed confirmed working. Nice observation. I don't have any border/cross qsos in my log, so difficult to notice myself!

PR #670, opened 2024-07-31
Apply user defined units to distance plots on map

phl0:
Distance plot on map was broken in https://github.com/wavelog/wavelog/pull/605 as user defined units where added. Results in distance being shown an NaN:

Screenshot from 2024-07-31 08-53-14


AndreasK79 (approved):

AndreasK79:
@phl0 thanks for fixing my mistake from earlier :)

PR #671, opened 2024-07-31
Fix "Check missing DXCC"-Updater

int2001:
Navigate to Admin -> Update Conutry-Files.

Old behaviour:
"Check QSOs missing DXCC Data" and "ReCheck all QSOs in Logbook" throws an error if there's a QSO without a DXCC

New behaviour:
"Check QSOs missing DXCC Data" and "ReCheck all QSOs in Logbook" 
updates the DXCC according to the DXCC derivated from Call and Date


HB9HIL (approved):
works :) catches the error

PR #673, opened 2024-07-31
Change exportall ADIF view to be switchable between internal render and download file

DB4SCW:
In light of pull request #664 and @HB9HIL's comments, I thought I could attempt the cleanup of the ADIF creation views to really complete the task.

I implemented the switch so that being a downloadable file is the default behaviour of the `exportall.php` ADIF creation view when no other indication is given. I chose that path to reduce impact overall.

New API endpoint now uses `exportall.php` and the old `exportapi.php` file is gone.

All functions providing a downloadable file should work as expected, as well as the new api endpoint from #664.

Tests can be done by calling the relevant functions in GUI or by using the new API ADIF pull endpoint:
`curl -d '{"key":"MYKEY","station_id":1, "fetchfromid":1}' http://log.address.here/index.php/api/get_contacts_adif`


HB9HIL (approved):
nice additon :)

int2001 (approved):
Tested all (new) dependencies: Looks good. Works.

Thank u!

PR #675, opened 2024-08-01
Update azimuth on updating distances

phl0:
This PR contains:

- [x] Update azimuth info on mass-updating distance data
- [ ] Store azimuth on QSO logging (only displayed in current dev state)
- [ ] Update / show azimuth on changing locator in log QSO pane


phl0:
Thinking a little more about this issue I came to the conclusion that this does not make sense. According to ADIF standard the field ANT_AZ (where I thought about storing the value) explicitely refers to antenna position. The case is with SAT QSOs which are probably the reason for the existence of this field (an ANT_EL). Example:
QSO with VO1/M0XUU. He was located in GN37PL. Calclulating the bearing from JO31 gives 288°. But this QSO was done via QO-100. And for that bird my antenna has to point 156.8° azimuth. So the antenna position is totally different from bearing to DX from my station.

PR #681, opened 2024-08-02
Migration Lockfile

HB9HIL:
The idea with the lockfile for migrations was good but it's currently not failsafe enough. There are several cases where the lockfile does not get deleted. 

With this PR there are two changes:

1. The lockfile lives not longer in `application/cache` but now in the webservers `/tmp` folder.
2. The lockfile has a limited lifetime of 300 seconds by default in config (fallback is 480 seconds)

If the lifetime exceeds the lockfile get's deleted and the migration can run again.


int2001:
Mainlogic works so far. But the previous shown warning is gone. Intended?

int2001 (approved):
tested / works

PR #688, opened 2024-08-03
Set 599 also for CW-* modes

HB9HIL:
Yaesu uses CW-U and CW-L. Both do not match with the ADIF modes, therefore we map them to `CW`


phl0 (approved):
Code works. Would remove the logging though.

PR #700, opened 2024-08-05
Default Radio

### LINKED ISSUE #418: Add the ability to set a default radio used for live QSO

HB9HIL:
Ability to set a default radio from the hardware interfaces list. 

image
image

Behaviour:
- if no default radio is set:
   Switch to QSO logging / Contesting -> no radio selected by default
   Log a QSO -> last used radio get's written to session data
   Go to another page and come back to QSO logging / Contesting
   Last used radio is already selected
   Radio is lost on logout / re-login

- if a default radio is set:
   Switch to QSO Logging / Contesting -> default radio is already set
   log a QSO -> last used radio get's written to session data
   Go to another page and come back to QSO logging / Contesting
   Last used radio is already selected

   log a QSO with another Radio which was already selected (default radio) -> last used radio get's written to session data
   Go to another page and come back to QSO logging / Contesting
   Last used radio is already selected (not default)

In addition to the new default radio feature, it shows which radio was last updated with CAT data.

`fixes #418`


application/models/Cat.php
```diff
@@ -109,6 +109,11 @@ function radio_status($id) {
 			return $this->db->query($sql);
 		}
 
+		function last_updated() {
+			$sql = 'SELECT * FROM cat WHERE user_id =' . $this->session->userdata('user_id') . ' ORDER BY timestamp DESC LIMIT 1';
```
int2001, line 113: Can u pse use binding here?

int2001 (approved):
tested this. works.
should make handling for CAT much easier

PR #703, opened 2024-08-06
Fixing XSS Vulnerability

HB9HIL:
There was no XSS clean on the QSO manual_mode GET parameter.

Special thanks to Mark Laurence Lat for reporting this.


int2001 (approved):
works.
and only "1" or "2" is possbile. so double-checked (xss_clean + enum)

PR #705, opened 2024-08-07
Fix calculation of distance if wked from same-grid or other errors

int2001:
Unter special circumstances the PHP-function returns „NAN“, which causes trouble at some places.

This one sets the distance to `0` instead of `NAN`

Tnx to @dg9vh for finding this one


phl0 (approved):
Easy!

PR #713, opened 2024-08-07
Fixes a bug where $band wasn't set

int2001:
Reproduceable by adding a new user and navigating (with that new user) directly to gridmap.

Reason was:
`$band` was iterated at the top of the view. So if the user had a QSO on `SAT` the Var `$band` stayed at SAT (even outside the loop).
This caused a correct behaviour (but was wrong implemented).

if there's a user without ANY QSO, `$band` is empty (because the loop found nothing to loop). This caused a PHP-Warning.

interesting bug ;)


HB9HIL (approved):
good to find leftovers :)

PR #716, opened 2024-08-08
Sat stuff

AndreasK79:
The start of a SAT map and a SAT planner. This is still very much in beta.
To get TLE's, you need to run wavelogurl/index.php/update/update_tle

The menu items are marked as beta, and will only show if dev mode is active.
image


int2001 (approved):
Screened (in a quick view) through the files.
Known as this is beta, i found no offense sec-flaws.

PR #720, opened 2024-08-08
[SFLE] RST was saved swapped around 

HB9HIL:
RST sent was saved as received
RST rcvd was saved as sent

This fixes this behaviour. RST is now saved correctly


DC1TC:
s/not/now?

Change looks reasonable to me.

int2001 (commented):
works. good find

PR #724, opened 2024-08-10
Sat stuff

HB9HIL:
Some small improvements of @AndreasK79 Code

- added untranslated strings to gettext workflow
- adjusted frontend and reduced some code
- use the user's set datetime format in the resultstable
- commented out the "start date" select as it's not in use

I also tried to use the timezones we already have in database. Issue here is that we do not support Summer/Winter Time in Wavelog....


AndreasK79 (approved):

AndreasK79:
@HB9HIL great. Thanks for them improvements!

PR #730, opened 2024-08-11
fix osm tile copyright attribution

violarulan:
In some places where we call Leaflet, the code still uses the hard-coded attribution: `Map data © <a href="https://openstreetmap.org">OpenStreetMap</a> contributors'`. 

This PR makes it use the value which is defined in database (row `map_tile_server_copyright` in `options` table).

Code description: 

JS variable `option_map_tile_server_copyright` is defined in `views/interface_assets/footer.php`. 
https://github.com/wavelog/wavelog/blob/646ccf2783042ba6390cea1ecc5ab4d3996b7b3c/application/views/interface_assets/footer.php#L6

All JS files which are changed in this PR can access it safely.

### DIFF
--- a/application/views/interface_assets/footer.php
+++ b/application/views/interface_assets/footer.php
@@ -737,7 +737,7 @@ function showActivatorsMap(call, count, grids) {
     var maidenhead = new L.maidenheadactivators(grid_four).addTo(map);
 
     var osmUrl='<?php echo $this->optionslib->get_option('option_map_tile_server');?>';
-    var osmAttrib='Map data © <a href="https://openstreetmap.org">OpenStreetMap</a> contributors';
+    var osmAttrib='<?php echo $this->optionslib->get_option('option_map_tile_server_copyright');?>';
     var osm = new L.TileLayer(osmUrl, {minZoom: 1, maxZoom: 12, attribution: osmAttrib});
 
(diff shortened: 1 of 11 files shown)


application/views/interface_assets/footer.php
```diff
@@ -737,7 +737,7 @@ function showActivatorsMap(call, count, grids) {
     var maidenhead = new L.maidenheadactivators(grid_four).addTo(map);
 
     var osmUrl='<?php echo $this->optionslib->get_option('option_map_tile_server');?>';
-    var osmAttrib='Map data © <a href="https://openstreetmap.org">OpenStreetMap</a> contributors';
+    var osmAttrib='<?php echo $this->optionslib->get_option('option_map_tile_server_copyright');?>';
```
AndreasK79, line 740: @violarulan I guess this could be changed to the global variables as, instead of fetching from options, since we already set it on top.

AndreasK79:
@violarulan nice find. Seems we have been lazy, and forgotten to change it. Thanks for fixing!

application/views/interface_assets/footer.php
violarulan: @AndreasK79 Thanks for the notice o7. I've created a new commit.

PR #737, opened 2024-08-12
Bindings

int2001:
Added Bindings to "Accumulated Stats"


HB9HIL (approved):
tested various Accumulated Stats with different Modes and Bands
everything works fine. Results are also the same as before these important fixes.

Not tested with SAT QSO's (because I don't have SAT QSOs haha)

AndreasK79 (approved):

PR #738, opened 2024-08-12
Preventing SQL-Injection 

int2001:


HB9HIL (approved):
did not test each function but code looks familiar and absolutely reasonable. Good work

HB9HIL (approved):

PR #740, opened 2024-08-13
Fix map attribution

violarulan:
Another part of #730.

This PR includes two commits:

1. In #730, a change was introduced to use the `option_map_tile_server_copyright` variable. But in some page, i.e. `/visitor/<callsign>`, the `/views/interface_assets/footer` is not included. So that `leafembed.js` cannot access that variable. This commit is a fix.
2. Make logbook_advanced map to use the attribution info. 

I'm still in the test to see if other page is affected with the change. So far there's no other impact. If I notice more I'll submit another PR.


AndreasK79:
@violarulan thanks again. I fixed this in the exportmap feature too.

AndreasK79 (approved):

PR #742, opened 2024-08-13
Cookies Fix

HB9HIL:
If a user set a cookie_prefix in the config.php the behaviour of the language and keep_login cookies was broken. 

This is a fix for this issue.


int2001 (approved):
works. rare-config with cookie-prefix...

PR #743, opened 2024-08-13
This fixes activated gridsquares where only a border or corner was worked

### LINKED ISSUE #726: "Gridsquare Activators" search issue with CALL/VO1 - multiple grid locators in "gridsquare"
When I looked in "Gridsquare Activators" I could not found those entries from a rover from CALL/VO1.

I tracked down the issue to all those QSO were made from a gridline. Both grids logged into "Gridsquare". With most of wavelog now supporting it, it causes a "ignore" in the search results.

- PHP:              8.1.29
- MySQL/MariaDB:  10.5.25-MariaDB
- Wavelog:          1,8.2 (was before also 1.7.2 )

2024-08-10 17_21_35-Window
2024-08-10 17_21_17-Advanced logbook - Wavelog – Mozilla Firefox

AndreasK79:
Fixed #726.

The problem is if you work a station only in grid border or corner, it will not get listed.

I am not 100% sure that this will work with older MySQL versions.

### DIFF
--- a/application/controllers/Activators.php
+++ b/application/controllers/Activators.php
@@ -30,7 +30,7 @@ class Activators extends CI_Controller
         if ($this->input->post('mincount') != NULL) {   // mincount is not set when page first loads.
             $mincount = $this->input->post('mincount');
         } else {
-            $mincount = 2; 
+            $mincount = 2;
         }
 
         if ($this->input->post('leogeo') != NULL) {   // orbit is not set when page first loads.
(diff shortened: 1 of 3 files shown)


HB9HIL (approved):
tested working, @phl0 should have a bigger database for testing this aswell 😄

int2001 (approved):
works as expected. good find

phl0 (approved):
With SQL group by fix now running fine in my test env.

PR #745, opened 2024-08-13
Fix visitor satellite page grids not showing when logged out

violarulan:
### Issue: 

If the user is logged out, the grids not showing in visitor satellite page (`/visitor/satellites/<slug>`). It shows nothing on the map. While being logged in, it shows again.

### Root cause: 

```
get_band_confirmed() 
get_band() 
get_band_worked_vucc_squares() 
get_band_confirmed_vucc_squares()
```

Each of these 4 functions takes 10 arguments, while we only give them 9 in `/application/controllers/Visitor.php`. 

The last two arguments are `.., $propagation, $logbooks_locations_array = NULL`. 

If we give it 9 arguments and we are logged in, the code will re-fetch the associated logbook id from session. 

And it seems that whatever propogation mode string we pass, it just works without warnings.

### DIFF
--- a/application/controllers/Visitor.php
+++ b/application/controllers/Visitor.php
@@ -224,7 +224,7 @@ class Visitor extends CI_Controller {
 
 
 		// Get Confirmed LoTW & Paper Squares (non VUCC)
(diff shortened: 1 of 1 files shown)


AndreasK79 (approved):

AndreasK79:
Nice find. Tested working!

PR #749, opened 2024-08-14
Added binding to queries

AndreasK79:
This is for the timeline_model.


int2001 (approved):
Reviewed Code under the aspect of security. Looks very good

PR #750, opened 2024-08-14
notes xss cleaning

HB9HIL:
some xss adjustments in the notes section

### DIFF
--- a/application/models/Note.php
+++ b/application/models/Note.php
@@ -6,8 +6,7 @@ class Note extends CI_Model {
         if ($api_key == null) {
 			$user_id = $this->session->userdata('user_id');
 		} else {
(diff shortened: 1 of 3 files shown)


application/models/Note.php
```diff
@@ -31,23 +30,37 @@ function add() {
 
 	function edit() {
 		$data = array(
-			'cat' => xss_clean($this->input->post('category')),
-			'title' => xss_clean($this->input->post('title')),
-			'note' => xss_clean($this->input->post('content'))
+			'cat' => $this->input->post('category', TRUE),
```
int2001, line 33: why? it's the same - just a different notation ;)

HB9HIL: consistency

PR #752, opened 2024-08-14
Prevent SQL Injections in Helvetia Model

HB9HIL:


int2001 (approved):
works. one more safe.

PR #755, opened 2024-08-15
Support both operators and special callsigns on same instance

derFogel:
This patch adds a special auth_level for special callsings and only shows the operator popup if the user has the special_callsign auth_level.

### DIFF
--- a/application/config/config.sample.php
+++ b/application/config/config.sample.php
@@ -73,6 +73,9 @@ $config['use_auth'] = true;
 $config['auth_table'] = 'users';
 $config['auth_mode'] = '3';
 
+# enable this if you want a special / club callsign
+# $config['auth_level'][2] = 'Special Callsign';
+
 $config['auth_level'][3] = 'Operator';
 $config['auth_level'][99] = 'Administrator';
 
@@ -631,7 +634,7 @@ $config['disable_manual_qrz'] = false;
 | Disables QSL-Image-Feature
 |--------------------------------------------------------------------------
 |
-| This disabled the whole QSL image feature if you don't need it and want to hide it. 
+| This disabled the whole QSL image feature if you don't need it and want to hide it.
 | Set to true will hide all QSL image related stuff in Wavelog
 |
 */
@@ -643,7 +646,7 @@ $config['disable_qsl'] = false;
 | Disables OQRS-Feature
 |--------------------------------------------------------------------------
 |
-| This disabled the whole OQRS feature if you don't need it and want to hide it. 
+| This disabled the whole OQRS feature if you don't need it and want to hide it.
 | Set to true will hide all OQRS related stuff in Wavelog
 |
 */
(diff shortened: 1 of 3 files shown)


application/config/config.sample.php
```diff
@@ -74,6 +74,7 @@
 $config['auth_mode'] = '3';
 
 $config['auth_level'][3] = 'Operator';
+$config['auth_level'][5] = 'Special Callsign';
```
HB9HIL, line 77: Please add this line in `install/config/config.php` aswell

HB9HIL (changes requested):
I like the idea to make it more flexible. However I see two issues:

1. If we add a new Auth-Level in the sample config and install config. The new authentication level is see by ALL new instances. Even if they probably never use it. This can maybe result in wrong expectations?
Solution could be to do not add this at all to the configs and just adjust the wiki or make the wiki and config more clear with comments. Maybe @int2001 has an opinion about that. IMHO the `install/config/config.php` and `config.sample.php` should be identical in every case. 

2. A user with auth level "Special Callsign" has theoretically and practical a lower permission level then a regular "Operator" user. Therefore it would make more sense to give this user a permission level of `2` instead of `5`. The whole authentication checks in Wavelog are already checking for level `2` which wouldn't bring a big change (I think we have only three places in the API Controller which is checking for `3` but this can be changed). 


I agree with the fact, that this allows to have normal operators and special callsign on the same instance is a cool thing. But I would be careful how we deploy that. Because adding this to the config.sample.php is imho a big change which shows ALL new installs a third auth level in the user menu. Probably better solution would be to add just a tutorial / "How to activate" in the Wiki and do not add this in the config at all. **(Or add it but comment it out!)**

@int2001 @AndreasK79 @phl0 
I think we should discuss this in team how to handle that.

HB9HIL:
@derFogel Please use your `dev` branch next time. Not your `master` branch
image

application/config/config.sample.php
derFogel: `5` was more or less random. But will change that to `2`.
I like the idea with the comment, need to figure out how to edit the wiki once this is merged / released.

HB9HIL: The Wiki is open to edit for everyone

https://github.com/wavelog/wavelog/wiki/Recommended-Setup-for-Special-Callsigns-and-Clubs

derFogel: Ah that doesn't work. That would disable api access which would suck when you use wsjt-x.

HB9HIL: @derFogel I don't see an issue to change API access to `authorize(2)` but the others have to agree with that

derFogel: Actually the api (I tested with the radio api) works when I change it to 2

derFogel: With a permission level of `2` you cannot log anything, so this will be a bigger change.

int2001: my 2ct on that:
i like the idea to configure it as well.
but imho an option (via the user_options_model) may fit better here.

reasons:
- The grants (in a classical sense) are still the same for those stations.
- it's "only" the popup which is shown to those special-logins. logic is completly the same like a normal user has (as a "normal user" you can set the OP-Call at the first QSO - see station_tab. OP is persitant for the session
- with the user_option we don't touch the config.php - it is minimal-invasive compared to a new Userlevel

HB9HIL:
Hello @derFogel 

I think we won't merge this one. We have other plans to address your request. I'm sorry. Please be patient. This will be addresses in the next bigger release (maybe 1.9 but I can't say for sure).

PR #756, opened 2024-08-15
Catches a non-numeric CAT-QRG

int2001:
Sometimes a CAT-Tool tries to publish strange QRGs like 'Error' or sth. like that.
This one catches those QRGs before a SQL-Error can appear


HB9HIL (approved):
Prevents 'shit in'. I like

PR #758, opened 2024-08-15
[GETTEXT] HTML encoding

HB9HIL:
We decided to completely encode HTML in gettext strings. This is more secure and we can get prevent issues with special chars. 

This is replacing PR #751


int2001 (approved):
good thing!
read the code and cannot find any flaws.

Be careful with the weblate here in future ;)

PR #760, opened 2024-08-16
Update fontawesome to v5.15.4

HB9HIL:
Updated our Fontawesome Library from Version 5.6.3 to Version 5.15.4

The existing icons stay the same but it adds new icons which will be used in the future.

> [!TIP]
> Search for icons here: https://fontawesome.com/v5/search?o=r&m=free


int2001 (approved):
tested various pages where flags and so on are embedded. works

PR #761, opened 2024-08-16
Dashboard warnings

HB9HIL:
image

Our authorizations "one-liners" currently redirect to the dashboard after setting a flashmessage. This flashmessage can't be shown because it's not implemented. 

- This PR fixes that. 
- It changed the "notice" (bs class 'info') to a "danger" alert
- Made the string translatable (find & replace)

PR #762, opened 2024-08-16
Impersonate

HB9HIL:
Allows administrators to log in as a non-admin user. This helps with debugging and can be used for special callsigns or club stations. 

Depends directly on #760 and indirectly on #761


int2001:
Great improvement. I tested it: Works.

i'd like to see a bit more security, since this is a really hot feature:
- Put a time-component into the hash. so if it is lost, the risk will be minimized
- Configurable via config.php (`$enable_impersonate` - default false?)


Additional:
- Perhaps adding the Source-User and OP-Call to the Hash, so one is able to derivate that for further use (Setting OP later, logging purpose, ...). Would fit a generic approach.

int2001 (approved):
great improvement!
works and - looks safe.

PR #763, opened 2024-08-16
Fix band unit

AndreasK79:
This PR fixed a few things regarding the qrg units.

1. Save didn't work correctly.
2. Units were fetched on EVERY page load, but not needed.
3. Had to fix footer.php with an isset on dxcc_list. I had an error while trying to figure this stuff out.

And thanks to @int2001 for some input which lead to the final solve :)


int2001 (approved):
works. good find

HB9HIL (approved):
Sorry, I could have published this earlier. 😄

Issue #766, opened 2024-08-17
Distance is not stored correctly


phl0:
The Logbook controller does not store the correct distance for a station in case a profile with a grid other than the one in config is used. The JS function for bearing display works correctly and bases its calculations on the grid of the active station profile. This needs to be used for the distance calculation for the ADIF field as well. Result of the jsonlookupcall function:

![Screenshot 2024-08-17 164835](https://github.com/user-attachments/assets/5331e189-7a4b-4813-bcf9-98807fa1e164)

The distance function within the Logbook controller is missing the active station profile and therefore always using the one from config. Results before and after the patch:

![Screenshot 2024-08-17 170504](https://github.com/user-attachments/assets/c7783bc0-d3ab-4ac3-b75d-7ecfcf02ab90)

phl0:
Fix proposed in https://github.com/wavelog/wavelog/pull/767.

phl0:
Fixed.

PR #767, opened 2024-08-17
Use grid of active station location for distance calculations

phl0:
This fixes https://github.com/wavelog/wavelog/issues/766

### DIFF
--- a/application/controllers/Logbook.php
+++ b/application/controllers/Logbook.php
@@ -139,7 +139,7 @@ class Logbook extends CI_Controller {
 
 		$return['callsign_name'] 		= $this->nval($callbook['name'] ?? '', $this->logbook_model->call_name($callsign));
 		$return['callsign_qra'] 		= $this->nval($callbook['gridsquare'] ?? '',  $this->logbook_model->call_qra($callsign));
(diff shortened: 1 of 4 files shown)


AndreasK79 (approved):

phl0:
Added function to update distance information for all QSOs. Could you please re-check @AnreasK79?

phl0:
Added function to update distance information for all QSOs. Could you please re-check @AndreasK79?

phl0:
I think we have an issue here with all those functions in country files section as they update *all* QSOs instead of only the ones for the current user. Needs investigation and will be handled separately.

phl0:
> Needs investigation and will be handled separately.

Handled in https://github.com/wavelog/wavelog/issues/769.

int2001:
Tested and works:
Idea: Can we add a spinner / lock for the button, since this is a long-running operation and the admin can't see if there's something happening in the BG?

phl0:
I will refactor the htmx stuff separately. Merging the current state.

phl0:
Spinner stuff to be handled in https://github.com/wavelog/wavelog/issues/771.

PR #778, opened 2024-08-19
[LoTW] Fix flash messages and catch the case if a user tries to upload a 'key-only' file

HB9HIL:
Fixed the flash messages in LoTW index and

if a user tries to upload a 'key-only' p12 file, Wavelog crashes with a Error 500. This catches this case and redirects to the lotw page with a warning (see second commit)


phl0 (changes requested):
Fix works. But we should strip the path from filename imho. Also updated wording a little in https://github.com/HB9HIL/wavelog/pull/110.

phl0 (approved):
Good to merge then.

PR #783, opened 2024-08-21
Stop empty or zero freq from breaking LoTW confirmations

phl0:
FREQ and FREQ_RX should only be included if not empty *and* not 0. Otherwise just omit these fields.


phl0:
In addition a load of useless " ?><?php" was removed ...

HB9HIL (approved):
didn't test, but code looks good :)

int2001 (approved):
Tested in prod. Works

PR #784, opened 2024-08-21
[QRG] Improve Frequency Display

### LINKED ISSUE #165: Improve Frequency input

HB9HIL:
This PR improves the display format of frequencies in Wavelog. 

Based in the used QRG unit in the `Bands` view the frequencies will be shown in the defined unit. 
The function `qrg_conversion()` is built as generic as possible. There are various options to return different strings. If needed we could add json return aswell but maybe to a later point. 

Frequency Input in QSO, Contest and SimpleFLE logging will be added to a later point. 

`fixes #165 `


int2001 (approved):
looks good. tested it as well - everything works as intended.

lukezaa:
Could you improve it even more to be more like in QRZ which is de facto standard.. 

Frequency then user input then hard coded unit. 

I often make mistakes when I input 28.470 not 28.470M.. 

image

HB9HIL:
@lukezaa this is the plan

Issue #780, opened 2024-08-20
CronManager doesn't execute some jobs


winnieXY:
I've used the latest docker image according to [the wiki](https://github.com/wavelog/wavelog/wiki/Installation-via-Docker).

The jobs do run manually, e.g. i can upload/download to lotw, eqsl, qrz, clublog, etc. using the menu (or in case of clublog using a url). I do not find errors within the log file which could indicate a issue. Executing the command in /etc/cron/wavelog manually also doesn't indicate a error. 

```
root@0dfe98d0fcc2:/var/www/html# curl --silent http://localhost/index.php/cron/run
CRON: clublog_download -> is due: false -> Next Run: 2024-08-21 00:07:00
CRON: clublog_upload -> is due: false -> Next Run: 2024-08-20 12:03:00
CRON: eqsl_sync -> is due: false -> Next Run: 2024-08-20 12:09:00
CRON: hrdlog_upload is disabled. skipped..
CRON: lotw_lotw_upload -> is due: false -> Next Run: 2024-08-20 08:00:00
CRON: qrz_download -> is due: false -> Next Run: 2024-08-20 12:18:00
CRON: qrz_upload -> is due: false -> Next Run: 2024-08-20 12:06:00
CRON: update_dxcc -> is due: false -> Next Run: 2024-09-01 00:20:00
CRON: update_lotw_users -> is due: false -> Next Run: 2024-08-26 01:10:00
CRON: update_update_clublog_scp -> is due: false -> Next Run: 2024-08-25 00:00:00
CRON: update_update_dok -> is due: false -> Next Run: 2024-09-01 00:00:00
CRON: update_update_pota -> is due: false -> Next Run: 2024-09-01 00:15:00
CRON: update_update_sota -> is due: false -> Next Run: 2024-09-01 00:05:00
CRON: update_update_wwff -> is due: false -> Next Run: 2024-09-01 00:10:00
```

**Version Information**
- OS (Host):             Debian 12
- PHP:              not relevant due to docker
- MySQL/MariaDB:   mariadb:11.3 image
- Wavelog:     ghcr.io/wavelog/wavelog:latest (Version 1.8.2)     

**Steps to Reproduce**
1. Fill in all credentials for qrz, eqsl, lotw, ...
2. Enable all cronjobs
3. Wait 2 days
4. See attached screenshot.

![image](https://github.com/user-attachments/assets/211a2469-7017-42b9-97ab-1556287db019)

[...]

PR #785, opened 2024-08-21
[CRON] Allow insecure self-signed certificates with config switch

HB9HIL:
Some users want to use the cronmanager while `$config['base_url]` contains a https url with a self-signed certificate. This does not work and results in failing jobs.

In this case you now can set `$config['cron_allow_insecure'] = true; ` in the `config.php` to allow the curl option `--insecure` (or `-k`)

In addition to that I did some small UI adjustments

### DIFF
--- a/application/config/config.sample.php
+++ b/application/config/config.sample.php
@@ -683,3 +683,16 @@ $config['sc_hide_usermenu'] = true;
 */
 
 $config['disable_impersonate'] = false;   
+
+
+/*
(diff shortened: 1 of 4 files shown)


HB9HIL:
Could fix #780 but not sure if this is the issue here.

int2001 (approved):
Works.
Maybe add a hint to the wiki, that - in special cases - the main-curl from the cron should also be called with `-k`

PR #795, opened 2024-08-23
Create a helper to read in a optional local_url to bypass the need to…

winnieXY:
Hey,

I think i solved the issue with the cronmanager (at least for me). 

This patch will create a config option "local_url" which can be used to provide an alternative url to access the cronmanager  (e.g. "http://localhost/".

If this option is not set the base_url will be returned instead.

### DIFF
--- a/application/controllers/Cron.php
+++ b/application/controllers/Cron.php
@@ -89,7 +89,9 @@ class cron extends CI_Controller {
 						echo "CRON: " . $cron->id . " -> RUNNING...\n";
 
 						$url = local_url() . $cron->function;
-						echo "CRON: " . $cron->id . " -> URL: " . $url . "\n";
+                        if (ENVIRONMENT == "development") {
+						    echo "CRON: " . $cron->id . " -> URL: " . $url . "\n";
+                        }
 
 						$ch = curl_init();
 						curl_setopt($ch, CURLOPT_URL, $url);
--- a/application/controllers/Cron.php
+++ b/application/controllers/Cron.php
@@ -88,7 +88,7 @@ class cron extends CI_Controller {
 						echo "CRON: " . $cron->id . " -> is due: " . $isdue_result . "\n";
 						echo "CRON: " . $cron->id . " -> RUNNING...\n";
 
-						$url = base_url() . $cron->function;
+						$url = local_url() . $cron->function;
 
 						$ch = curl_init();
 						curl_setopt($ch, CURLOPT_URL, $url);
--- a/application/controllers/Cron.php
+++ b/application/controllers/Cron.php
@@ -89,6 +89,7 @@ class cron extends CI_Controller {
 						echo "CRON: " . $cron->id . " -> RUNNING...\n";
 
 						$url = local_url() . $cron->function;
+						echo "CRON: " . $cron->id . " -> URL: " . $url . "\n";
 
 						$ch = curl_init();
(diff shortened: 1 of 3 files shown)


winnieXY:
Okay, tested it on my local instance. Seems to work for me. Please also test and kindly review :-)

application/controllers/Cron.php
```diff
@@ -88,7 +88,8 @@ public function run() {
 						echo "CRON: " . $cron->id . " -> is due: " . $isdue_result . "\n";
 						echo "CRON: " . $cron->id . " -> RUNNING...\n";
 
-						$url = base_url() . $cron->function;
+						$url = local_url() . $cron->function;
+						echo "CRON: " . $cron->id . " -> URL: " . $url . "\n";
```
HB9HIL, line 92: Since this is not necessary for the normal user I would depend this on the developement mode


```suggestion
                        if (ENVIRONMENT == "development") {
						    echo "CRON: " . $cron->id . " -> URL: " . $url . "\n";
                        }
```

system/core/Config.php
```diff
@@ -322,6 +322,36 @@ public function base_url($uri = '', $protocol = NULL)
...
+	 *
+	 * @uses	CI_Config::_uri_string()
+	 *
+	 * @param	string|string[]	$uri	URI string or an array of segments
+	 * @param	string	$protocol
+	 * @return	string
+	 */
+	public function local_url($uri = '', $protocol = NULL)
+	{
+		$local_url = $this->slash_item('local_url');
```
HB9HIL, line 338: This does not work if `$config['local_url']` is not defined. I would solve this like that


```suggestion
		$local_url = $this->slash_item('local_url') ?? $this->slash_item('base_url');
```

system/helpers/url_helper.php
```diff
@@ -88,6 +88,28 @@ function base_url($uri = '', $protocol = NULL)
...
+	 * If local_url is not defined return base_url instead.
+	 * 
+	 * @param	string	$uri
+	 * @param	string	$protocol
+	 * @return	string
+	 */
+
+	 function local_url($uri = '', $protocol = NULL)
+	 {
+		return get_instance()->config->local_url($uri, $protocol) ?? get_instance()->config->base_url($uri, $protocol);
```
HB9HIL, line 109: Since you should add ` ?? $this->slash_item('base_url');` in the `core/Config` you should remove this here


```suggestion
		return get_instance()->config->local_url($uri, $protocol);
```

HB9HIL (changes requested):
Needs to be bullet proof for users, which did not add `$config['local_url']`. See the comments

system/helpers/url_helper.php
winnieXY: Done

system/core/Config.php
winnieXY: Done.

HB9HIL (approved):
Seems good to me. Tested with the cronmanager behind a proxy.

Grace Period of max. 24 hours before merge so @int2001, @phl0 or @AndreasK79 have a chance to see this change aswell

HB9HIL:
@winnieXY 

Please modify the Wiki article here: https://github.com/wavelog/wavelog/wiki/Recommended-Cron-Jobs-and-Cronmanager
Add the `local_url` config here so the users know about. :)

PR #796, opened 2024-08-23
Just set the status of the cronjob if the job was run or was disabled.

### LINKED ISSUE #794: CronManager overwrites failed status after one Minute again with "healthy"
If a internal cronjob fails, the cronjob will state correctly "failed". However after one minute (after the cron/run is called once again) the failed status is changed back to healthy - thus a user will always see healthy states (except one minute after the failed job) although the job is failing all the time.

**Version Information**
- OS:               Debian 12
- PHP:              Docker container
- Wavelog:          dev

**Steps to Reproduce**
Create a cronjob which fails and check the status after each cron/run run. 

**Expected Behavior**
The failed status is preserved as long as it was not successful.

winnieXY:
This small patch will just set the "healthy" flag if the cron was executed or was disabled/enabled, so that a failed flag persists.

This a attempt to close #794

### DIFF
--- a/application/controllers/Cron.php
+++ b/application/controllers/Cron.php
@@ -75,7 +75,7 @@ class cron extends CI_Controller {
 					$cronjob = $this->cronexpression;
 					$dt = new DateTime();
 					$isdue = $cronjob->isMatching($dt);
-					$set_status = false;
+					$set_status = true;
 
 					$next_run = $cronjob->getNext();
 					$next_run_date = date('Y-m-d H:i:s', $next_run);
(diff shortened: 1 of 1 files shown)


winnieXY:
Tested it and verified that with #795 applied (and wrong local_url) the failed state will be persistent, so the state gives indeed a good indicator if the cronjob is healthy or not.

Caveat - if you enable a cronjob it will stick to "pending" till the first run. Currently it will state "healthy" as soon as a time for the next execution is calculated.  However I think that this is acceptable, as the cronjob is indeed pending and we do not know (yet) if it is healthy or not.

HB9HIL (approved):
nice fix :) thx!

PR #797, opened 2024-08-23
Add EDI export for contests in REG1TEST format

DB4SCW:
This PR adds the ability to export a contest as EDI format. 

Since EDI is only a colloquial term and technically references the REG1TEST format, this terminology is used in Code and (to some extend) in the detail view.

Known "inconsistencies" with the format and its implementation in this PR:

- format can only contain 7 bit ASCII characters. This is not enforced and left to the user.
- format has several arguments such as "Bonus per each new Exchange" in the header or "QSO-Points" in the QSO details that wavelog knows nothing about and is contest specific. These fields are intentionally left empty on the assumption that all programs receiving REG1TEST files will overwrite those values anyway. The format documentation states that fields non-critical to scoring may be left empty, which all of these are in my opinion. Those fields are documented in the code comments.
- since wavelog does no scoring, all scoring result fields are left empty as they are in wavelogs CBR export and permitted by format description
- REG1TEST may only be used on bands higher than 50MHz. You can create such a file using this export for bands lower than that, but will receive the value of "invalid" in the band argument. Should be a non-issue since no contest using REG1TEST should be on bands below 50 MHz.
- REG1TEST supports bands higher than what wavelog supports. For those pretty rare QSOs, since input is not possible, so is exporting as EDI.
- "Crossmode-QSOs", which are possible and explicitly specified in the format (e.g. RX in CW, TX in SSB) are not possible since there is no possibilty to log those in wavelog (to my knowledge).

Format description used to implement this:
[Format description](https://www.ok2kkw.com/ediformat.htm)

Tests can be done by exporting a contest on 50 MHz and above using the provided UI.


HB9HIL:
@DB4SCW Please see https://github.com/DB4SCW/wavelog/pull/1

DB4SCW:
added @HB9HIL s bindings and fixed redundant variable `$year` from various functions regarding band loading

phl0 (changes requested):
Code works. Maybe add a hint to the band selection? Selecting a band lower than 6m results in invalid. This should be displayed to the user before executing the export I guess:

image

phl0 (approved):
The warning stuff about band looks good to me now.

PR #806, opened 2024-08-25
Allow AMSAT status page upload for QSOs logged via API

phl0:
Currently only manually logged QSOs via GUI are sent to AMSAT status page. This patch allows for API logged QSOs to be reported to the AMSAT SAT status page.


int2001 (changes requested):
since import_bulk is used for ADIF-Import as well, this would slow down the import of ADIF-Files significantly.

Suggestion:
- Push (relevant!) Amsat-Notifications - usually last 3-5 days - to an extra-array
- Process Bulk-Upload as before
- Inform Amsat with the previous built array

Benefits:
- ADIF Import happens as well, if amsat is laggy/not reachable.
- No flooding of amsat with QSOs which are ancient.

phl0:
Refactored now and based on timestamps.

PR #809, opened 2024-08-25
Update buttons locked and with spinner class

HB9HIL:
Added button locks and spinner class for the update view.


HB9HIL:
Will continue on this later. Needs some UI adjustments

phl0 (approved):
Good hint ... forgotten. Works properly now.

PR #811, opened 2024-08-26
Consistent propmode selection for accumulated timeline

int2001:
New Selections for Accumulated Timeline (like Timeline-Analytics):

- All is now All
- SAT sets prop_mode implicit to SAT
- New prop_mode "All but SAT"


phl0 (approved):
Worx ok here :)

PR #812, opened 2024-08-26
[Contesting] Added Exchangetype "Serial + Grid + Exchange"

HB9HIL:
Added Exchangetype "Serial + Grid + Exchange" to support for example the Moon Contest (which also requires #797 )

In addition I added some security in the contesting model (bindings and xss_clean)

### DIFF
--- a/application/models/Contesting_model.php
+++ b/application/models/Contesting_model.php
@@ -8,7 +8,7 @@ class Contesting_model extends CI_Model {
 		$this->load->model('Stations');
 		$station_id = $this->Stations->find_active();
 
-		$qsoarray = explode(',', $qso);
+		$qsoarray = explode(',', $this->security->xss_clean($qso));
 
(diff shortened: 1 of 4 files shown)


phl0 (approved):

HB9HIL:
Added an additional button which contains the CopyToField feature and an option to choose the sequence of the exchanges. This is handy especially for different types of contest. 

For example: 

Moon Contest need `Serial` - `Grid` - `Exchange`
The "Frankencontest", a german DOK contest needs: `Serial` - `Exchange (with copy to dok)` - `Grid`
Thx @DB4SCW for this hint

It would be absolutely awful to have the wrong sequence during the contest.

Last TODO is to fix the "Press Space to jump to the next input" feature. 

image

phl0 (approved):
(Still) working as designed. :)

PR #816, opened 2024-08-26
Intercept eqsl image errors (which are not rate limits)

phl0:
Sometimes eqsl fails to provide images. Looks like this:

Screenshot from 2024-08-26 17-26-49

We try to catch this error and proceed with the rest of the queue. After patch:

Screenshot from 2024-08-26 17-25-16


HB9HIL (approved):
Nice improvement :)

PR #823, opened 2024-08-27
Allow for ITU zone to be fetched from qrz/hamqth lookup

phl0:
Now as we have input fields for ITU zone we should get it from callbook queries like.
Implemented:
- qrz.com
- hamqth.com


HB9HIL (approved):
seems to work. tested qrz and hamqth.

PR #824, opened 2024-08-27
ITU Single Lookups 

HB9HIL:
Added aswell Single Callbook Update (in qso edit)
and in LBA Update from Callbook


phl0 (changes requested):
I think we should priorize the callbook data and use DXCC zone info as backup like https://github.com/wavelog/wavelog/pull/823/files#diff-15d104bd141cae4f18c96e6791c4f262d0e05450b17d6b5d239102914cc4986aR900?

phl0 (approved):
Works :)

PR #830, opened 2024-08-27
Impersonate maintenance

HB9HIL:
When maintenance mode is active it should be possible for an admin to still impersonate an user, while the user is not able to login from outside a session.


int2001 (approved):
Looks ok to me

PR #832, opened 2024-08-28
Prevent HTTP error 500 on import with incorrect ADIF values

phl0:
Imports with incorrect ADIF data causes error 500. Example:

```
<QSO_DATE_OFF:9>-00011130
```

First measure: catch exception and continue with rest of code.


phl0:
We are awaiting the malicious ADIF file which caused the HTTP 500 initially to verify this patch.

phl0:
Patch working verified with the original file.

int2001 (approved):
worx. one more fence

PR #836, opened 2024-08-28
Fix Issues where Dates/Times of QSOs are not well filled

int2001:


phl0 (approved):
Works on my mysql db system :)

phl0 (approved):

PR #839, opened 2024-08-29
[Logbookadvanced] Added button to find "invalid" QSOs

AndreasK79:
New button in LBA to search for "invalid" QSOs.

Checks if call/mode/band is empty.
Checks if date is empty, or the date is 01/01/1970.


int2001 (approved):
works. good to have a check here

PR #850, opened 2024-08-31
Fix user login / edit set cookies

toseppo:
**Fix user login / edit set cookies**

Function [set_cookie](https://github.com/wavelog/wavelog/blob/master/system/core/Input.php#L348) already has a time() added to the `expire` array element, hence no need to add `time()` twice.

In the current code(where `expire` equals to the `time()+1000`) system call `setcookie` silently fails probably due to the `expire` being too far in the future (twice the time() + extra). And this leads to the Internal Server Error.

There are other places in the code that set `expire` as `time() +- extra` == need to be fixed too.


HB9HIL:
Hello @toseppo 

Please do not create Pull Requests against `master`. Always against `dev`.

[https://github.com/wavelog/wavelog/wiki/Development-Notes](https://github.com/wavelog/wavelog/wiki/Development-Notes)

HB9HIL:
Hello @toseppo 

You are right by the fact that using `time()` here is indeed not correct. However I can't see any relation to your `mod_rewrite` issue. I tested your change and I pushed the other expire values aswell to this PR. 

But to be honest: I don't see any relation to your created Issue. You may want to explain what you figured out?

See the Issue for more details about my thoughts to this.

toseppo:
mod_rewrite is a post effect only probably, but the real problem was in setting cookie too far in the future  for me.

So the behaviour I faced with the code `expire: time() + 1000` in https://github.com/wavelog/wavelog/blob/master/application/controllers/User.php#L897 is that PHP application hangs silently (no errors, nothing in the logs and no any code(I had logging in place there when debugging) is being executed after set_cookie one line 901 with wrong time).

toseppo:
I believe this https://github.com/wavelog/wavelog/pull/850/commits/d92258e9aebbc9725dd3944f250881b188b77aad fix may need some more work as [set_cookie](https://github.com/wavelog/wavelog/blob/master/system/core/Input.php#L342) on negative `expire` will set it to 1 despite the negative quantity. Or is that desired?

HB9HIL:
Which browser are you using? 

On all tested browsers (Chrome, Firefox, Edge) the Cookie gets deleted when setting expire time to `-3600` (expired one hour ago)

toseppo:
Tested with Chrome and Safari

toseppo:
I've caught the actual error.

Code to reproduce:
```
setcookie('language', 'en', time() + time(), '/', '', false, false);
```

Error:
```
Fatal error: Uncaught TypeError: setcookie(): Argument #3 ($expires_or_options) must be of type array|int, float given in /var/www/html/index.php:3 Stack trace: #0 /var/www/html/index.php(3): setcookie('language', 'en', 3450575422.0, '/', '', false, false) #1 {main} thrown in /var/www/html/index.php on line 3
```

toseppo:
Similar error in the past: https://github.com/magicbug/Cloudlog/discussions/2565

HB9HIL:
Okay @toseppo 

I was able to reproduce at least the cookie stuff.

```php
class Test extends CI_Controller {

    public function cookietest() {
        error_reporting(-1);
		ini_set('display_errors', 1);

        // Test with time() + 3600
        setcookie('test_a', 'test', time() + 3600); //success but does not include codeigniter's cookie prefix

        // Test with 3600
        setcookie('test_b', 'test', 3600); //failed silently

        // Test with time() + 3600
        $this->input->set_cookie('test_c', 'test', time() + 3600); // success but too far in the future (sets to 1 month instead to 1 hour)

        // Test with 3600
        $this->input->set_cookie('test_d', 'test', 3600); // success
    }
}
```

But I was not able to reproduce your login issue. Did you test this branch and does it solve it your login issue?

HB9HIL:
> I believe this [d92258e](https://github.com/wavelog/wavelog/commit/d92258e9aebbc9725dd3944f250881b188b77aad) fix may need some more work as [set_cookie](https://github.com/wavelog/wavelog/blob/master/system/core/Input.php#L342) on negative `expire` will set it to 1 despite the negative quantity. Or is that desired?

The reason why it's set to a negative value (`-1` should also work) is to delete the cookie immediately. This should also work in your browsers.

toseppo:
Thanks!

HB9HIL:
@toseppo Whats your callsign?

PR #857, opened 2024-09-02
Changed QSL-Label to "to radio"

int2001:


phl0 (approved):
Look ok to me and works.

PR #858, opened 2024-09-02
[WIP]: One sat per distance record

phl0:
Tnx @AndreasK79 for the "fast" (i.e. MySQL 8+) query. Will need to find a solution for MySQL < 7.


AndreasK79 (approved):

AndreasK79:
Tested working, both queries. Good to go as far as I can tell.

PR #859, opened 2024-09-02
Add input/edit boxes for antenna azimuth / elevation

phl0:
As the fields for az and el are handled via API already it would be fine to also expose them to the user and give the chance to log / edit them.


int2001 (approved):
Works for me.

was a bit concerned about the collection (`input->post`) of the values. But: They're parsed as double at inserting. so no real problem.

PR #862, opened 2024-09-03
Put az/el boxes to the sat pane and handle empty / 0 values

phl0:


int2001 (approved):
Tested with MariaDB:
- Appearance (only at SAT / check)
- saving NULL - check
- saving values - check

PR #866, opened 2024-09-03
fix REG1TEST format and expand it to include relevant parts of scoring

DB4SCW:
After consultation with one of the organizers of a German VHF contest, I fixed and extended the REG1TEST format:

- fixed format of time in qso details
- completely overhauled the way the qso detail data is calculated to allow for scoring
- added band multiplicator as an input field in GUI with a default value of 1. According to my contact, other values are possible, but highly unlikely.
- added calculation of claimed points in qso detail
- added calculation of the flags concerning "new DXCC", "new Exchange" and "new Locator"
- added claimed score to header

Testing can be done by exporting a contest above 50Mhz from the GUI and changing the band multiplicator field.
Also for verification of the format, change the dates of your QSOs to February 23rd, 2025 and [upload the log here](https://ham-awards.de/award/AC_upload.php). All the test logs will be removed before the next contest starts. Thanks to DL4EAX for providing the dummyupload!


HB9HIL (approved):
works :) did not test every detail but can't spot any issues here

phl0 (approved):
good to merge now.

PR #867, opened 2024-09-03
Just give empty result on frequency being 0

phl0:
It seems that some ADIF export allow FREQ (and FREQ_RX) being 0 which does not make much sense to display:

image

So in case of any of the values being zero we should return `Null` so it looks like this:

Screenshot from 2024-09-03 14-43-24


HB9HIL (approved):
works :) thx

PR #870, opened 2024-09-03
show the "To Radio" string in the upper right corner

HB9HIL:
In addition to #857 let's show the "To Radio" in the upper right corner. 

Needs to be testet with various common formats but seems to work. I needed to introduce the `$additionalText` because the title is usually shown directly above the table which results in "Confirming QSOs" to be above "To Radio" which is not the idea. This way it works. "Confirming QSOs" is now an additional text under the title but above the QSO table. 

@int2001 please test this one

image

image


phl0 (approved):
Nice tweak. Works here.

int2001 (approved):
Easy hack with great positive impact.
tnx for investing the effort on that, @HB9HIL 

works

PR #875, opened 2024-09-04
Make the Header Logo link on OQRS page go back to public slug page

phl0:
Initially set up to make the user go back to public sluc page from OQRS page.
Ended up cleaning visitor's header and footer from CI calls and assigned the variables in corresponding controllers.

### DIFF
--- a/application/controllers/Oqrs.php
+++ b/application/controllers/Oqrs.php
@@ -59,7 +59,7 @@ class Oqrs extends CI_Controller {
 		$this->load->model('oqrs_model');
 		$data['result'] = $this->oqrs_model->getQueryDataGrouped($this->input->post('callsign'));
 		$data['callsign'] = $this->security->xss_clean($this->input->post('callsign'));
-		
+
 		if($this->input->post('widget') != 'true') {
 			$this->load->view('oqrs/request_grouped', $data);
 		} else {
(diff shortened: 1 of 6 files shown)


application/controllers/Oqrs.php
```diff
@@ -15,17 +15,23 @@ function __construct() {
...
+		$data['disable_oqrs'] = $this->config->item('disable_oqrs');
 		$data['stations'] = $this->oqrs_model->get_oqrs_stations();
 		$data['page_title'] = __("Log Search & OQRS");
 		$data['global_oqrs_text'] = $this->optionslib->get_option('global_oqrs_text');
 		$data['groupedSearch'] = $this->optionslib->get_option('groupedSearch');
 
 		$this->load->view('visitor/layout/header', $data);
 		$this->load->view('oqrs/index');
-		$this->load->view('interface_assets/footer');
+		$this->load->view('interface_assets/footer', $data);
```
HB9HIL, line 34: we should not pass `$data` twice to a page


```suggestion
		$this->load->view('interface_assets/footer');
```

HB9HIL (changes requested):
can't spot any other issues in this PR :) good job!
after removing `$data` mentioned in the comment this one is good for me.

HB9HIL (approved):
looks good to me :) let's wait for @AndreasK79 approval aswell

AndreasK79 (approved):

AndreasK79:
Works while testing. Also removed the index in the url made it nicer. And code is even simpler now with the dynamic mapping function.

PR #876, opened 2024-09-04
Show open and total number of OQRS requests

phl0:
What do you think about this @AndreasK79?


AndreasK79 (approved):

AndreasK79:
@phl0 you know I like all those small tweaks here and there :)

PR #882, opened 2024-09-05
Fixes isset-issue (fields are always set, when out of DB)

int2001:
Tested with several constellations. File was processed by LoTW

@AndreasK79 / @HB9HIL / @phl0 pse check carefully


phl0 (approved):
I can confirm that this (besides patch by @violarulan) fixes the issue of errorneous LoTW uploads.

PR #883, opened 2024-09-05
[Advanced Logbook] Added continent to columns

AndreasK79:
Added column in table.
Added filter.
Added check to invalid check.
Added column to user options.
Added continent to batch edit.


phl0 (approved):
Code workds. Nice feature.

int2001 (approved):
tested and read code: works.
Nice one

PR #888, opened 2024-09-06
Added dialog to inform when no location is selected

AndreasK79:
Fence in place:
image


phl0 (approved):
Works fine.

phl0:
Maybe we wanna handle the empty location also in PHP?
Just my suggestion: https://github.com/AndreasK79/wavelog/pull/33

PR #890, opened 2024-09-06
[Awards] Added Worked All Continents

AndreasK79:
image


phl0 (approved):
Works :) Thank you for some more open columns and rows to fill

phl0 (approved):
Fixed a syntax error. Still working good now :)

PR #895, opened 2024-09-06
unset freq_in_mhz(_rx) before each QSO

int2001:


phl0 (approved):
Tested and confirmed fix with FREQ and FREQ_RX separately.

PR #898, opened 2024-09-06
fix public map

HB9HIL:


int2001 (approved):
Tested / Dots are back on map now

PR #903, opened 2024-09-07
perf(ux): use sans-serif as font

imlonghao:
The default font `serif` in Chinese look nightmare to me, especially on Windows.

So I proposed to change the fallback font to `sans-serif`, it is not the best but way better than `serif` I think.

image

image

image

image


HB9HIL (approved):
can't spot any difference in latin letters. If this improves readability in chinese I'm fine with it :)

PR #905, opened 2024-09-07
[CONTESTING] Fix some issues in contesting

HB9HIL:
Got some issue reports from a user and try to fix some stuff

### DIFF
--- a/application/controllers/Contesting.php
+++ b/application/controllers/Contesting.php
@@ -38,6 +38,11 @@ class Contesting extends CI_Controller {
 		$data['contestnames'] = $this->contesting_model->getActivecontests();
 		$data['bands'] = $this->bands->get_user_bands_for_qso_entry();
 
+		$footerData = [];
+		$footerData['scripts'] = [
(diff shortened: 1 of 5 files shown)


HB9HIL:
The following commit handles a more tricky case. 

if a user starts a new session with contest "A" 
logs a QSO with contest "A" and switches then the contestname to contest "B"

the whole session will be broken. Reason for this is, that the contest_id in the contest_session get's updated, but not the qso data which is necessary to get the correct QSO list. As we don't want to edit already logged QSO's (the first logged QSO was logged with contest "A") a change of the contest during logging should be handled like a new session. Now it would be really bad to directly and with no further information just reset the session. 

The more safe way is to just disable the contestname select during after logging the first QSO. If you need to change the contest you have to start a new session. The wrongly logged QSO in the beginning is in the database but not part of the session anymore.

HB9HIL:
I think its ready for a first test.

int2001 (approved):
- scrolled through code. looks good as seen via "Eye-Integral".
- tested a bit (live contest / restart / post contest / etc.): feels a bit more resilient.

PR #907, opened 2024-09-08
Qrz state modified

HB9HIL:
After editing a QSO the new QRZ.com state is defined as "modified" in database "M"

we could show this also in the Log Tables to make clear that this QSO was modified and will be uploaded again.

image


int2001 (approved):
More transparency for the users.

phl0:
Should we use another color here? IIRC yellow was used for another state already.

int2001:
> Should we use another color here? IIRC yellow was used for another state already.

as far as i remember yellow is "re-upload". at classic-QSL it is "requested"

phl0 (approved):
Let's leave it like this then. Too many different colors also not the best idea :)

PR #914, opened 2024-09-09
Disable map click if modal is shown

phl0:
Disabling cklick input upon showing the modal works. But re-enabling it is still an issue.

### DIFF
--- a/assets/js/leaflet/geocoding.js
+++ b/assets/js/leaflet/geocoding.js
@@ -95,6 +95,9 @@ function onMapMove(event) {
 };
 
 function onMapClick(event) {
+	if ($('.modal-dialog')[0]) {
+		return;
+	}
 	var LatLng = event.latlng;
 	var lat = LatLng.lat;
 	var lng = LatLng.lng;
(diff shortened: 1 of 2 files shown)


assets/js/leaflet/geocoding.js
```diff
@@ -95,6 +95,9 @@ function onMapMove(event) {
 };
 
 function onMapClick(event) {
+	if ($('.modal-dialog')[0]) {
+		return;
+	}
```
HB9HIL, line 100: this is not necessary
```suggestion
	
```

assets/js/sections/gridmap.js
```diff
@@ -200,8 +200,8 @@ function spawnGridsquareModal(loc_4char) {
 					}]
 				});
 			    dialog.realize();
-		    		$("#gridsquare_map").append(dialog.getModal());
-		    		dialog.open();
+					$("#gridsquare_map").append(dialog.getModal());
```
HB9HIL, line 203: change the id instead will fix the issue
```suggestion
					$("body").append(dialog.getModal());
```

HB9HIL (changes requested):

phl0:
Closing this in favor of https://github.com/wavelog/wavelog/pull/943 which is the better solution.

PR #918, opened 2024-09-10
Use only certs that belong to the station_id/user

int2001:


phl0 (approved):
Works ok. Added info/error message for the user.

PR #920, opened 2024-09-10
[ITU Zones] Added a checker

AndreasK79:
image
The biggest problem is that our data source is lacking.

ITU Zone for Alaska is not set.
ITU Zone 4 is not set.

Others might also not be set, I haven't checked the table, it was just those I noticed while trying it out myself.


HB9HIL (approved):
works :)

PR #921, opened 2024-09-10
[Advanced Logbook] Added CQ and ITU zone to the bottom

AndreasK79:
image


HB9HIL (approved):
works great! nice addition

Issue #712, opened 2024-08-07
Feature Request: Add Version check to Cronjob


Loki1737:
Hi, I would beliefe it would be beneficial, if we can add the version check to the cronjob. And if a new version is available, create a notification (or even auto-update)

PR #929, opened 2024-09-12
Add a version check

phl0:
This adds functionality to check if there is a newer release of Wavelog on github and displays a banner to admin users if that's the case. The update check should be triggered by cron so it is non-blocking and interval can be user-configured.
This addresses https://github.com/wavelog/wavelog/issues/712.

### DIFF
--- a/application/config/config.sample.php
+++ b/application/config/config.sample.php
@@ -695,4 +695,16 @@ $config['disable_impersonate'] = false;
 | 
 */
 
-$config['cron_allow_insecure'] = false;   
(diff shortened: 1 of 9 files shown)


int2001 (approved):
like it.
additional Wiki-Entry for config-option and cron?!

phl0:
> additional Wiki-Entry for config-option and cron?!

Will add (latest on release)

HB9HIL (approved):

int2001 (approved):
tested it. works

PR #930, opened 2024-09-12
Update only QSOs of station_ids of User

int2001:
Update only QSOs of station_ids which belong to the User when DLing LoTW QSLs


phl0 (approved):
Fixed a small code error. Now works as expected.

PR #934, opened 2024-09-12
Notes are getting exported

HB9HIL:
As `NOTES` is an official ADIF field we exported this field ever since. Therefore we remove the misleading info that these field are for internal use only


int2001 (approved):
Works. makes it much more clearer for the user.

phl0 (approved):

PR #935, opened 2024-09-12
Lookup ITU zone from DXCC if callbook results are empty

phl0:


AndreasK79 (approved):

AndreasK79:
Thanks for improving my thought :)

PR #939, opened 2024-09-13
Mark QRZ-QSOs as Invalid when specific Error at qrz while uploading

int2001:
When trying yo upload a QSO to QRZ.com which happened out of valid Dateranges (setted at QRZ), the QSO is tried to upload over and over.

This fix marks the QSL-State (QRZ) as invalid, if that happens.


HB9HIL (approved):

phl0 (changes requested):
Needs adjustment because match does not exactly cover qrz.com's response and thus not working.

phl0 (approved):
The patch now works for me. Added two more commits:
- Show I state in log overview in grey
- Also catch "wrong station callsign" case

PR #943, opened 2024-09-13
fixes buggy modal dialog

HB9HIL:
replacement for #914


phl0 (changes requested):
Fixes the errror in browser view but breaks the click action in fullscreen view ... :(

HB9HIL:
> Fixes the errror in browser view but breaks the click action in fullscreen view ... :(

All issues solved I guess. @phl0 and @AndreasK79 please check. 

I created two new functions in common.js which can be used again if this issue appears somewhere else aswell. 

I also added a dark background additionally to make the modal in the map more visible in fullscreen mode. This adds two backgrounds in the normal view which makes it likely darked. But imho something we can live with.

HB9HIL:
@phl0 

PS: Your code 

```js
        if ($('.modal-dialog')[0]) {
		return;
	}
```

was necessary aswell as the onMapClick() stuff differs a bit from the standard leaflet behaviour

phl0 (approved):
This works nicely now. Closing https://github.com/wavelog/wavelog/pull/914 in favor of this. Tnx fer digging in.

AndreasK79 (approved):

PR #946, opened 2024-09-14
Fix some minor eQSL-Issues where values aren't set

int2001:
shoud fix those ones:

```
ERROR - 2024-09-14 09:15:22 --> Severity: Warning --> Undefined array key "qsl_sent" /var/www/html/application/libraries/EqslImporter.php 159
ERROR - 2024-09-14 09:15:59 --> Severity: Warning --> Attempt to read property "nodeValue" on null /var/www/html/application/controllers/Eqsl.php 323
```


HB9HIL (approved):
looks good by eye integral. did not test with eqsl

PR #948, opened 2024-09-14
[ITU Zone] Determine ITU zone from grid.

AndreasK79:
In QSO Entry, the entered grid is used to determine ITU zone.
In the case a border or cross grid is entered, the zone might in some cases be wrong if you happen to be on a the border for zones.


HB9HIL (approved):
works :) tested and read through the code

AndreasK79:
Seems these kind of queries might not work as universal as we thought.

Exception: Longitude -200.000000 is out of range in function st_geomfromgeojson. It must be within (-180.000000, 180.000000].

Trying to figure out if a solution exists.

AndreasK79:
The compatibility between MariaDB and MySQL is not there.
MariaDB works. MySQL on the other hand, is a problem.
Some info here: https://bugs.mysql.com/bug.php?id=101359

I am closing this.

PR #954, opened 2024-09-15
[Advanced Logbook] set qrz to modified when update from callbook or batch edit

HB9HIL:
When using the "Update from Callbook" Button in LBA and there is an API key for the given station location we should update the Upload Status to QRZ.com to "Modified".
The same behaviour for batch editing


AndreasK79 (approved):

HB9HIL:
found a bug. qrx

HB9HIL:
@AndreasK79 Please check again. Found a bug where al QSO were marked as modified even they weren't ever uploaded. 
I also changed the query to run once with a IN clause instead for each QSO. 

Checking for a qrz key is not necessary as if it was marked manually as uploaded it should also be marked as modified. If the Upload status is 'N' it wont be marked as modified.

AndreasK79 (approved):

PR #955, opened 2024-09-15
show nicer numberformat for total qsos and show also number of users

HB9HIL:
just some minor stuff because I like it that way

- changes the number format of "total qsos" to add thousand seperators `'`
- added number of users on this instance


int2001 (approved):
nice addition

PR #960, opened 2024-09-17
Order LoTW certs by callsign

phl0:
Ordering by callsign seems more intuitive to me :)


HB9HIL (approved):
shouldn't harm anything

PR #961, opened 2024-09-17
[ADIF] Catch some more ADIF errors

HB9HIL:
catched some more ADIF issues. 

had a case where 

`<MODE:19>DIGITAL VOICE (DMR) ` caused some trouble and some `$band` were undefined. 

(Correct would be: `<MODE:12>DIGITALVOICE <SUBMODE:3>DMR`)


phl0 (changes requested):
Reading the code looks ok

phl0 (approved):
Reading the code looks ok

PR #963, opened 2024-09-18
fix an issue with station ids

HB9HIL:


int2001 (approved):
tested this. fixes the issue.

PR #965, opened 2024-09-18
Adds an optional "limit" to the "get_contacts_adif"-API

int2001:
Adds an optional "limit" to the "get_contacts_adif"-API.

Given the optional parameter limits the export to the Number which is posted within that field.
if filed is not given or `-1` it returns everything


DB4SCW:
Works with Hamawardz!

PR #966, opened 2024-09-18
Mark Pirates as invalid

HB9HIL:
QRZ.com does not accept D1 or D0 callsigns as these pirate stations do not have a valid DXCC. Therefore we can safely mark them as invalid for QRZ.com uploads

### DIFF
--- a/application/controllers/Qrz.php
+++ b/application/controllers/Qrz.php
@@ -119,6 +119,16 @@ class Qrz extends CI_Controller {
 
 		if ($data['qsos']) {
 			foreach ($data['qsos']->result() as $qso) {
+
+				// if the callsign starts with D1 or D0 we skip the upload as QRZ.com do not accept these callsigns
+				if (substr($qso->COL_CALL, 0, 2) == 'D1' || substr($qso->COL_CALL, 0, 2) == 'D0') {
+					$this->markqso($qso->COL_PRIMARY_KEY,'I');
+					$errormessages[] = 'Callsign starts with D1 or D0: ' . $qso->COL_CALL . ' Band: ' . $qso->COL_BAND . ' Mode: ' . $qso->COL_MODE . ' Time: ' . $qso->COL_TIME_ON;
+					$errormessages[] = 'QSO not uploaded to QRZ.com as QRZ.com do not accept callsigns starting with D1 or D0 (due the lack of DXCC information).';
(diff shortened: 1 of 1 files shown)


int2001 (commented):
Should we try to create a more generic approach?
Something like: "QRZ says no - so mark as I".

i am not really happy with the edge-case to check for those D0/D1-Stations

AndreasK79:
Perhaps we should consider adding something more general like a flag for invalid? I did look into the log table, and it seems we do not have this for the time being.

HB9HIL:
@int2001 @AndreasK79 

I'm open to suggestions.

This is the response of QRZ.com for the Callsign D1CW

```
QRZ upload failed with the following message: STATUS=FAIL&RESULT=FAIL&REASON=Replace error on record: DXCC could not be determined for D1CW&EXTENDED=
```

AndreasK79:
> Perhaps we should consider adding something more general like a flag for invalid? I did look into the log table, and it seems we do not have this for the time being.

My suggestion was a more general thing to mark QSOs as invalid, which would mean more changes needed in the code.
In the end, the upload code would only consider those not having the invalid flag.

HB9HIL:
>  In the end, the upload code would only consider those not having the invalid flag.

so you mean for all third-party services?

AndreasK79:
> > In the end, the upload code would only consider those not having the invalid flag.
> 
> so you mean for all third-party services?

Yeah, that could be a way, if we need such a flag that is.

Perhaps the easiest would be to catch the error from QRZ and mark those QSOs during upload as invalid.

int2001:
> ```
> QRZ upload failed with the following message: STATUS=FAIL&RESULT=FAIL&REASON=Replace error on record: DXCC could not be determined for D1CW&EXTENDED=
> ```
> 

yep. try to catch that at https://github.com/wavelog/wavelog/blob/71f919c2845d78c25ba41f7aae6a1426bd96572c/application/controllers/Qrz.php#L144

and mark it as "I" at the QRZ_QSL-Field

phl0:
> yep. try to catch that at (around that that block - there are already other catches)
> and mark it as "I" at the QRZ_QSL-Field

This is the solution I'd vote vor. My guess is that we need to catch different errors for different 3rd party services anyway.

HB9HIL:
#968  is the better solution for this issue

PR #968, opened 2024-09-18
Ignore invalid DXCC for qrz.com uploads

phl0:
Alternative for https://github.com/wavelog/wavelog/pull/966.


phl0:
Patch working here. Also added corresponding color and tooltip to LBA.

HB9HIL (approved):

PR #969, opened 2024-09-18
Basics for exporting station locations to a JSON file

phl0:


int2001:
i think it's still in progress. but since a review was requested:

* Export creates the json, but it's shown in the browser. A simple header at the view may change that it is presented as download.
* Perhaps a warning like "File contains sensitive information"
* Import is missing (perhaps this is part of an update?)
    * When it comes to import: It's not safe, that the same station_id would be inserted (station_id may exist, and so on)

Issue #977, opened 2024-09-19
Add Support for Prefilling Callsign via URL Parameters in Logging Window


besynnerlig:
**What Feature is Missing?**

I would like to request the addition of a feature in Wavelog that allows the logging window to accept query parameters in the URL to prefill form fields, specifically the callsign. This means that when accessing a URL like:

```
https://wavelog.example.com/index.php/qso?callsign=SE6S
```

Wavelog would open the logging window with the callsign field already populated with "SE6S". This functionality would enable external applications or scripts to initiate the logging process with prefilled data, enhancing interoperability.

**Why is This Feature Important?**

I use an SDR radio (SmartSDR connected to a FlexRadio) that displays spots on its panadapter. When I click on one of these spots, my setup can be configured to automatically open a URL in my browser. If Wavelog could accept the callsign as a query parameter, it would streamline the process of start logging a QSO initiated from my radio. This integration would make starting a QSO by clicking a spot in the panadapter quick and seamless, significantly improving my operational efficiency.

Moreover, this feature would reduce manual data entry errors and save time.

**Is This Feature Personal or Beneficial to Others as Well?**

While this feature addresses my specific use case, I believe it would be beneficial to many other Wavelog users. Operators who use SDR radios or other equipment that can trigger browser actions would find this integration valuable. It would enable them to:

- **Enhance Efficiency**: Quickly start logging QSOs with prefilled information, reducing the steps required.
- **Improve Accuracy**: Minimize manual data entry, reducing the likelihood of errors.
- **Increase Flexibility**: Allow for better integration with various hardware and software setups, catering to diverse operating styles.

Implementing this feature would make Wavelog more versatile and appealing to a broader user base, encouraging adoption and satisfaction among amateur radio operators who seek seamless integration between their equipment and logging software.

![image](https://github.com/user-attachments/assets/d22628d2-f886-4928-a387-3eb75d70f59b)
_Spots appearing on the SmartSDR Panadapter_

phl0:
Is there documentation about this feature? For example parameter names etc.?

besynnerlig:
Thanks @HB9HIL 
Sounds great and I will test it tomorrow and let you know the results!

HB9HIL:
Implemented

besynnerlig:
Dear Wavelog Development Team,

[...]

PR #978, opened 2024-09-19
Open QSO logging with call

HB9HIL:
This is addressing #977  which is kinda tricky to solve. The assumed solution with a simple GET url will cause some issues in UX and behaviour of the QSO logging page in case of reloads (it would fill the call over and over again). Also it would open a new tab of the QSO windows everytime the URL is called which is not really clever. 

Therefore I want to try another approach by calling an "transfer-URL" which handles the backend stuff and calls the QSO view with already existing infrastructure. 

To use this you simply call 

```
https://<URL to Wavelog>/index.php/qso/log_qso?callsig=CALLSIGN
```

This will open a page which do some Javascript magic and pipes the callsign to a already existing QSO windows (or opens a new one and pipes then). It uses a similar code like the bandlist stuff of @int2001 but slightly different and imho a bit more efficient as I did not use fix timeouts, but directly process the stuff when a message arrives. The transfer page gets closed when the data was piped to the QSO view. 

This is a proof of concept. Test and try to break it. 

> [!IMPORTANT]
> Clear the cache before testing!


int2001 (changes requested):
Nice approach. But it fails with default Chrome-Settings because of a missing pop-up-permission. Don't know the reason (Code looks like the same, but at DXCluster this doesn't happen).

Furthermore the tab with the callsign-param stays open.

possible solutions:
- Either adding a hint: "Pse enable pop-ups" or fixing it
- closing of the tab which was called / or redirect with a post from there.

PS: Delay was there for a reason. Implemented it, because sometimes the new-tab taked a bit longer to render - depends on the PC.

besynnerlig:
I have now tested this implementation on my local Wavelog installation using the development branch, with the pull request successfully merged. I tested it using Mozilla Firefox, which prompted me to allow popups for Wavelog, and after enabling them, the solution worked flawlessly.

I am genuinely impressed by how smoothly this solution functions. The handling of the callsign transfer and the efficient way the existing QSO window is updated (or a new one is opened) without delay is a very elegant approach. It feels robust and well-thought-out, especially considering the complexity of dealing with reloading and preventing multiple tabs from opening.

The only minor enhancement I would suggest is that the "Time on" field doesn't seem to refresh automatically unless I close the tab and let a new one open the next time. Aside from that, everything works perfectly, and I can see this being incredibly useful for my workflow with SmartSDR.

Overall, **this is a fantastic solution**, and I appreciate the time and effort put into developing it. **Thank you!**

HB9HIL:
Just discovered an generic issue with modern browsers. Currently idk how to fix this:

image

HB9HIL:
The issue with the "Scripts may close only blablabla" is that I called the URL by typing in the link into the browsers address field. In this case the browser tab was not opened by code, but by the user (mine) input. In this case the browser can't close the tab with the command `window.close()` for security reasons. 

But if the URL is called by an external link. Let's say you create another script which opens this URL or you click on this link in another software or page, the tab was opened by software and can close itself by `window.close()`

To catch the case of a non closing tab in UX manner I changed the redirect message after the data was redirected. So the user knows that he can close the tab.

int2001:
if there's really no chance to "autoclose" we may leave it as it is.
Code looks okay, functionality is given.

this is an edge-case. i can give a "soft-approval" with this message.

besynnerlig:
Hi there,

I have done more testing. I really appreciate the improvements being made. I did notice a small issue that I wanted to bring up.

As I had previously mentioned, the "Time On" field doesn't automatically refresh to the current time when a new logging window is opened. While this can be resolved manually by clicking the "reset start time" button, I noticed something a bit more concerning: the same issue also occurs with the QSO date field.

When starting a new logging session, both the date and time fields are carried over from the previous QSO. This means that if no QSO has been logged on the current day, the date field is set to the previous QSO's date without any way to reset it, as there's no "reset" button for the date field. As a result, this could cause incorrect logging entries, especially if a user doesn't realize the date is out of sync.

Ideally, both the QSO start date and time should reset to the current date and time when a new logging window is opened—similar to what happens when clicking "Live QSO" from the menu. This would ensure that log entries are always accurate without requiring manual intervention.

Thanks again for your efforts, and I hope this feedback is helpful!

besynnerlig:
I’ve noticed a small issue that occurs when I start logging a QSO (which I often do just to gather information about a callsign without actually completing the log) and leave the browser tab open. When I later begin logging a QSO for a different callsign and hover the mouse pointer over the LoTW image, I sometimes see two black popups, or occasionally just one, showing LoTW data from the previous callsign I had started logging but didn’t finish. It seems that the wrong LoTW image is also displayed in these cases.

Thank you for everything you've done so far! This is just a small observation that I hope can help improve the experience even further.

image

Browser used Mozilla Firefox 130.0.1 (64-bit) on Windoze

HB9HIL:
> As I had previously mentioned, the "Time On" field doesn't automatically refresh to the current time when a new logging window is opened. While this can be resolved manually by clicking the "reset start time" button, I noticed something a bit more concerning: the same issue also occurs with the QSO date field.
> 
> When starting a new logging session, both the date and time fields are carried over from the previous QSO. This means that if no QSO has been logged on the current day, the date field is set to the previous QSO's date without any way to reset it, as there's no "reset" button for the date field. As a result, this could cause incorrect logging entries, especially if a user doesn't realize the date is out of sync.

Hey @besynnerlig 

unfortunately I was not able to reproduce this issue. Did you clear your browser cache? (Windows `Shift + F5` or Mac `Cmd + Shift + R`)

I assume that you may already have a POST QSO Windows open which gets filled with the callsign Information. This should not happen as clicking on a callsign in your application assumes that you want to log a qso NOW and not in the past. Therefore it should open a LIVE QSO window and not use the POST QSO window. Am I right that you experience this issue with a POST QSO window open? (same "bug" would be in the bandlist dxcluster)

besynnerlig:
Thanks @HB9HIL 
I cleared the cache.

I always have a Wavelog tab open  https://wavelog.MYSERVER/index.php/logbook It's a "pinned tab" in Firefox.  (No other wavelog URL:s opened at all)

My app opens a new logging tab at https://wavelog.MYSERVER/index.php/qso?manual=0

image

"Time off" field shows the current time. The "Time on" and the "Date" is fetched from my previous QSO

My previous QSO:
image

So, no  POST QSO window open ...

HB9HIL:
> I always have a Wavelog tab open https://wavelog.MYSERVER/index.php/logbook It's a "pinned tab" in Firefox. (No other wavelog URL:s opened at all)

Okay. I was able to reproduce it. The keyword was "Firefox"

This does not happen in Chrome. I will try to figure out what's the issue here.

HB9HIL:
@besynnerlig 

I found the issue. It was a racecondition, which caused several micro bugs with the timers. I restructured `qso.js` and moved all functions out of the `$(document).ready(function () {}`

The second one which played with that aswell is that we currently fill in the last QSO data in the date/time fields. I changed that so this only happens in POST QSO. 

so the issue should be solved now. Please test again. 


@int2001, @phl0 and @AndreasK79 

This was absolutely necessary to avoid future code collisions. Please test everything. I tested this quite well and next to the fact that the microbugs are gone I can't spot any issues.

besynnerlig:
Thanks! It seems to work now, the "time on" fields get the current time but only if I don't have a logging QSO windows opened already. If I already have a QSO logging window open (https://wavelog.MYSERVER/index.php/qso?manual=0) the new callsign appears but the "Time on" and probably "Date" doesn't update to current time.

HB9HIL:
> Thanks! It seems to work now, the "time on" fields get the current time but only if I don't have a logging QSO windows opened already. If I already have a QSO logging window open (https://wavelog.MYSERVER/index.php/qso?manual=0) the new callsign appears but the "Time on" and probably "Date" doesn't update to current time.

again the question: did you clear the cache? we have some experience with that xD

HB9HIL:
The point is. When you have a LIVE QSO window open the time should run automatically

HB9HIL:
ezgif-2-b1b0724528

HB9HIL:
Here again the correct behaviour 


https://github.com/user-attachments/assets/cb1fb8d3-c86b-4875-a4cb-8c3da67a1e9b

HB9HIL:
Things which were changed in this PR and needs to be testet:

- [ ] New feature with logging URL. See example URL in the PR description
- [ ] QSO broadcast logic from bandlist to QSO logging view

besynnerlig:
Good morning! Thank you so much for all the hard work on this PR. I did another test this morning, just to be sure, and merged the PR again after clearing my Firefox browser cache. Everything works as you showed in your video, the date and time both display correctly at first.

When I work callsign, I typically click on the spot (if available), and the browser opens a new logging tab with the correct date and time, which is great. However, it's quite common that no QSO takes place, so I sometimes leave the logging tab open **without logging anything**. Later, when I click on another spotted callsign, the existing tab is reused, which is great! But I noticed that while the new callsign appears, the "Time On" field doesn’t update to the current time.

I’ve attached a video that shows what I mean. Thanks again for your time and for all your help!

https://github.com/user-attachments/assets/968e84c4-1436-43ea-8786-6bd25eb2bf68

int2001 (approved):

int2001 (changes requested):
Ok, works fast, but:

Click something at DXCluster - call will be copied to QSO-Tab
Now go to Overview/LBA/Whatever at that QSO-Tab.
Go again to Live-QSO. Call is still filled (even on that fresh page) with the Call.

HB9HIL:
> Good morning! Thank you so much for all the hard work on this PR. I did another test this morning, just to be sure, and merged the PR again after clearing my Firefox browser cache. Everything works as you showed in your video, the date and time both display correctly at first.
> 
> When I work callsign, I typically click on the spot (if available), and the browser opens a new logging tab with the correct date and time, which is great. However, it's quite common that no QSO takes place, so I sometimes leave the logging tab open **without logging anything**. Later, when I click on another spotted callsign, the existing tab is reused, which is great! But I noticed that while the new callsign appears, the "Time On" field doesn’t update to the current time.
> 
> I’ve attached a video that shows what I mean. Thanks again for your time and for all your help!
> 
>  Recording.2024-10-05.085232.mp4 
> EDIT: I did not notice "New feature with logging URL. See example URL in the PR description", I will check that out now.

Ahh okay so you mean

LIVE QSO Tab open and a callsign already FILLED

then using the "Log QSO URL" and the time does not get updated

--

Well this one would need a discussion. 
Shall we "reset" all fields and clear the "not-logged" callsign out
or
Shall we open a new tab if in the existing one the callsign field is not empty

HB9HIL:
> Ok, works fast, but:
> 
> Click something at DXCluster - call will be copied to QSO-Tab Now go to Overview/LBA/Whatever at that QSO-Tab. Go again to Live-QSO. Call is still filled (even on that fresh page) with the Call.

Good find! 

Issue was that the `onmessage` listener for `ready` was initialized and sends the same call over and over again. Is fixed now so the call will be sent once when clicking on it in the bandlist

assets/js/sections/qso.js
```diff
@@ -1,393 +1,1216 @@
...
-				$(this).removeClass('changed');
+var bc = new BroadcastChannel('qso_wish');
+bc.onmessage = function (ev) {
+	if (qso_manual == 0) {
+		if (ev.data.ping) {
+			let message = {};
+			message.pong = true;
+			bc.postMessage(message);
+		} else {
+			// console.log(ev.data);
```
HB9HIL, line 280: @besynnerlig 

Sugestion 1: "Clear existing callsign and fill the new one"

```suggestion
			// console.log(ev.data);
			if ($("#callsign").val() != "") {
				reset_fields();
			}
```

HB9HIL:
@besynnerlig 

My personal opinion: Solution 1 (clearing not-logged data and replace with new) is the better solution instead opening a new tab if the operator did not click save. This is maybe a bit rude, on the other hand we can't take responsibility for everything 😂

besynnerlig:
> @besynnerlig
> 
> My personal opinion: Solution 1 (clearing not-logged data and replace with new) is the better solution instead opening a new tab if the operator did not click save. This is maybe a bit rude, on the other hand we can't take responsibility for everything 😂

I agree 100%. I will check it out immediately!

besynnerlig:
I've tried it but the "Time On" field remains with the old data when the new callsign data appears (I have cleared my browser cache)

HB9HIL:
strange... could you describe more precise or make a screen recording which tabs are open

besynnerlig:
> strange... could you describe more precise or make a screen recording which tabs are open

I've pulled the PR a gain and merged it. I even checked that the code changes in commit eb09ffb47e2ab4137ca514f93a293125069bfebf are applied to my source in qso.js I have cleared the cache....

1. I start with only https://wavelog.MYSERVER/index.php open in a Firefox tab

2. I click a spot in my SmartSDR Panadapter and a new logging window opens in FF Tab with URL: https://wavelog.MYSERVER/index.php/qso?manual=0 .
 image

3. I don't log that callsign shown above.
Then I click on another spot.  The same tab gets new data. 
image

but as you can see the "Time On" is a remain from the previous QSO.
 
On both images above the "Time On" field reads: 09:02:10

I will try clearing the cache again and restart the browser....

int2001 (approved):

int2001 (approved):
Tested a lot with QSO. Everything seems to work. Couldn't test everything.
Explicit tested:
- Prefill Call via Link
- DXCluster --> QSO
- "Normal" logging with Reset, etc.

Codebase too huge to check. But seems well done

besynnerlig:
I’m happy to report that clearing the cache and restarting Firefox completely solved the issue! It’s working perfectly now. I realize I should have restarted Firefox after my earlier attempts to clear the cache, but I’m just so relieved and grateful that it’s all resolved now! Thank you so much for your help and patience!

HB9HIL:
@phl0 and @AndreasK79 

Do you want to have a look aswell?

phl0 (approved):
Tested various variants. All working fine as far as I can see.

HB9HIL:
@besynnerlig 

MERGE TIME

PR #980, opened 2024-09-20
Mark QSOs without Band (as reported from qrz) as invalid at qrz

int2001:
Mark QSOs without Band (as reported from qrz) as invalid at qrz.

QRZ reports a failure, if one tries to upload a QSO without band. in rare circumstances the QSO may exist without a proper freq or band-information.
In this case QRZ_QSL_SENT will be marked as "I"nvalid


phl0 (approved):
Makes sense. 👍

PR #988, opened 2024-09-21
QRZ state modified on quick buttons

HB9HIL:
When editing a QSO, the QRZ.com Upload state is changed to "modified" to tell QRZ.com there are some new data. 

But when using the quick buttons in the QSO details dialog the state does not get changed. This is classified as a bug. 

image

For the detailed changes to fix this bug see the second commit seperatly

You will notice that this appears to be a really big PR. I assume to "ignore whitespace" during review ([click here to see without whitespaces](https://github.com/wavelog/wavelog/pull/988/files?diff=split&w=1))

The reason I run the code formatter over the logbook model is that over the years a lot of bad code formattings took place. The tab-spaces to not match, a lot of braces are different then the way we agreed (k&r). This is really annoying when editing a file. I hope this will be accepted..

image

image

### DIFF
--- a/application/models/Logbook_model.php
+++ b/application/models/Logbook_model.php
@@ -2,639 +2,636 @@
 
 class Logbook_model extends CI_Model {
 
-	private $station_result=[];
+	private $station_result = [];
 	public function __construct() {
 		$this->oop_populate_modes();
 	}
 
-	private $oop_modes=[];
+	private $oop_modes = [];
 	private function oop_populate_modes() {
 		$r = $this->db->get('adif_modes');
-		foreach($r->result_array() as $row){
-			$this->oop_modes[$row['submode']][]=($row['mode'] ?? '');
+		foreach ($r->result_array() as $row) {
+			$this->oop_modes[$row['submode']][] = ($row['mode'] ?? '');
 		}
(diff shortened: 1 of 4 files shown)


int2001:
Formatting: Okay
I spotted a change regarding LEO/MEO at activators as well in this PR. It isn't documented?! @phl0 can you have a look around L668?

Regarding the original purpose: Seems legit to update QRZ on paperQSL.

Besides those things i am going to test during the day, but i'd like to make sure the Rover-Activations are still fine.

int2001:
At Overview and all other buttons this works.
Cosmetics-Bug: M-State is only shown after Refresh.

But when setting QSL-Sent via LBA, there's no change.

HB9HIL:
> At Overview and all other buttons this works. Cosmetics-Bug: M-State is only shown after Refresh.
> 
> But when setting QSL-Sent via LBA, there's no change.

fixed. Please test agn

int2001 (approved):
tested it: works from overview / detail-view as well as from LBA.
Good work.

due to the massive changes (i know "only" reformatting) i can't be 110% sure that something else is broken. but hey: we have dev ;)

PR #996, opened 2024-09-24
Refactor SAT stuff

phl0:
Cleaning up a few things:

- [x] Make naming of satellites consistent (use LoTW names where assigned)
- [x] Rename SAT_NAME in log table where non-LoTW names were stored
- [x] Add column to satellite table showing if LoTW accepted or not
- [x] Ignore non-LoTW accepted satellites in LoTW upload


phl0:
This is now brushed up and imho ready to test. Would like at least @int2001 to give it a go :)

int2001 (approved):
tested nearly every view. Well done. Huge effort for that. but it's worth it.

AndreasK79 (approved):

PR #1002, opened 2024-09-26
Use harmonized line endings in ADIF exports

phl0:
This harmonizes line ending char(s) in ADIF exports. Headers were exported with \n only while QSO lines used \r\n. This now uses \r\n for all lines in ADIF exports.


int2001 (approved):
works as expected

PR #1003, opened 2024-09-26
Call CAT first, even if QSO-Window doesn't exist (yet)

int2001:
Old behaviour:
- if QSO-Window does not exist, the CAT-Call isn't made and the window will be opened.

New behaviour:
- CAT-Call is made first before QSO-Window-logic takes effect.


AndreasK79 (approved):

phl0 (approved):
works for me

Issue #983, opened 2024-09-20
API callsign lookup function always returns workedBefore: false,


besynnerlig:
The API callsign lookup function in application/controllers/Api.php always returns workedBefore: false, even for callsigns that have been worked before. The expectation is that the field should accurately reflect whether the callsign has been previously logged as a QSO.

**Version Information**

    OS: Debian 12
    PHP: 8.2.20
    MySQL/MariaDB: 10.11.6-MariaDB-0+deb12u1
    Wavelog: 1.8.5

**Steps to Reproduce**

1. Use a third-party app to post a callsign lookup request to the API.
2. Inspect the returned result.
3. Notice that workedBefore is always false, even for callsigns that have been worked before.

Expected Behavior workedBefore should return true for callsigns that have been worked before and are present in the log.

Screenshots
![lookup](https://github.com/user-attachments/assets/164a144c-6d21-46c0-80ed-f947b498ffc0)

Desktop Information:

    OS: Ubuntu 22.04
    Browser: Firefox
    Version: 117.0.1

Smartphone Information: N/A

**Additional Context**

This bug affects the callsign lookup API, where workedBefore always returns false regardless of whether the callsign is in the logbook or not. This behavior creates issues for external services relying on accurate lookup results.

int2001:
Which API are you calling?
What does our Wiki/Documentation say about the API-Endpoint you're using?

besynnerlig:
Dear Joerg, you edited your reply before I posted my answer, but here it is.

Thank you so much for your quick response and for pointing me in the right direction.

To clarify, I am performing a DXCC lookup for a callsign using the same API that WaveLogGate relies on, located at https://wavelog.example.com/index.php/api/lookup. I assumed this was the appropriate endpoint, as it's being used successfully by other applications in the same ecosystem.

Regarding the Wiki and documentation, I did check the page at https://github.com/wavelog/wavelog/wiki/API but couldn’t find any specific information about this particular API endpoint. I realize that sometimes documentation can lag behind the codebase, which is perfectly understandable in an active project like this. That said, if I missed something or used an undocumented part of the system, please let me know. My intention is certainly not to overstep or misuse any functionality—just to be helpful and contribute where I can.

I appreciate your guidance, and I’m happy to make any changes or adjustments as needed. I’m definitely not here to complain, and I’d love to continue assisting in any way that benefits the project.

Thank you again for your support and for maintaining such a valuable tool!

Warm regards

int2001:
[...]

PR #1017, opened 2024-09-30
Added a new api: /api/private_lookup

int2001:
This api checks the (API-)owners logbook for confirmations.
Difference between `/api/lookup` is the private-part. `/api/lookup` is used by other 3rd-party tools (like DXCluster) and cannot be changed.

The PR addresses #983 

Details:

`curl https://[URL]/api/private_lookup -X POST -d '{"key":"[key]","callsign":"VK4XY","band":"20m","mode":"SSB"}'`

returns:
```
{
    "callsign": "VK4XY",
    "dxcc": "AUSTRALIA",
    "dxcc_id": "150",
    "dxcc_lat": "-22",
    "dxcc_long": "135",
    "dxcc_cqz": "30",
    "dxcc_flag": "\ud83c\udde6\ud83c\uddfa",
    "cont": "OC",
    "name": "",
    "gridsquare": "",
    "location": "",
    "iota_ref": "",
    "state": "",
    "us_county": "",
    "qsl_manager": "",
    "bearing": "",
    "call_worked": true,
    "call_worked_band": true,
    "call_worked_band_mode": true,
    "lotw_member": false,
    "dxcc_confirmed_on_band": true,
    "dxcc_confirmed_on_band_mode": true,
    "dxcc_confirmed": true,
    "call_confirmed": true,
    "call_confirmed_band": true,
    "call_confirmed_band_mode": true,
    "suffix_slash": ""
}
```

Mandatory Fields within payload (api will be documented after merge):
* key (your API-Key from AccountSettings --> API)
* callsign (the Call you want to check)

Optional Fields:
* band
   * if given: api checks if the DXCC of the call was confirmed for this band
   * if not given: dxcc_confirmed_on_band and dxcc_confirmed_on_band_mode will be false
* mode
   * if given: api checks if the DXCC of the call was confirmed for this band/mode combination
   * if not given: dxcc_confirmed_on_band_mode will be false
* station_ids (array!)
   * if given: api checks QSOs within those station_profiles. If ommited: Every station_id of the key-owner will be checked.
   * if given but no given ID was granted: Every station_id of the key-owner will be checked.
   * if given but only a few IDs were granted: The granted IDs will be checked
   * if given as string and not array: Every station_id of the key-owner will be checked.

The old Fields work as before. so workedBefore represents if the explicit Call was worked before in general

### DIFF
--- a/application/controllers/Api.php
+++ b/application/controllers/Api.php
@@ -592,11 +592,216 @@ class API extends CI_Controller {
 
 	}
 
-	function lookup() {
+	function private_lookup() {
 		/*
 		 *
 		 *	Callsign lookup function for Wavelogs logging page or thirdparty systems
 		 *	which want to show previous QSO data on their system.
+		 *	This endpoint returns detailled data about your personal log!
+		 *
+		 */
+
+		// Make sure users logged in
+		$raw_input = json_decode(file_get_contents("php://input"), true);
+		$user_id='';
+		$this->load->model('user_model');
+		if (!( $this->user_model->authorize($this->config->item('auth_mode') ))) {				// User not authorized?
+			$no_auth=true;
+			$this->load->model('api_model');
+			if (!( ((isset($raw_input['key'])) && ($this->api_model->authorize($raw_input['key']) > 0) ))) {			// Key invalid?
+				$no_auth=true;
+			} else {
+				$no_auth=false;
+				$user_id=$this->api_model->key_userid($raw_input['key']);
+			}
+			if ($no_auth) {
+				http_response_code(401);
+				echo json_encode(['status' => 'failed', 'reason' => "missing api key or session"]);
+				die();
+			}
+		} else {
+			$user_id=$this->session->userdata('user_id');
+		}
+
+		$this->load->model('stations');
+		$all_station_ids=$this->stations->all_station_ids_of_user($user_id);
+		
+		if ((array_key_exists('station_ids',$raw_input)) && (is_array($raw_input['station_ids']))) {		// Special station_ids needed and it is an array?
+			$a_station_ids=[];
+			foreach ($raw_input['station_ids'] as $stationid) {	// Check for grants to given station_id
+				if ($this->stations->check_station_against_user($stationid, $user_id)) {
+					$a_station_ids[]=$stationid;
+				}
+			}
+			$station_ids=implode(', ', $a_station_ids);
+		} else {
+			$station_ids=$all_station_ids;				// Take all of user if no station_ids were given
+		}
+
+		if ($station_ids == '') {					// No station_ids found for user or no station_id of given ones were granted? exit!
+			http_response_code(200);
+			echo json_encode(['status' => 'failed', 'reason' => "no station_profiles are matching the User with this API-Key"]);
+			die();
+		}
+
+		if (array_key_exists('band',$raw_input)) {
+			$band=$raw_input['band'];
+		} else {
+			$band='NO_BAND';
+		}
+		if (array_key_exists('mode',$raw_input)) {
+			$mode=$raw_input['mode'];
+		} else {
+			$mode='NO_MODE';
+		}
+
+		$lookup_callsign = strtoupper($raw_input['callsign'] ?? '');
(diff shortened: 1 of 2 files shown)


besynnerlig:
I’ve done some initial testing, and I really appreciate the hard work that’s been put into this.

In [this comment](https://github.com/wavelog/wavelog/issues/983#issuecomment-2382345541), you mentioned the fields `call_worked_on_band_mode` (boolean) and `call_worked` (boolean). It’s clear that the `workedBefore` field corresponds to `call_worked`, but `call_worked_on_band_mode` is not present. That said, a field like `call_worked_on_band_mode` or even `call_worked_on_band` would be incredibly useful for my app. I’d be very grateful if this could be considered.

As you mentioned earlier, DXCC confirmation depends on how I configure "Default QSL-Methods." I’ve now set this to LoTW only, which gives me the desired result. While I no longer see the other QSL services in the Logbook, I can live with that since all the QSL services I use still appear in the advanced Logbook.

Thanks again for all the great work on this. The next step for me is to integrate the new API into my application, and I’ll proceed with that shortly and report here.... I’ll be sure to follow up with my results and any feedback.

int2001:
indeed. i mentioned call but meant DXCC. Sorry for that (will correct the description).
Adding more and more fields (like worked_on_band and so on) will slow down the query. Let's start with that what is currently given with the PR

i'd suggest to start with the given dxcc_fields and the already existing worked_before

int2001:
@besynnerlig - i managed to add the 3 requested fields without performance-loss.

API returns now (additional to the existing fields, and depending on input):
- call_cnf
- call_cnf_band
- call_cnf_band_mode

all three boolean. same logic as with the DXCC.

pse test.

phl0 (approved):
Tested and works on my system(s)

besynnerlig:
> @besynnerlig - i managed to add the 3 requested fields without performance-loss.
> 
> API returns now (additional to the existing fields, and depending on input):
> 
>     * call_cnf
> 
>     * call_cnf_band
> 
>     * call_cnf_band_mode
> 
> 
> all three boolean. same logic as with the DXCC.
> 
> pse test.

This is fantastic! I've already integrated the new features into my app, and everything is working smoothly so far. I truly appreciate your help in pushing these updates forward!

However, I did notice one discrepancy while testing. When running the API with a callsign like TI5/N3KS using curl, it returns "lotw_member": "9". Meanwhile, when logging a QSO with the same callsign and hovering over the LoTW button, it says LoTW User. Last upload was 2 days ago. A direct check on LoTW shows: Last upload for TI5/N3KS: 2024-09-29 17:27:08Z. I’m curious about where the "9" comes from, as I expected it to reflect either false or a number representing the number of days since the last upload.

You’ve also added the following fields:

    call_cnf
    call_cnf_band
    call_cnf_band_mode

These seem to be working well, but the data I primarily need is:

    call_worked (already available as workedBefore)
    call_worked_band
    call_worked_band_mode

The reason the 'worked' state is more relevant than 'confirmed' when displaying spots on the SmartSDR Panadapter is that I (and likely others) aim to avoid contacting the same callsign on the same band and mode, regardless of whether the QSO has been logged by the other station.

As a side note, please forgive me if it's a bit OT, the Wave-Flex Integrator is almost ready for beta testing! I’m planning to reach out to the FlexRadio community to find testers. The program also simplifies things by eliminating the need for external CAT control software, since the communication between FlexRadio and Wavelog is now integrated hence reducing potential points of failure.

Thank you once again for your support and invaluable contributions! Below are a few screenshots.

image


image
Spot colors can be customized and is not optmal on the above screenshot.

int2001:
c'mon. will add worked_band and worked_band_mode later. but at sometime i'd like to finish this thing.

regarding the different LoTW-Times: Wavelog downloads the lotw-stats by itself (via cron). They may differ from the website.

besynnerlig:
Thank you so much for committing 1672a80251148746a3f50a96ff66145d91196198, @int2001! It truly feels like Christmas came early! I genuinely appreciate all the hard work and dedication you’ve put into making these changes. Your effort has made a big difference!

HB9HIL (approved):
nice piece :)

PR #1018, opened 2024-09-30
Fix empty eQSL received date

phl0:
Empty eqsl.cc received date results in PHP error:

image


int2001 (approved):
works.

b4: 
image

after patch:
image

PR #1019, opened 2024-09-30
Change Gridmaster Award text for non-US DXCCs

phl0:
This changed the text for the non-US Gridmaster awards.


int2001 (approved):
tsted - works and is generic for other Gridmasters that are not supported by any official league

PR #1023, opened 2024-10-01
Empty eQSL creds if obviously wrong

int2001:
Just a quick fix which removes the eQSL-Creds if they're reported as wrong from eQSL.

Fires at:
- cron upload
- manual upload

not a final solution, but heals the fact that huge instances are spamming eQSL over and over again.

Perhaps sth. for Version 2.0.: Some kind of message-center where we display failures out of crons/background-jobs.

PS: Diff is small. Compare without whitespaces


phl0 (approved):
Works as expected and removes my (wrong) eqsl.cc credentials :)

PR #1024, opened 2024-10-01
Make sure distance is numeric otherwise set to null

int2001:
Make sure distance is numeric otherwise set to null.

Affects:
- ADIF Import
- API


AndreasK79 (approved):

AndreasK79:
More fences is good :)

PR #1026, opened 2024-10-01
Numeric fence for RX_PWR

int2001:
Heals this one:

`Import error: Incorrect double value: '' for column COL_RX_PWR`


AndreasK79:
#################
#     We like fences    #
#################

PR #1031, opened 2024-10-02
Correct VFDB Link (missing dot for sub domain)

phl0:


int2001 (approved):
don't know how to test. but z01.vfdb.org gives a result. so seems right.

PR #1033, opened 2024-10-04
Set curl-Useragent for qrz-requests as recommended

int2001:
Add the Useragent according to QRZ-Guidelines properly:

https://forums.qrz.com/index.php?threads/logbook-api-documenation-update-application-identification-mostly-for-developers.931366/


phl0 (changes requested):
Code works fine. Just some minor improvements in https://github.com/int2001/wavelog/pull/14. But basically good to go imho.

phl0 (approved):

PR #1035, opened 2024-10-04
Lotw cert hint

phl0:
(Hopefully) Prevent some more user tickets about failed LoTW cert uploads ...

image


int2001 (approved):
Tested with German-Lang as well as with default one. Works and should(!) help the users.

PR #1038, opened 2024-10-04
Handle UTF8-Chars for eQSL-Credentials at upload correct

int2001:
When set a password with special UTF8-Chars like `é` or `ë` for eQSL the Upload failed.
This one should fix it.

### DIFF
--- a/application/models/Eqslmethods_model.php
+++ b/application/models/Eqslmethods_model.php
@@ -41,7 +41,7 @@ class Eqslmethods_model extends CI_Model {
 
 	function uploadUser($userid, $username, $password) {
 		$data['user_eqsl_name'] = $this->security->xss_clean($username);
-		$data['user_eqsl_password'] = html_entity_decode($this->security->xss_clean($password));
+		$data['user_eqsl_password'] = $password;
 		$clean_userid = $this->security->xss_clean($userid);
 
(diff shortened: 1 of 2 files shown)


application/models/Eqslmethods_model.php
```diff
@@ -41,15 +41,15 @@ function downloadUser($userid, $username, $password) {
 
 	function uploadUser($userid, $username, $password) {
 		$data['user_eqsl_name'] = $this->security->xss_clean($username);
-		$data['user_eqsl_password'] = html_entity_decode($this->security->xss_clean($password));
```
HB9HIL, line 44: this was added by me 2 months ago in commit 7c82d51916ae0f20d6d65990cd608b4b704c823c

reason was that special chars in the password caused the upload to break. See PR https://github.com/wavelog/wavelog/pull/696

this may solves the UTF8 issue but brings back the password encoding issue

HB9HIL: corresponding issue https://github.com/wavelog/wavelog/issues/695

int2001: Tested this, too.
Characters like `<` or `>` are converted with the new `rawurlencode` at ADIF-Generation.

phl0:
Though I am late to the show I just tested with various passwords and all major special characters worked. Tnx fer fixing this :)

PR #1039, opened 2024-10-04
[CABRILLO] Fixed a small bug and added some more optional fields

HB9HIL:
There was a bug present that the value of "CLUB" was not written to the cbr file. 

In addition I added some more fields

LOCATION  - This field is needed in CQ and ARRL contests, RSGB-IOTA and RDXC contest
CATEGORY TIME - some contests need this to determine the operation duration. 
CERTIFICATE - Some contests allow to request a certificate over the cabrillo file. This is the field for that. Default is `""` and hidden in the cabrillo


int2001 (approved):

phl0 (commented):
Tested and works here.

PR #1040, opened 2024-10-05
Remove keeping the session date for POST logging QSOs

phl0:
Since we now log/store the QSO via AJAX request we do not need to re-read the date/time from session data for POST logging QSOs. Instead the date/time is kept because the page is not reloaded completely.


phl0:
Integrated in https://github.com/wavelog/wavelog/pull/978 to prevent merge conflict.

PR #1044, opened 2024-10-06
Prevent endless looping in case of br0ken ADIFs

int2001:
In some cases ADIFs aren't complete (User provided or from 3rd-party-services).
A bug in our parser produced an endless loop in this case.

this one prevents the endless-loop


HB9HIL (approved):
good solution to fix "Index out of bounds" errors

phl0 (approved):
Patch approved. Prevents endless loop on qrz download

Issue #1046, opened 2024-10-06
Advanced Logbook - Globe Map qso path no correct with far side of earth station


wafer-li:
**This is for reporting bugs with the code, if this issue is regarding setting up on your local server please use the discussions area**

**Describe the bug**
If you contact a station located at the far side of the earth, the Globe map will show its path line straight through inside of the sphere.

Showing as the following image, the red line seems straight through the earth sphere at the yellow area.

![捕获](https://github.com/user-attachments/assets/76aebcc8-020b-4d8e-ad48-b49fc5818f19)

**Expected Behavior**
The QSO Path should be visible in all its path, just like the qso with the near side of earth station.

violarulan:
#1048 should brings a fix.

wafer-li:
Upgraded to 1.8.6, maybe 0.35 still not enough

<img width="822" alt="image" src="https://github.com/user-attachments/assets/82bed282-b5b2-4ca6-8b26-aec52a15d0f4">

You could see the sink area shrink, but still below the horizon

wafer-li:
> Upgraded to 1.8.6, maybe 0.35 still not enough
> 
[...]

PR #1048, opened 2024-10-06
fix: globe map pathline #1046

violarulan:
#1046. Originally reported by @wafer-li.

According  to the Globe.gl [documentation](https://globe.gl/):

> `arcAltitudeAutoScale`: Arc object accessor function, attribute or a numeric constant for the scale of the arc’s automatic altitude, in terms of units of the great-arc distance between the two points. A value of `1` indicates the arc should be as high as its length on the ground. Only applicable if `arcAltitude` is not set.

A `0.35` should fix the bug.


HB9HIL (approved):
works :) thx for that fix!

PR #1053, opened 2024-10-07
[Advanced Logbook] Added more sorting for label print

AndreasK79:
Added missing sorting so that multiple QSOs would print on same label.


int2001 (approved):
works

PR #1054, opened 2024-10-07
Session Management - Too many old session files

HB9HIL:
CodeIgniter 3 relies on PHP's native grabage collector. But the native garbage collector is disabled by default since the default probability the gc to run is 0/1000 which means it basically runs never. 

This leads into thousands of files in /tmp which won't get deleted unless you reboot the server or delete them manually. Especially on big installations which have a really long uptime this will cost a lot of performance. 

To fix this we set the probability of the GC to run manually to 1/1000 which means that approx. every 1000 request will trigger the GC to clean up stuff in /tmp. 

This does not affect the "Keep-Login" Feature or any active session. 

PS: The command `gc_collect_cycles();` only cleans up memory and has no effect for the sessions. 

More Information you can find here:
https://www.php.net/manual/en/session.configuration.php#ini.session.gc-probability
https://www.php.net/manual/en/session.configuration.php#ini.session.gc-divisor
https://osvaldas.info/enabling-codeigniters-garbage-collector/


HB9HIL:
Tnx to @m0urs for this hint

int2001 (approved):
This one works.

BUT:
If the user setted a different lifetime at php.ini (which overrides this one), the sessions are still growing. For the "standard-intance-owner" this one is fine.

Issue #1055, opened 2024-10-08
Fix certain QSOs are not shown in gridmap


violarulan:
This is another part of #992 which I forgot to test.

We fixed that some unnecessary QSOs were shown in 2-digit grid locator condition by only appending the `%` to the end of the query.

Now if we have a QSO whose GL are more than one, for example `PN17,PN27,PN16,PN26` (right at 4 grids intersection), we can only see the QSO in the pop out list when clicking the `PN17` grid, nothing shown when clicking `PN27` or `PN16` or `PN26`. 
Will try to submit a PR later this day.

phl0:
I can at least confirm the bug with a QSO on the grid intersection of four grids. Will test the patch.

phl0:
Fix merged to dev in https://github.com/wavelog/wavelog/pull/1056.

PR #1056, opened 2024-10-08
Fix: certain QSOs are not shown in gridmap QSO list

violarulan:
Bug description: #1055 

Workaround: By adding a comma to the search pattern, we can search for QSOs with more than one grid locator.

i.e. There's a QSO logged as `PN17,PN27,PN16,PN26`, if we click:
- PN: search pattern: `PN%` or `%,PN%`, (we described it in #992) the first condition works
- PN17: search pattern: `PN17%` or `%,PN17%`, the first condition works
- PN27: search pattern: `PN27%` or `%,PN27%`, the second condition works
- PN16: search pattern: `PN16%` or `%,PN16%`, the second condition works
- PN26: search pattern: `PN26%` or `%,PN26%`, the second condition works

Can you guys help me to test it?


phl0 (approved):
Confirmed that this fixes the issue. Good find and thanks for fixing this.

PR #1060, opened 2024-10-09
Add None to DXCCs

int2001:
ADIF_DXCC = 0 was not part of the dxcc_entities table.

So a lot of Users skipped the step to fill in their own DXCC. As a result some exports weren't working (That was a bug).

This one adds the dxcc "None" hard to the DXCC-Table and forces the User to chose it's own DXCC

### DIFF
--- a/application/config/migration.php
+++ b/application/config/migration.php
@@ -22,7 +22,7 @@ $config['migration_enabled'] = TRUE;
 |
 */
 
-$config['migration_version'] = 221;
+$config['migration_version'] = 222;
 
 /*
 |--------------------------------------------------------------------------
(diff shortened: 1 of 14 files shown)


int2001:
@HB9HIL could you pse add a check for the installer? i'm to stupid for that JS-Magic ;)

phl0 (approved):
Works fine so far.

AndreasK79 (approved):

HB9HIL:
Just added a DXCC requirement to the installer

HB9HIL:
PS: There is no DXCC "NONE" for /MM or /AM in the installer, as the callsign should always be the personal operator callsign without any pre- or suffixes. Therefor it always has a DXCC

Issue #1062, opened 2024-10-09
Country broken in QSO details page (left join two columns with same name?)


violarulan:
_I'm here again, still a QSO detail page bug :)_

In Gridmap page, if we click on a grid, a detail page shows.

If we select `SAT` in Band, then the `Country` column becomes the Satellite name.

![Snipaste_2024-10-09_22-49-52](https://github.com/user-attachments/assets/56248dbf-0be1-404f-afdf-469edd4438ec)

The `dxcc_entities` and the `satellite` tables both have the `name` column, causing `name` overwritten?

The `Country` also uses `name` to pass the value.

https://github.com/wavelog/wavelog/blob/536191802ffa68c925151ffc4e80c3721d429c3e/application/views/view_log/partial/log_ajax.php#L29

I think the problem is at https://github.com/wavelog/wavelog/blob/536191802ffa68c925151ffc4e80c3721d429c3e/application/models/Logbook_model.php#L448

I don't know what does this line of code do. But if I comment this line, it seems working as expected.

phl0:
Tnx report. I will check.

phl0:
Fix proposed in https://github.com/wavelog/wavelog/pull/1063.
(just leaving out the join on satellite table would break SAT orbit type filters).

phl0:
Patch is in dev branch now.

PR #1063, opened 2024-10-09
Only select needed columns (and avoid dupe names)

phl0:
resolves https://github.com/wavelog/wavelog/issues/1062


int2001 (approved):
took a bit to find the edgecase. works

PR #1064, opened 2024-10-10
Use QSLRDATE provided by eqsl.cc (current date only as fallback)

phl0:
eQSL does now provide the EQSL_QSLRDATE field as date when the eQSL card was received. So we should use this date instead of the current date (at the time we are importing).


int2001 (approved):
tested it - works.

fallback is nice as well

PR #1065, opened 2024-10-10
Show fail date only if failed indeed

phl0:
Prevent showing last fail date if LoTW upload never failed before (and shows 1970-01-01).

image


int2001 (approved):
wks

HB9HIL (approved):

Issue #1066, opened 2024-10-10
API/QSO - Import Count Mismatch: Successful Response Despite No QSO Import


besynnerlig:
I am using the API/QSO endpoint as described in the [Wavelog Wiki](https://github.com/wavelog/wavelog/wiki/API#apiqso) for logging QSOs from a third-party app into Wavelog.

Following the example provided in the Wiki, I use the command below:

```bash
curl -X POST https://<WAVELOG_URL>/index.php/api/qso -H "Content-Type: application/json" -H "Accept: application/json" -d "{"key":"<API_KEY>","station_profile_id":"<STATION_PROFILE_ID>","type":"adif","string":"<call:5>N9EAT<band:4>70cm<mode:3>SSB<freq:10>432.166976<qso_date:8>20190616<time_on:6>170600<time_off:6>170600<rst_rcvd:2>59<rst_sent:2>55<qsl_rcvd:1>N<qsl_sent:1>N<country:24>United States Of America<gridsquare:4>EN42<sat_mode:3>U/V<sat_name:4>AO-7<prop_mode:3>SAT<name:5>Marty<eor>"}"
```

This command works perfectly, and the response is:

```json
{"status":"created","type":"adif","string":"","imported_count":1,"messages":[""]}
```

However, if I attempt to import a QSO where the station callsign in the ADI string does not match the active station profile (for example, `<station_callsign:4>SE6S`), the response from the API remains identical:

```json
{"status":"created","type":"adif","string":"","imported_count":1,"messages":[""]}
```

Even though the QSO does not import successfully due to the callsign mismatch, the `imported_count` field still shows a count of 1.

**Expected Behavior**  
I expected the `imported_count` field to reflect the actual number of successfully imported QSOs (in this case, it should be 0). An alternative approach could be to add a new field, such as `successful_imported_count`, to clarify how many QSOs were truly imported, allowing third-party apps to better handle these discrepancies.

**Version Information**  
- Wavelog: 1.8.6

**Steps to Reproduce**  
1. Use the API/QSO endpoint to submit an ADIF string with a station callsign that doesn't match the active station profile.  
2. Review the JSON response and observe that `imported_count` is 1, even though no QSOs were actually imported.

**Expected Behavior**  
The `imported_count` field should reflect 0 when a QSO fails to import due to a station callsign mismatch.

**Additional Context**  
[...]

PR #1067, opened 2024-10-10
API response to show processed and failed ADIF items

phl0:
Closes https://github.com/wavelog/wavelog/issues/1066

imported_count renamed to adif_count
added adif_errors to show number of not imported items

### DIFF
--- a/application/controllers/Api.php
+++ b/application/controllers/Api.php
@@ -208,7 +208,8 @@ class API extends CI_Controller {
 			$this->adif_parser->feed($obj['string']);
 			$obj['string']='';
 			$return_msg=[];
-			$return_count=0;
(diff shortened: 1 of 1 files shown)


besynnerlig:
**Thank you for this PR!** I really appreciate the quick action and the effort you put into it. I tested importing an invalid ADIF record (with an incorrect `station_callsign`), and it now returns the following response:

```
{"status":"created","type":"adif","string":"","adif_count":1,"adif_errors":1,"messages":[""]}
```

I also tested a valid ADIF:

```
{"status":"created","type":"adif","string":"","adif_count":1,"adif_errors":0,"messages":[""]}
```


This is exactly the behavior I was hoping for. Great work, and thanks again for this!

PR #1070, opened 2024-10-10
[Advanced Logbook] Add some validation to station grid when mapping

AndreasK79:
Stop log spamming if a user had entered an invalid gridsquare in station location when mapping QSOs in the Advanced Logbook.


phl0 (approved):
#fencestotheusers #approve

PR #1071, opened 2024-10-11
Align visitor view for logged-in users (cosmetical)

phl0:
Just a very minor thing that bugged me. The visitors view missed the SAT name in frequency columns for SAT QSOs.


HB9HIL (approved):
wks :)

PR #1072, opened 2024-10-11
[INSTALLER] Bugfixes and Callsign Check

HB9HIL:
Fixed some bugs in the installer and added a new callsign check (which should not contain any pre- or suffixes)

In Detail:

- [x] Migration 215 needed a fix as the contest_session table is empty when running the installer. This caused (sometimes?) the installer to fail. Why just sometimes I can't say but now it's more reliable. 
- [x] DXCC was not stored in LocalStorage
- [x] Added a check if the callsign contains a pre- or suffix. This should not be the case as the callsign is later used as operator callsign in the user account


AndreasK79 (approved):

AndreasK79:
Installer tested and working.

PR #1073, opened 2024-10-11
[Advanced Logbook] Added station name

AndreasK79:
Added station name to the columns in the advanced logbook:
image


HB9HIL (approved):
works :)

PR #1079, opened 2024-10-14
Use Binding for KML-Export

int2001:
Use Binding for KML-Export


int2001:
No danger here, since the function is never used.
But(!) the whole thing needs some refactoring.

PR #1088, opened 2024-10-17
Set QRZ only to Modify if uploaded before or invalid

int2001:
Set QRZ only to Modify if uploaded before or invalid.

Otherwise it'll be set to "M" before it was uploaded for the first time, which is wrong.


AndreasK79 (approved):

AndreasK79 (approved):

phl0 (approved):
I had that in mind but forgot to implement ... Works fine and totally makes sense.

PR #1098, opened 2024-10-19
Static Map API

HB9HIL:
# Static Map API
As a native replacement for our imager service we need a possibility to create static leaflet maps in PHP and serve them to be integrated in QRZ.com. As leaflet is based on Javascript this is no easy task. Therefore we now use a PHP Lib which creates static map images directly from OpenStreetMap.

image

I found this [library](https://github.com/DantSu/php-osm-static-api) which can solve our issues. It is currently a work in progress and it definitly needs some more work. 

## Development Progress:

- [x] Make Static API to work
- [x] Get last QSOs and display them on the map
- [x] Cache the map tiles so recalling the map is much faster
- [x] Improve the caching so it's based on the slug and parameters
- [x] Set Cache Path by already existing config option `$config['cache_path']`
- [x] Set the map center to the users QTH and **FIX THE EARTH WRAP ISSUE** (see below) **UPDATE: solved**
- [x] Also allow dark map based on the global option theme
- [x] Use the user defined markers.
- [x] `qsocount` needs to be used from the exportmap settings, `GET` option can override it
- [x] Add Wavelog watermark overlay or/and text

Can be done later:
- [x] Nightshadow overlay
- [x] Pathlines overlay
- [x] CQ Zones
- [x] ITU Zones

## Usage

Current usage is working with `GET` commands (easier for faster development). This is maybe not ideal as we already have settings for the exportmap available and can reuse them. So this needs to be changed. Anyway the current usage is:

`https://[wavelog url]/index.php/staticmap/render/[slug]`

Possible `GET` Options currently implemented:

> [!NOTE]
> All Options are optional! 
> The `qsocount` option overrides the user option for exportmaps

| option | value  |
|----|----|
| `qsocount`     | amount of QSOs displayed on the map (default: `250`)  |
| `band`     | display `qsocount` QSOs of the certain band  |
| `continent`     | you can display a dedicated continent view.  recommended to use with `qsocount`, use continent format like `AS`, `EU`, `NA`, ... |
| `hide_home`     | set to `1` if you want to hide your location on the map  |
| `theme` | overrides the global appeance theme. Set `light` or `dark` |
| `ns` | Enabled the Nightshadow Overlay. Set `1` or `0` | 
| `pl` | Shows the Pathlines Overlay. Set `1` or `0` | 
| `cqz` | Shows the CQ Zones Overlay. Set `1` or `0`. Attention: This one needs some time to render. Make sure Development Mode is disabled to use caching | 
| `ituz` | Shows the ITU Zones Overlay. Set `1` or `0`. Attention: This one needs some time to render. Make sure Development Mode is disabled to use caching | 
|  ...      |  other options can be added as needed  |

### Example

Definition: Show the last 500 QSO on the 160m Band with Pathlines enabled but Nightshadow disabled

`https://[wavelog url]/index.php/staticmap/render/[slug]?qsocount=500&band=160m&pl=1&ns=0`

### Caching

This feature uses three caches. 

The icons which are used in the map based on the user settings.
The maptiles are cached based on theme and maparea.
The Image itself is cached based on all options and logbook. 

To temporarily disable caching, enable the Development mode. To clear the image cache you can:

- log a qso or import an adif
- edit user settings
- edit the logbook
- edit a station location

To completely clear the cache remove the three folders: `fas_icons_cache`, `staticmap_images`, `tilecache`

## Earth Wrap Issue (Update: solved)

One of the bigger issue are currently to load the correct map tiles when the users QTH is not near 0° Lat / 0° Long
Here an example with Longitude of 110°

image

This breaks the caching and map. To solve that we need a logic to load the "other side" of the tiles. Default map center is currently 0° Lat / 0° Long to avoid this issue. I created an issue here https://github.com/DantSu/php-osm-static-api/issues/21


HB9HIL:
Earth Wrap Issue solved in commit https://github.com/wavelog/wavelog/pull/1098/commits/da3fbecf65432377d59c6169cb017b4caee5727e

image

HB9HIL:
The map center is now set to a latitude of 0, as latitude wrapping is not supported. The longitude is calculated as the center of all stations included in this logbook.

HB9HIL:
Caching improved

HB9HIL:
IMHO this is now ready 4 a first review although some things are still missing. But basic functionality is working!

Caching: 
- Checks if already a image exists, if yes return this -> performance due to only a file return
- If no Image exists, generate a new one. 
- Map Tiles get cached, so even a new image is generated this only affects the markers on the map if tiles already exist

HB9HIL:
image
Dark Mode is working. It is based on the global appearance option and the darkmode server is hardcoded. Light-Mode server is used from the one in database

HB9HIL:
Close.. 😄 
image

phl0 (approved):
Works and is rather fast on my system. Good work!

phl0 (approved):
Nice with hide_home now.

HB9HIL:
> Nice with hide_home now.

Thx. I also changed the filenames to something generic. placing the raw options in the filename looks weird xD

Example: `staticmap_log_72ec6904abfc.png`

phl0:
> > Nice with hide_home now.
> 
> Thx. I also changed the filenames to something generic. placing the raw options in the filename looks weird xD
> 
> Example: `staticmap_log_72ec6904abfc.png`

HB9HIL:
@AndreasK79  Nightshadow works

GET Option = `ns` 
value = `1` enabled, `0` disabled

if `NULL` it takes the user option from exportmaps

image

HB9HIL:
image

Pathlines and CQ Zone Overlays are working now

PR #1100, opened 2024-10-20
[QSLPrint] Added profile name. Fixed table columns when adding QSOs

AndreasK79:
Profile name added to columns.
Fixed a long outstanding bug when searching for additional QSOs and adding them to the table.
We had a column mismatch here.


int2001 (approved):
good find. works

m0urs:
Thanks a lot for adding the "Profile name" column! :-)

AndreasK79:
@m0urs no problem. Some things are quick to add. It also lead to a bugfix as well :)

PR #1103, opened 2024-10-21
Contest edit qso

AndreasK79:
Can now click on a call to bring up detail view, also edit qso.


HB9HIL (approved):
works awesome! fantastic job!

did also test changing exchange values

PR #1105, opened 2024-10-21
Fix radio value = 0

HB9HIL:
Fixes a bad bug which was introduced in #1060 and caused another issue which was solved in #1095 

Bug caused the whole CAT, Band.on(change) and QRG stuff to fail as there was no value for the radio.


int2001 (approved):
I feel ashamed... sorry

HB9HIL:
No reason for that. Mistakes happen :)

PR #1106, opened 2024-10-21
QSLMSG in API Imports

HB9HIL:
If a QSO was imported by the API, the default QSLMSG of the particular station location was not set. This fixes this and set's the default message, if one is set in the station location. If the ADIF already does contain a QSLMSG this one overrides the default in the Stationlocation

Other side changes:

- I found some `$this->logbook_model->method()` in the logbook model itself. This causes a slightly higher load. `$this->method()` is enough if you call methods in the same model
- Removed `$this->load->model('user_options_model');` as it's already loaded by the autoloader


int2001 (approved):
works.
tested:
- ADIF-Upload with and without QSLMsg
- Single-QSO API with and without QSLMsg

PR #1111, opened 2024-10-22
[Advanced Logbook] Added text to choose column when editing QSOs

AndreasK79:
Added text so that the user understands that a column must be chosen:
image
Save button is disabled until at column is chosen.


phl0 (approved):
Very nice :)

PR #1112, opened 2024-10-22
[WIP]: QO-100 Dx club instant upload

phl0:
The previous code allowed API QSOs top be uploaded to QO-100 Dx Club. After switching to bulk_import this functionality got lost. We should re-integrate that.


phl0:
This was a dead end. Better check for numbers of QSOs being sent via API and if only one skip batchmode (and get insert ID). See https://github.com/wavelog/wavelog/pull/1113.

PR #1113, opened 2024-10-22
Allow for export upon single QSO ADIF being sent to API

phl0:
(Working) implementation for https://github.com/wavelog/wavelog/pull/1112.

$user requested QO-100 Dx Club uploads for (single) QSOs being sent via API, e.g. Logging from WSJT-X improved. 

This disables skipExport for API calls and adds a distinction for number of ADIF records sent via API. If its only one QSO we insert directly (and this get the insert ID) instead of using batch mode. Getting the ID allows for exports like QO-100 Dx Club instant uploads for QSOs sent via API.

@int2001 @AndreasK79 @HB9HIL would like you to double check if we can invert skipExport here especially for multi-user envs like DCL.

### DIFF
--- a/application/controllers/Api.php
+++ b/application/controllers/Api.php
@@ -226,7 +226,7 @@ class API extends CI_Controller {
 				};
 				$record='';	// free memory
 				gc_collect_cycles();
-				$custom_errors = $this->logbook_model->import_bulk($alladif, $obj['station_profile_id'], false, false, false, false, false, false, false, true, false, true, false);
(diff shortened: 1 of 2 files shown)


int2001 (approved):
This works.
But - be aware of the fact that the amsat-notifier isn't triggered here.

I am fine with that as an interims-solution. But we should refactor the webadif-part to a logic which is already implemented at qrz/eqsl/lotw. Trigger it via cron as well and add a setting for the user where he's able to chose between RT-Enabled and Enabled

phl0:
Think I will skip this one and instead try to adapt webadif / QO-100 Dx Club to the existing structures.

PR #1117, opened 2024-10-22
Invalid CQ Zone Value

HB9HIL:
If the CQ Zone is `(result.callsign_cqz >= 1 && result.callsign_cqz <= 40)` we can safely use it, otherwise we should use the DXCC data

Classic Shit-In / Shit-Out by callbook provider


phl0 (approved):
Fix works.

PR #1126, opened 2024-10-24
Update/add MESAT and SONATE-2 and reset LoTW upload for those

phl0:
MESAT1 and SONATE-2 are now accepted on LoTW. This PR:

- Renames MESAT1/MESAT-1 to MO-122 in satellite table (if found)
- If not found adds that satellite to satellite table
- Renames SAT name in already logged QSOs
- Resets LoTW upload to N and LOTW upload date to empty (to enable re-upload with correct name)

Same goes for SONATE-2 (recognized as SONATE by LoTW).

### DIFF
--- a/application/config/migration.php
+++ b/application/config/migration.php
@@ -22,7 +22,7 @@ $config['migration_enabled'] = TRUE;
 |
 */
 
-$config['migration_version'] = 224;
+$config['migration_version'] = 225;
 
 /*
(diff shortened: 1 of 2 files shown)


int2001 (approved):
works.
tested (with SONATE and MESATE QSO's in log):
- Upgrade
- Downgrade
- Upgrade agn.

PR #1127, opened 2024-10-25
[QSO] Improvement and Bugfixes

HB9HIL:
This PR fixes some "microbugs" and brings some improvements to the QSO logging

### Standarize some functions

In some places we use `.change( function ( { } ) );` and in some places `.on(event, function ( { } ) );`. We should stick to the more modern one. `.change`, `click` and so on are deprecated.

### "Ø Problem"

As the replacement of 0 to Ø in the callsign input was broken, the callsign suggestion for these callsigns were broken aswell. Fixed that.

| Before Fix                                           | After Fix                                            |
|------------------------------------------------------|------------------------------------------------------|
| image | image |

### Dupes returned in SCP result

As you can see in the screenshots of the "Ø Problem", the callsigns were returned twice. This happens because we use two different SCP sources to get as much results as possible. Due the fact we are handling two types of zeros (0 and Ø) it could happen that a callsign was returned twice. Using the callsign as array key with the value true solves this issue. 

| Before Fix                                           | After Fix                                            |
|------------------------------------------------------|------------------------------------------------------|
| image | image |

### Fix highlighting in selectize dropdowns

After my qso.js restructuring all functions are now loaded correctly. But as this happens the `highlight()` function, which highlights parts of the callsigns in the callsign-suggestion is broken as this conflicts with the `highlight()` function of the selectize library. This results in a broken text-highlighting

| Before Fix                                           | After Fix                                            |
|------------------------------------------------------|------------------------------------------------------|
| image | image |

### Prevent lookups while resetting the fields

We have a racecondition here. It can happen, that the callsign lookup runs multiple times when switching between windows and/or try to reset the fields. Due the fact that the `reset_fields()` function is obviously much faster then a ajax call it happens that fields are getting filled after the reset was triggered. This is not really nice and is fixed with my last commit. 

https://github.com/user-attachments/assets/699229f5-8d1b-4729-a0c2-39003425f6c5

https://github.com/user-attachments/assets/cc8b0d4a-e260-4bb4-87da-ac708c13a459

### DIFF
--- a/application/controllers/Lookup.php
+++ b/application/controllers/Lookup.php
@@ -66,73 +66,67 @@ class Lookup extends CI_Controller {
 	public function scp() {
 		session_write_close();
 		$uppercase_callsign = strtoupper($this->input->post('callsign', TRUE) ?? '');
-
+	
 		// SCP results from logbook
 		$this->load->model('logbook_model');
-
+	
 		$arCalls = array();
-
+	
 		$query = $this->logbook_model->get_callsigns($uppercase_callsign);
-
-		foreach ($query->result() as $row)
-	    {
-	    	if (in_array($row->COL_CALL, $arCalls) == false)
-			{
-					$arCalls[] = str_replace('0', 'Ø', $row->COL_CALL);
-			}
-	    }
-
+	
+		foreach ($query->result() as $row) {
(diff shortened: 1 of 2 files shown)


phl0 (approved):
All tested successfully except the race condition would not happen ;)

PR #1128, opened 2024-10-25
[Quick Lookup] Added ITU zone

AndreasK79:
Since ITU zone is added several places, I added ITU zone to Quick Lookup as well.
image


phl0 (approved):
Works. Wanna add the popup link to the ITU zone column in LBA results as well?

image

AndreasK79:
@phl0 yes, link was planned all along. Refresh and test again :)

phl0 (approved):
Link lookup also works :)

PR #1135, opened 2024-10-28
[Advanced Logbook] Adjusted filters and column order

AndreasK79:
I made the following changes in the advanced logbook:
1. Filters are hidden if the corresponding column is hidden
2. Same column order in the table as in the option to show/hide column
3. Moved SOTA, DOK, WWFF and SIG to be after IOTA, so that most of the award stuff is grouped.


phl0 (approved):
Nice idea. Works as expected.

PR #1136, opened 2024-10-28
Translated E-Mails

HB9HIL:
Since Wavelog is multilingual, it makes sense to translate the emails as well. The language of the email is set to match the recipient’s preferred language.

Due this change the way to load and write emails changed a little bit. 

1. Instead loading the email itself with `$this->load->view()` we now use the new `$this->email->load()`
2. The email is written in PHP and returned as json

You can assess the email subject due `$message['subject']` and the message/body itself with `$message['body]`

A good example how a dynamic and more complex email looks like take a look at `application/views/email/oqrs_request.php`


phl0 (approved):
Yeah. Works fine so far.

PR #1142, opened 2024-10-31
[SimpleFLE] Fixed an issue at saving QSOs

HB9HIL:
It can happen that, if the network connection is slow, the SimpleFLE page gets cleared and reloaded before all QSOs were saved. Reason for that ins basically a bad code, which made one Ajax Call for each QSO (race-cond) 

I refactored that to send all qsos as json in ONE ajax call you the controller.


phl0 (changes requested):
QSOs get saved into the DB but an (empty) error message:

image

phl0:
I can provide the SFLE text via PM.

AndreasK79:
Got the same here with the sample data.

HB9HIL:
@phl0  and @AndreasK79 please test agn. Should be fixed

AndreasK79:
@HB9HIL error is gone now :)

phl0 (approved):
Fix confirmed!

image

PR #1146, opened 2024-11-01
[Advanced Logbook] Added tx power to column and edit

AndreasK79:
Station power is added to column in the table.
Power can also be edited via batch edit.


phl0 (approved):
Works. Good thing to have added :)

PR #1147, opened 2024-11-02
Exclude NONE DXCC for cfd exports

phl0:
We need to skip DXCC NONE for cfd exports.


int2001 (approved):
Currently (due to personal circumstances 😉) not able to check in a running system, but it looks good. TNX!

PR #1148, opened 2024-11-02
Clublog hint about app passwords if 2FA is enabled

HB9HIL:
image
Added a small info text below to push the user in the correct direction


phl0 (approved):
Good point.

PR #1149, opened 2024-11-03
Remove DXCC NONE from DXCC award

phl0:
As "NONE" is not valid for DXCC award we remove it from the list.

Note: The fetchdxcc function was not changed as it is used in other places as well where NONE should be included. So 0/NONE is removed from the result array of that function only for the DXCC award and its map.


int2001 (approved):
TIL: No array_splice needed. unset does the job as well. Looks good. TU!

PR #1150, opened 2024-11-03
(Re-)Include 160m into cfd exports

phl0:
160m was not included in cfd exports because it has been re-labelled as 'mf' in https://github.com/magicbug/Cloudlog/pull/2590 by @HB9HIL. So we have to manually add the band here (just as 6m).


int2001 (approved):
Yeah. 160m back in place. Tnx

PR #1152, opened 2024-11-03
Clear distance (and bearing display) if grid is empty(ed)

phl0:
If a callsign is entered and (wrong) grid info is fetched from callbook the op would delete the grid data from the input field. But the bearing and distance is already calculated based on the (wrong) callbook grid. Even with empty grid the wrong distance is then stored in the QSO:

https://github.com/user-attachments/assets/ac1889a4-44f2-4ba0-ad17-c060587914b5

Result:

Screenshot from 2024-11-03 14-28-00

With this PR the bearing and distance info is cleared if the grid locator field is empty and the op tabs out. Now no distance/bearing is displayed and no distance is stored along with the QSO.

### DIFF
--- a/assets/js/sections/qso.js
+++ b/assets/js/sections/qso.js
@@ -1004,6 +1004,13 @@ $("#locator").on("input focus", function () {
 	}
 });
 
+$("#locator").on("focusout", function () {
+	if ($(this).val().length == 0) {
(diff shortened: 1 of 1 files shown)


AndreasK79 (approved):

AndreasK79:
@phl0 nice find!

PR #1154, opened 2024-11-03
Add orbit type to static maps for SAT QSOs

phl0:
This adds an option 'orbit' to staticmap parameters allowing to create different maps based on orbit type for SAT QSOs.


HB9HIL (approved):

phl0:
Tnx for additions. Merge time!

PR #1155, opened 2024-11-03
Fix inconsistent font family in grid activators page

violarulan:
A minor bug: the grid column uses `monospace` font family (hard coded), while other columns don't.

I think it would be nicer to use the same font family in the whole table.

Snipaste_2024-11-03_22-43-30

### DIFF
--- a/application/views/activators/index.php
+++ b/application/views/activators/index.php
@@ -129,7 +129,7 @@ function write_activators($activators_array, $vucc_grids, $custom_date_format, $
                 <td>' . $i++ . '</td>
                 <td>' . $line[1] . '</td>
                 <td>' . $line[0] . '</td>
-                <td style="text-align: left; font-family: Consolas, monospace;">' . $line[2] . '</td>
+                <td style="text-align: left; font-family: Consolas, monospace; text-transform: full-width; letter-spacing: -.3em;">' . $line[2] . '</td>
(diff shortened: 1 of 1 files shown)


HB9HIL (approved):

HB9HIL:
@phl0 You are user of this page, please cnfm aswell

phl0:
To be honest: I don't like the view:

image

I will have a look if there are other options which allow for using the same fonts and keep the structure of the view.

phl0:
Would this meet your requirements?

image

phl0:
This is the change I did: https://github.com/phl0/wavelog/commit/cd83aa30bc4c4ac918f0bf59a86a276fb3ab27b0

violarulan:
> This is the change I did: [phl0@cd83aa3](https://github.com/phl0/wavelog/commit/cd83aa30bc4c4ac918f0bf59a86a276fb3ab27b0)

You are right. It should have constant spacing on this. 

But - In zh_cn locale, the browser maps the `monospace` to system's default font [NSimSun](https://en.m.wikipedia.org/wiki/File:NSimSun_font_example.svg) , which looks a bit odd.

Snipaste_2024-11-04_08-57-29

To my knowledge, `Consolas` is widely used on Windows and it's not available on MacOS.

If we change the `font-family` to `Consolas, monospace`, it will work on both platform.

HB9HIL:
image
This is how it looks like on Mac (Chrome)

violarulan:
> This is how it looks like on Mac (Chrome)

Hmm, it looks rendered as `Consolas`. I'll close it for now and see if there's a better workaround.

PR #1156, opened 2024-11-03
library for trx-control

HB9HIL:
Even we don't have any features yet which do support the communication from Wavelog to a [trx-control](https://github.com/hb9ssb/trx-control) server it's nice to have this possibility. 

For future use this brings a lib to Wavelog which allows to communicate with a trxd daemon.

Usage:

```php
$this->load->library('trxd');
$this->trxd->request($command, $to, $parameters);

// for example
// Turns on the power plug supplying the shack ;-)
$this->trxd->request("power", "tasmota", array("state" => "on"));

```

### DIFF
--- a/application/config/config.sample.php
+++ b/application/config/config.sample.php
@@ -719,3 +719,31 @@ $config['cron_allow_insecure'] = false;
  */
 
 $config['disable_version_check'] = false;
+
+/*
+|--------------------------------------------------------------------------
+| trx-control Configuration
+|--------------------------------------------------------------------------
+|
+| This defines server and port of your personal trx-control server.
+| If you don't have a trx-control server, you can ignore this.
+|
+| trxd_server_ip            IP of your trx-control server
+| trxd_server_port          Port of your trx-control server
+| trxd_connection_type      Connection type of your trx-control server (ws, wss or plain)
+|                           ws:     normal websocket
+|                           wss:    secure websocket (requires a valid certificate on trx-control server)
+|                           plain:  plain tcp/ip socket connection
+| trxd_ws_path              Path of your trxd websocket server (only required for ws and wss)
+| trxd_server_timeout       Timeout before the connection to trx-control server is closed
+|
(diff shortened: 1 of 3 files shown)


gx1400:
Comment from our conversation on trxd Element: This implementation has wavelog connect directly to the trxd daemon, which is good if locally hosted or if wavelog as a direct connection or vpn tunnel to trxd for example.

My use-case: trxd would be on my home network with wavelog hosted on a cloud-based VPS.  trxd can connect to wavelog api but wavelog cannot connect directly to trxd.

Suggestion (another project): a middle-ware websocket gateway/relay that could run in a docker container along side wavelog.  

trxd <-> relay <-> wavelog

Middleware could announce connections, disconnections, etc events. The wavelog trx-control library could theoretically query if it  has a direct connection or a relayed connection.

HB9HIL:
> Comment from our conversation on trxd Element: This implementation has wavelog connect directly to the trxd daemon, which is good if locally hosted or if wavelog as a direct connection or vpn tunnel to trxd for example.
> 
> My use-case: trxd would be on my home network with wavelog hosted on a cloud-based VPS. trxd can connect to wavelog api but wavelog cannot connect directly to trxd.
> 
> Suggestion (another project): a middle-ware websocket gateway/relay that could run in a docker container along side wavelog.
> 
> trxd <-> relay <-> wavelog
> 
> Middleware could announce connections, disconnections, etc events. The wavelog trx-control library could theoretically query if it has a direct connection or a relayed connection.

This is true. If your Wavelog Instance is on a VPS you may have difficulties to connect to a local hosted trxd.

A relay server or something like that is beyond the scope for this at the moment. Another solution would be to make trx-control public available and add some kind of authentication to it. But aswell out of scope for this PR :)

AndreasK79:
I have no way to test, but as it's disabled by default, it should be ok.

AndreasK79 (approved):

PR #1158, opened 2024-11-04
Staticmap - Options for Contest and Date Range

HB9HIL:
Added these options to staticmap:

`contest` 
Ability to filter for a contest. Use Contest ID (Adif Name)

`start_date`
Shows only QSOs from a certain date

`end_date`
Shows only QSOs to a certain date

`qsocount=all`
Shows all available QSOs in Database


phl0 (changes requested):
This will not work (with default config) as "-" is not a valid URI character:

image

Specified start_date as "2024-11-04" in this case.

phl0 (approved):
My previous comment was sh*t. Forget about it :)
Works as expected.

PR #1159, opened 2024-11-04
[Advanced Logbook] Added distance column

AndreasK79:


phl0 (approved):
Fine so far =)

phl0 (approved):
Very nice. I guess I can make use of that sorting code snippet in other places =) Thank you

Issue #987, opened 2024-09-21
QSO-field for confirmed long path contact?


Beiri22:
Hi,

I just wonder if there could be a check box (create qso, edit qso) for usage of long path -- which would result in a different distance calculation and make a long path drawn on the globe map...

HB9HIL:
It's totally legit to ask for that as `ANT_PATH` is an official ADIF field. We currently have no possibility to edit that. I plan to do some frontend stuff for the QSO edit field. I will address this aswell then, okay?

I currently have a lot of private stuff to do (Moving Home from A to B). After that I will take care of that. Please be patient.

Beiri22:
i will be

HB9HIL:
@Beiri22 

First step is done. You now can edit ANT_PATH in the QSO edit. The path drawings and distance calculation will be addressed in another PR.

HB9HIL:
Distance calculation is also included. The path drawings will be adressed some other day as they are really complex. The main request is solved so I guess we can close this issue.

PR #1162, opened 2024-11-04
QSO Edit Modal

HB9HIL:
Did a little refurbish of the QSO edit modal. The HTML was inconsistent and it looked a bit odd. 

In addition I added the ability to edit the Antenna Path which was requested in #987. We have this field in database, it gets imported from adif, so why not allow editing it.

### DIFF
--- a/application/models/Logbook_model.php
+++ b/application/models/Logbook_model.php
@@ -1427,6 +1427,7 @@ class Logbook_model extends CI_Model {
 			'COL_DARC_DOK' => strtoupper($this->input->post('darc_dok')),
 			'COL_QTH' => $this->input->post('qth'),
 			'COL_PROP_MODE' => $this->input->post('prop_mode'),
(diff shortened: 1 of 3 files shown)


phl0 (approved):
Works fine. Maybe add ANT_PATH to the displayQso or logbook/view popup to show that info if stored?

HB9HIL:
> Works fine. Maybe add ANT_PATH to the displayQso or logbook/view popup to show that info if stored?

Good Point!!

HB9HIL:
> Works fine. Maybe add ANT_PATH to the displayQso or logbook/view popup to show that info if stored?

I would do this in a second PR as this brings some other points up. (Distance Calculation and so on). Fine with that?
PS: Prop Mode is missing aswell :D

phl0:
> > Fine with that?

Jupp

PR #1163, opened 2024-11-05
Config passwords

HB9HIL:
Moving the callbook passwords from config.php to the database (encrypted)


HB9HIL:
This will need another approach

PR #1165, opened 2024-11-05
Antenna Path in QSO Logging and QSO Details

HB9HIL:
Since we have Antenna Path in the Database, supporting it also in ADIF, we should calculate distances based that. 

This PR includes the Distance Calculation in the QSO Details View and the QSO Logging itself. For this it also includes the basic necessary changes in the distance calculation.

Check with whitespaces hidden.

Maps will be addresses some other day since this will be a bigger change.


HB9HIL:
@AndreasK79 I tried to implement the longpath pathlines in the staticmap.. Quiet complex so let's address maps (staticmap and lba) in another PR on another day

phl0 (changes requested):
Code works fine on my end show shows distance for LP correct upon logging and in the database. The QSO dislpay however shows the SP value:

image

IMHO this should be corrected.

application/views/view_log/qso.php
```diff
@@ -145,6 +149,26 @@
                                         $distance .= " nmi";
                                         break;
                                 }
+
+                                if ($ant_path != null) {
+                                    log_message('error', 'Ant Path: '.$ant_path);
```
phl0, line 154: Debug message to be removed?!

phl0 (changes requested):

phl0 (changes requested):
Changing the ANT_PATH in QSO edit does not trigger a re-calculation of distance. Logged a LP QSO with HB9HIL and changed ANT_PATH to SP in QSO edit results in:

image

HB9HIL:
> Changing the ANT_PATH in QSO edit does not trigger a re-calculation of distance. Logged a LP QSO with HB9HIL and changed ANT_PATH to SP in QSO edit results in:

should be fixed. Please test agn

HB9HIL:
QRX. There is a bad interfering when calling the edit ajax in QSO view due to identical ID's

phl0 (approved):
Edit QSO works fine now and switches SP/LP as desired. Good to merge imho.

PR #1167, opened 2024-11-06
Skip AMSAT status upload for invalid/non-existent grids

phl0:
If we use a station profile with MY_VUCC_GRIDS instead of a single grid in MY_GRIDSQUARE the Logbook model fails because the amsat status URL cannot be configured due to the gridsquare being undefined. In due course the saving of the QSO hangs because the underlying request fails:

image

We need to check both variables for a valid grid. In case MY_VUCC_GRIDS is set we use the first grid of the list because amsat status can only handle a single grid. In addition if we cannot determine a grid the amsat status upload is skipped to prevent the script from failing and the request to store the QSO pending.

### DIFF
--- a/application/models/Logbook_model.php
+++ b/application/models/Logbook_model.php
@@ -1178,12 +1178,20 @@ class Logbook_model extends CI_Model {
 		} else {
 			$sat_name = $data['COL_SAT_NAME'];
 		}
-		$datearray = date_parse_from_format("Y-m-d H:i:s", $data['COL_TIME_ON']);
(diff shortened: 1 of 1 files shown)


application/models/Logbook_model.php
```diff
@@ -1178,12 +1178,20 @@ function upload_amsat_status($data) {
...
+		if (array_key_exists('COL_MY_GRIDSQUARE', $data)) {
+			$amsat_source_grid = $data['COL_MY_GRIDSQUARE'];
+		} else if (array_key_exists('COL_MY_VUCC_GRIDS', $data)) {
+			$amsat_source_grid = strtok($data['COL_MY_VUCC_GRIDS'], ',');
+		}
+		if ($amsat_source_grid != '') {
+			$datearray = date_parse_from_format("Y-m-d H:i:s", $data['COL_TIME_ON']);
+			$url = 'https://amsat.org/status/submit.php?SatSubmit=yes&Confirm=yes&SatName=' . $sat_name . '&SatYear=' . $datearray['year'] . '&SatMonth=' . str_pad($datearray['month'], 2, '0', STR_PAD_LEFT) . '&SatDay=' . str_pad($datearray['day'], 2, '0', STR_PAD_LEFT) . '&SatHour=' . str_pad($datearray['hour'], 2, '0', STR_PAD_LEFT) . '&SatPeriod=' . (intdiv(($datearray['minute'] - 1), 15)) . '&SatCall=' . $data['COL_STATION_CALLSIGN'] . '&SatReport=Heard&SatGridSquare=' . $amsat_source_grid;
+			$ch = curl_init();
+			curl_setopt($ch, CURLOPT_URL, $url);
```
int2001, line 1191: i'd suggest to add a timeout here.

phl0: Good point. Added.

int2001 (approved):
Yeah. Great

PR #1168, opened 2024-11-06
Adjust regex to accept VUCC grids (again)

phl0:
In https://github.com/wavelog/wavelog/pull/1134 we changed the code to accept up to 10-char maidenhead grids but we crossed that station gridsquare can also be a VUCC gridsquare or gridline. So LBA would only map QSOs with a single station gridsquare or two or four 10-char grids (which is not what we want). So we need to make the 3rd to 5th pair of chars/numbers optional.


phl0:
Tested with following grids as station gridsquare:

- XX00
- XX00XX
- XX00XX00
- XX00XX00XX
- XX00,XX00
- XX00XX,XX00XX
- XX00,XX00,XX00,XX00

AndreasK79 (approved):

AndreasK79:
@phl0 oops. Never meant to do that. Guess too late in the evening, and too quick to fix :)

PR #1169, opened 2024-11-06
Autocomplete and Pretty zeros

HB9HIL:
Disabled the browsers native autocomplete feature in the QSO and Contest Callsign input. This is pretty annoying and can result in wrong inputs. 

Aswell I found some callsign zeros which can be pretty Ø (for you @phl0  😄 )


phl0 (approved):
I like 🥰

PR #1175, opened 2024-11-08
Show DXCC as badge with status in the search results

HB9HIL:
image


int2001 (changes requested):
Searchbar is working, but at add QSO it throws an error:

image

phl0:
> Searchbar is working, but at add QSO it throws an error:

And here is the fix: https://github.com/HB9HIL/wavelog/pull/116.

phl0:
Other than that a nice idea. Tested ok with https://github.com/HB9HIL/wavelog/pull/116.

HB9HIL:
tnx for the fix @phl0 

missed the method above

int2001 (approved):
works

PR #1177, opened 2024-11-08
php-gd in Docker

HB9HIL:
for the Static Map #1098 we need php-gd also in docker


int2001 (approved):
tbh: i don't like the new dependency for the "mlocati"-extension.
GitHub builds as well as local.

but if it helps others to build it ....

PR #1178, opened 2024-11-08
Individual CAT URL

int2001:
Added the ability to set a different CAT-Callback-URL for the dxcluster for each radio.

telegram-cloud-photo-size-2-5292044692957032404-y

Old behaviour: Click on a Call at Bandmap-list, and the browser requested a QSY via http://127.0.0.1:54321/QRG to WaveLogGate

New behaviour: 
- One can set individual URLs for each radio. So it is - e.g. - possible to run WavelogGate (or other tools) on different machines than the Browser. 
- A click on a Call at Bandmap-list, and the browser requested a QSY via the configured URL (at HW-Interfaces) to WaveLogGate.
- The default-URL is set to http://127.0.0.1:54321 to ensure all standard-installations are still running

### DIFF
--- a/application/config/migration.php
+++ b/application/config/migration.php
@@ -22,7 +22,7 @@ $config['migration_enabled'] = TRUE;
 |
 */
 
-$config['migration_version'] = 225;
+$config['migration_version'] = 226;
 
 /*
 |--------------------------------------------------------------------------
--- a/application/controllers/Radio.php
+++ b/application/controllers/Radio.php
@@ -116,10 +116,10 @@ class Radio extends CI_Controller {
 	}
 
 	public function saveCatUrl() {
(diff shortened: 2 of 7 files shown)


HB9HIL (approved):

phl0 (commented):
This breaks compatibility with MySQL Servers: 

image

We could either remove the default value or change to VARCHAR(255).

gen-intel:
FYI I just got this Individual CAT URL working with node-red and a FlexRadio. 
One thing is almost certain, you need to use HTTPS. 
Most browsers (as configured today) did not like fetching with HTTP to anything except 127.0.0.1.
Hope to see Individual CAT URL in the master soon.

int2001:
yep. that's indeed a browser-problem. CORS is very complicated here.

PR #1182, opened 2024-11-10
Static Map - Some improvements

HB9HIL:
Some smaller bugfixes and improvements. 

- fixed a special case in wrapping (if the path/marker crosses the 180° longitude)
- need much less memory yet
- fixed a bug in URL options


int2001 (approved):
tested only for performance and function: Works much faster. Good improvement

PR #1187, opened 2024-11-11
Cache clearing for Static Maps - Bugfix

HB9HIL:
fixed the multiple cache clearing and found a bug in simplefle which I fixed directly aswell

the cache clearing is now called in `import_bulk()` instead the `import()` function (now only once)


phl0 (approved):
Cache clearing working fine on my end.

PR #1190, opened 2024-11-12
[QSO] Bug in Start Timer

HB9HIL:
Fixed a bug in the start timer. 

if you clicked "Reset Start Time" during a QSO, the start timer was not reset to the current time. This PR fixes that.


int2001 (approved):
yeah. nice finding

PR #1193, opened 2024-11-12
Omit various fields for dynamic callsigns (/P, /M, /MM, ...)

HB9HIL:
In #543  we added the omitation, that values for QTH, Locator and so on aren't returned from callbook if a dynamic callsign is logged 

This worked well for QRZ.com and HamQTH but was never implemented for returns of the database. This PR fixes that.

### DIFF
--- a/application/controllers/Logbook.php
+++ b/application/controllers/Logbook.php
@@ -165,7 +165,7 @@ class Logbook extends CI_Controller {
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


int2001 (approved):
tested. works.

- takes cached when fixed and worked b4
- takes qrz/hamqth when fixed and not worked b4
- takes only name when portable/mobile

PR #1195, opened 2024-11-12
QSO - More efficient DXCC Summary

HB9HIL:
The DXCC summary now only shows worked/confirmed modes and shows a red frame around the current mode/band


int2001 (approved):
like it. works

PR #1196, opened 2024-11-12
Enhanced QRG Input

HB9HIL:
# Enhanced QRG Input

Using always **Hertz** as QRG unit is not just annoying, but also not really user friendly. Therefore we should allow to use more user-friendly units to set/display the frequency. A first step was done in #544 and #784 but the QRG input itself was missing. This now changed. Base issue is #165 

image

## Usage

There are now several ways to enter a frequency. A user is free to choose which one is prefered.

### Same QRG Unit
If the QRG unit is `kHz` and the user wants to use `kHz` he can simply type in the frequency and hit enter
ezgif-7-700f45041d

#### Change by input
If a user want to change the QRG quickly and wants to use a different QRG unit he can simply type that unit after the frequency. Wavelog will recognize that and change the unit accordingly.
ezgif-7-48dcc44282

#### Change by click
If the user rather want to use the mouse to change the QRG unit he can simply click on the unit in the frequency input
ezgif-7-144f6b4c57

## Technical

The QRG units are already stored in the session data (server side). The new code now stores the QRG units additionally in the localStorage (browser side, only in QSO and Contesting view) to avoid slow ajax calls and make the whole logic faster. The localStorage gets cleared when reloading the page. If a user want to use a different QRG Unit then already set then he can change the unit per band in the 'Bands' menu (User -> Bands). 

## Progress

- [x] Show QRG unit based on user options in the QSO logging
- [x] Make the QRG input dynamic
- [x] Add ability to change unit by clicking on the button
- [x] Use localStorage to avoid ajax calls (to get the user options)
- [x] Port everything to contesting

## Note about SimpleFLE

SimpelFLE is based on FLE by DF3CB which always uses **MHz**. To stay close to the original FLE syntax we use MHz here aswell.


int2001 (approved):
works.
tested:
- LiveQSO with CAT
- LiveQSO without CAT
- PostQSO
- Contesting with CAT
- Contesting without CAT
- DXCluster / Spot-Click

## Important:
since this one affects the core of logging, id appreciate another approve besides mine. Maybe i've overseen something

int2001:
Checked again. Now SAT is working too

AndreasK79 (approved):

AndreasK79:
Nice addition. Tested and clicked around. Logged a few QSOs. Seems ok to me.

PR #1198, opened 2024-11-12
Cabrillo Fix

HB9HIL:
There was an bug in the modes of cabrillo exports.

Additionally we had no possibility to include the gridsquares in the cabrillo file. Since Wavelog does not know that the grid was chosen as exchange type (we do not store this yet...), I had to add a little option in the export form. (Default: No)

image

 This affects for example the DARC FT4 contest.
 
 The sequence is 
 
 `QSO: [QRG in kHz]  [MODE*]   [DATE]   [TIME]  [MY_CALL]  [RSTs]  [SERIAL]  [GRID]  [EXCH]  [RSTr]  [SERIAL]  [GRID]  [EXCH] `
 
 Serial, Grid and Exchange: Each if not null / empty  and grid if option is true
 
 *Modes based on the official cabrillo documentation: https://wwrof.org/cabrillo/cabrillo-qso-data/


HB9HIL:
@dg9vh confirmed that this is the correct cabrillo syntax

int2001 (approved):
works.

PR #1204, opened 2024-11-13
Order same time QSOs by PK desc

phl0:
For QSOs added with same time (e.g. SAT passes) there is a difference in sort order for those QSOs. Dashboard shows the correct order, i.e. same QSO sorted by PK desc:

Screenshot 2024-11-13 172723

The logbook overview however reverses these QSOs

Screenshot 2024-11-13 172740

We should add a secondary sorting criteria to take PK into account also (supposed that SAT QSOs are logged in a timely order, i.e. increasing PK).


phl0:
Seems it is handled differently on different databases. Need to investigate a little more.

PR #1205, opened 2024-11-14
Use PK as secondary sorting criteria after time

phl0:
For post-logging satellite QSOs (transcribing recordings) it often happens that users use the same timestamp for all QSOs of a SAT pass (LoTW matches QSOs on a +- 15 min basis). The current state uses different sorting criteria for listing QSOs.

Dashboard:

dashboard

Logbook:

logbook

Logbook Advanced:

lba

The LBA order is correct. It uses time as first and primary key as second sorting criteria. This should be used in the other places as well. Fixes: Logbook view, Dashboard view and QSO lists popup (for example Activated Gridsquares).


int2001 (approved):
helps and works :)

PR #1217, opened 2024-11-18
[LBA] Fix edit query

phl0:
In https://github.com/wavelog/wavelog/pull/1121 we added a distinction for specific DXCCs (i.e. NONE). This generates an error for LBA edit function as "adif" is not part of the query. It results in a PHP error on the response:

image

Fixed by adding dxcc_entities.adif to the SQL query.


int2001 (approved):
yeah. and i wondered about the error-message. good finding

AndreasK79 (approved):

phl0:
Tnx fix -> merge time!

PR #1219, opened 2024-11-18
Fix overridelogic at ADIF-Upload

int2001:
Logic at override (ADIF-Upload / mark as uld) was wrong for LoTW and Clublog


HB9HIL (approved):
not tested but code looks good to me

phl0 (approved):
Makes sense to me and works during a simple test importing an ADIF file.

PR #1220, opened 2024-11-18
[Advanced Logbook] Added empty search for textboxes

AndreasK79:
Added * for wildcard search instead of empty (search works like before).
If * is removed, an empty search is performed.


HB9HIL (approved):
Really nice and convenient :) I Like!

phl0 (approved):
Nice idea. Works on my end.

PR #1221, opened 2024-11-18
Staticmap Watermark

HB9HIL:
Some users want to remove the watermark. Okay for me. There is still the text which says "Created by Wavelog"


int2001 (approved):
as long as we don't lose the advertising-effect...
imho this is indeed discussable (from the branding-pov). but, ok for me.

PR #1225, opened 2024-11-19
Added "yellow" for clublog-modification state at UI

int2001:
The Modified state for Clublog-QSL was missing.
Added it at:
- Overview
- LBA
- Search


int2001:
[WIP] because of wished regarding the code-style

HB9HIL (changes requested):
two small adjustments necessary. 

besides that we have an inconsistency. If the Clublog Status is `M` but the Upload date is empty it shows a yellow arrow in LBA but a red arrow in the Logbook Overview. I would probably remove the dependency that the date can not be empty and show just "Modified" if the date is empty

image
image
*same QSO, first in LBA, second in Logbook Overview

HB9HIL (approved):

PR #1229, opened 2024-11-19
[Advanced Logbook] Added icons on buttons

AndreasK79:
image


int2001 (approved):
needs a minimum browser-width imho - otherwise the text will be wrapped.
but looks nice and works.

fine with it.

phl0 (approved):

HB9HIL (approved):
works

PR #1235, opened 2024-11-20
[Advanced Logbook] Fix for dxcc name when updating from Callbook

AndreasK79:
Bug reported by @HB9HIL. DXCC was set to NONE in the table after callbook update in LBA


HB9HIL (approved):
works! thx for this fix

PR #1239, opened 2024-11-21
Enable at least the manual uploading of a file for LoTW

int2001:
Disabling the LoTW-Userimport (at config.php) can cause a problem.

E.g.:
- A User uploads/logs QSOs (via ADIF/logging) for today
- Cron syncs with LoTW
- QSOs will (perhaps) be confirmed
- Last-QSL-Date has the value of today.

Now User likes Wavelog and uploads his history.
The LoTW-QSLs will never be confirmed via cron, because they're all older than today.

This one re-enables the ability to manual-upload the LoTW-Report to Wavelog. So there's a chance to "heal" those QSOs.

@Reviewers: Check with "ignore whitespace" - it isn't that much

### DIFF
--- a/application/controllers/Lotw.php
+++ b/application/controllers/Lotw.php
@@ -99,8 +99,7 @@ class Lotw extends CI_Controller {
 	| and processing of p12 files and storing the data into mysql
 	|
 	*/
-	public function do_cert_upload()
-    {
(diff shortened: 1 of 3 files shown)


HB9HIL (approved):
like this one

PR #1247, opened 2024-11-21
Long Path Fixes

HB9HIL:
There are some more places where distances are calculated

Test:
- LBA Mapping -> Distance is shown in the little map modals
- LoTW Update (via File) -> maybe @phl0 can check that
- Logbook Controller methods `distance()` (not in use?) and `bearing()` (is in use but does not need any further adjustment
- Regtest1 Format
- QSO View VUCC Multis Distance

### DIFF
--- a/application/controllers/Logbook.php
+++ b/application/controllers/Logbook.php
@@ -1259,7 +1259,7 @@ class Logbook extends CI_Controller {
 	}
 
 	/* return station bearing */
-	function bearing($locator, $unit = 'M', $station_id = null) {
+	function bearing($locator, $unit = 'M', $station_id = null, $ant_path = null) {
 		if(!$this->load->is_loaded('Qra')) {
 			$this->load->library('Qra');
 		}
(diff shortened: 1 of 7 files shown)


phl0:
LoTW does not handle/care about distances. So we will not retrieve that info from LoTW files.

phl0:
OK misunderstood. It is just re-calculating on LoTW import. Will test.

int2001 (approved):
tested:
- Logging
- Editing
- Update via "Show distances"
- Update via "Admin-Menu"

works.

not able to test it in reality, because i don't fill in SP or LP. But from the Test-POV it looks okay to me.

phl0:
> @phl0 did you test

Planned for tonight. :)

phl0:
Fix in https://github.com/HB9HIL/wavelog/pull/117. You cannot use $data['COL_ANT_PATH'] in line 3627/3630 as $data was unset before. So it would be null in any case leading to distance always being calculated as SP.
Instead we have to pull it from the QSO we previously looked up to confirm.

phl0 (approved):
LoTW Import now works correctly.

PR #1248, opened 2024-11-21
Fetch email from callbook / last QSOs

phl0:
Now we have the field for e-mail we could fill it with QRZ/HamQTH data on lookup...


int2001 (approved):
works

PR #1250, opened 2024-11-22
[QSOs] Added QSOs per month

AndreasK79:
Continuing with QSO stats. Now we have a months of the year tab:
image


phl0 (approved):
Works:

image

PR #1254, opened 2024-11-23
Update map when entering POTA/SOTA/WWFF

HB9HIL:
When entering a SOTA, POTA, WWFF in the QSO logging it updates the Gridsquare but not the map.

This fixes it with triggering the input event (Listener in qso.js is working `.on('input focus')..`)


int2001 (approved):
works.
i like those small changes / improvements!

PR #1257, opened 2024-11-23
Wavelog Dataservice Implementation

HB9HIL:
This PR brings the Integration for the Wavelog Dataservice. This provides a possibility to delivery big data, which is to big and cumbersome to store directly in Wavelog. It's fully scaleable and does not need much attention.

This includes:
- Global Option to change the Dataservice URL
- Global Option to disable the Dataservice completely

- First feature which uses the dataservice -> Shows the local time at the QSO Partners location (grid based)
This feature required a mapping library to calculate the timezone based on coordinates. This lib is that big, that it's not smart to store this directly in Wavelog. This gave the idea to start this little side-project. The dataservice is already running beta and can be tested. -> https://data.wavelog.org/


HB9HIL:
Based on different feedbacks this will be closed for the moment. We will find another way to implement the Local Time Feature.

PR #1259, opened 2024-11-24
Add UTC/Local display switch

pomtom44:
Hopefully I have done this right as this is my first proper PR on github.
And I have tried a few different ways to get it to push properly.

This is a feature update to allow for switching between local and UTC time on logs.
The data is stored in UTC and only updated when rendered to the screen.
This will allow for third party services to keep using the UTC time.
Timezones are handled by PHP and 'should' take into account daylight savings.

I had to re-do the timezone database, so I hope that is ok.
I couldn't see any conflicts with it, and there wasn't any errors during my testing.


phl0:
What is the overarching goal here? You want Wavelog to display QSO data in local time rather than UTC? Imho this introduces a layer of unneeded complexity. According to current best practices all logging is done in UTC. Why would you display time in local time? You would have to correct times for filling QSL cards for example. You would have to deal with daylight saving ... etc.
TBH: I do not see the need to implement this. Would subvert all the reasons why we use UTC.

application/models/User_model.php
```diff
@@ -468,6 +470,8 @@ function update_session($id, $u = null, $impersonate = false) {
 			'user_default_band'	 => $u->row()->user_default_band,
 			'user_default_confirmation'	 => $u->row()->user_default_confirmation,
 			'user_qso_end_times' => isset($u->row()->user_qso_end_times) ? $u->row()->user_qso_end_times : 1,
+			'user_timezone' => $this->getTimezone($u->row()->user_timezone),
+			'user_show_log_utc' => isset($u->row()->user_show_log_utc) ? $u->row()->user_show_log_utc : 1,
```
HB9HIL, line 474: You stored the option `user_show_log_utc` in the users table directly. This is not necessary as there is a table called `user_options` available which can store dynamic options like this.

application/models/User_model.php
```diff
@@ -609,12 +613,20 @@ function users() {
...
 		return $ts;
+	// FUNCTION: string getTimezone()
+	// Returns the timezone name from the ID
+	function getTimezone($id) {
+		$query = $this->db->query('SELECT name FROM timezones WHERE id = ?', array($id));
+		if ($query->num_rows() > 0) {
+			return $query->row()->name;
+		}
+		return null;
 	}
```
HB9HIL, line 630: You removed a `}` by accident which causes the whole User_model to fail. 

In addition you removed the `ORDER BY offset` which causes the timezone tables in the UI to be unsorted. There is no reason to remove that.

application/views/dashboard/index.php
```diff
@@ -44,6 +44,17 @@ function echo_table_col($row, $name) {
...
+//Not sure if it can be moved to a central file to only be used once
+
+//Converts UTC time to local time using PHP's timezones.
+//This accounts for Daylight Savings (apparently)
+function UTCtoLocal($timestamp, $timezone){
+	$utc_date = (new DateTime('@' . $timestamp))->setTimezone(new DateTimeZone('UTC'));
+	$local_date = $utc_date;
+	$local_date->setTimeZone(new DateTimeZone($timezone));
+	return $local_date;
+}
```
HB9HIL, line 57: You are right. All these function you placed in the views shouldn't be there. You could place a function like this in the Library GenFunctions and call it in the Controller to pass the already calculated time to the view. 

Views always should only display data. I know we have some functions and PHP code in the views aswell but some are leftovers and need a refactor and others are necessary. Whenever possible we should stick to the MVC Pattern

Models - Database Actions and "I do/calculate something"
Views - Display the data
Controllers - Decide which data to gets display. 

-> Librarys are used for specific things and could be used here

application/views/dashboard/index.php
```diff
@@ -167,10 +178,10 @@ function echo_table_col($row, $name) {
 
     		<thead>
 				<tr class="titles">
-					<th><?= __("Date"); ?></th>
-
+					<!-- Display UTC or Local in header based on settings -->
+					<th><?= __("Date"); ?><?php if($this->session->userdata('user_show_log_utc') == 1){ echo __(" (UTC)"); }else{ echo __(" (Local)"); } ?></th>
```
HB9HIL, line 182: Now there are coming a lot of views which display the new time.

This could be omitted. Yeah it's nice to display that but I'd rather generate this in the Controller to avoid redudant code

application/views/dashboard/index.php
```diff
@@ -200,10 +211,21 @@ function echo_table_col($row, $name) {
...
+						<?php if(($this->config->item('use_auth') && ($this->session->userdata('user_type') >= 2)) || $this->config->item('use_auth') === FALSE || ($this->config->item('show_time'))) { ?>
+							<td><span data-bs-toggle="tooltip" data-bs-placement="right" data-bs-custom-class="custom-tooltip" data-bs-title="<?php $timestamp = strtotime($row->COL_TIME_ON ?? '1970-01-01 00:00:00'); echo UTCtoLocal($timestamp, $this->session->userdata('user_timezone'))->format('H:i'); ?> (Local)"><?php $timestamp = strtotime($row->COL_TIME_ON ?? '1970-01-01 00:00:00'); echo date('H:i', $timestamp); ?></span></td>
+						<?php } ?>
+
+					<?php }else{ ?>
+					
+						<td><span data-bs-toggle="tooltip" data-bs-placement="right" data-bs-custom-class="custom-tooltip" data-bs-title="<?php $timestamp = strtotime($row->COL_TIME_ON ?? '1970-01-01 00:00:00'); echo date($custom_date_format, $timestamp); ?> (UTC)"><?php $timestamp = strtotime($row->COL_TIME_ON ?? '1970-01-01 00:00:00'); echo UTCtoLocal($timestamp, $this->session->userdata('user_timezone'))->format($custom_date_format); ?></span></td>
```
HB9HIL, line 225: This is a lot of code to just display a time..

application/views/logbookadvanced/index.php
```diff
@@ -20,9 +33,17 @@
...
+			$showUTC = "true";
+		}else{
+			$showUTC = "false";
+		}
+		$timezone = $this->session->userdata('user_timezone');
         $options = "{
             \"datetime\":{\"show\":\"true\"},
-            \"de\":{\"show\":\"true\"},
+			\"showutc\":{\"show\":\"$showUTC\"},
+			\"timezone\":{\"value\":\"$timezone\"},            \"de\":{\"show\":\"true\"},
```
HB9HIL, line 46: I don't know what you are doing here exactly but this does not look right to me.

assets/js/sections/logbookadvanced.js
```diff
@@ -27,6 +27,25 @@ $('#selectPropagation').change(function () {
...
+    const formattedDate = [
+        String(localDate.getDate()).padStart(2, '0'),  // Day
+        String(localDate.getMonth() + 1).padStart(2, '0'),  // Month
+        String(localDate.getFullYear()).slice(-2),  // Year
+    ].join('/') + ' ' + [
+        String(localDate.getHours()).padStart(2, '0'),  // Hours
+        String(localDate.getMinutes()).padStart(2, '0'),  // Minutes
+    ].join(':');
+    return formattedDate;
+}
```
HB9HIL, line 48: You decided to write a second function which converts UTCtoLocal in Javascript. 
In this case this is not necessary as the data is coming from the controller. Why not converting in the controller where the PHP function is available.

install/assets/install.sql
```diff
@@ -366,92 +366,431 @@ INSERT INTO `options` (`option_id`, `option_name`, `option_value`, `autoload`) V
...
-INSERT INTO `timezones` VALUES ('136', '9.0', '(GMT+09:00)-Osaka, Sapporo, Tokyo');
-INSERT INTO `timezones` VALUES ('139', '9.5', '(GMT+09:30)-Adelaide');
-INSERT INTO `timezones` VALUES ('141', '10.0', '(GMT+10:00)-Brisbane');
-INSERT INTO `timezones` VALUES ('142', '10.0', '(GMT+10:00)-Canberra, Melbourne, Sydney');
-INSERT INTO `timezones` VALUES ('144', '10.0', '(GMT+10:00)-Hobart');
-INSERT INTO `timezones` VALUES ('146', '11.0', '(GMT+11:00)-Magadan, Solomon Is., New Caledonia');
-INSERT INTO `timezones` VALUES ('147', '12.0', '(GMT+12:00)-Auckland, Wellington');
-INSERT INTO `timezones` VALUES ('149', '13.0', '(GMT+13:00)-Nuku\'alofa ');
-INSERT INTO `timezones` VALUES ('150', '-4.5', '(GMT-04:30)-Caracas');
+INSERT INTO `timezones` (`id`, `name`) VALUES
```
HB9HIL, line 376: One of the biggest problems of this PR.

You changed the install.sql in the installer. This is not how database changes work. It's the same with your additional column of the utc option in the users table. 

You did the change to your database. So it works for you. But this would require that all users which use Wavelog would need to change their database manually. Otherwise their installation would be broken. 

What you would need to do is a migration which modifies all databased. Changing the install.sql is always a bad idea and should be really good considered..

HB9HIL (commented):
Dear @pomtom44 

First let me say thank you for putting so much effort in Wavelog and contribute to this project. We really appreaciate your engagement for this but I fear this PR is in a state, where it's far far away from mergable. There are a lot of things which are not solved correctly and will break all Wavelog Installations out there if we merge it like this. 

I will now explain some of the issues

HB9HIL:
@pomtom44 

Please contact me once directly: hb9hil@wavelog.org

pomtom44:
> What is the overarching goal here? You want Wavelog to display QSO data in local time rather than UTC? Imho this introduces a layer of unneeded complexity. According to current best practices all logging is done in UTC. Why would you display time in local time? You would have to correct times for filling QSL cards for example. You would have to deal with daylight saving ... etc. TBH: I do not see the need to implement this. Would subvert all the reasons why we use UTC.

The goal is since my UTC time is +12 or +13 depending on daylight savings, I keep local time logs as well for my own records.
I still keep UTC for QSL, third party services, etc.
It is just nice to know the local time a log was made without having to do UTC conversions.

pomtom44:
> Dear @pomtom44
> 
> First let me say thank you for putting so much effort in Wavelog and contribute to this project. We really appreaciate your engagement for this but I fear this PR is in a state, where it's far far away from mergable. There are a lot of things which are not solved correctly and will break all Wavelog Installations out there if we merge it like this.
> 
> I will now explain some of the issues

Thanks for the deep dive into the issues.
I was expecting there to be quite a few as im not a proper programmer, I make very basic personal projects and this is the first time I have contributed to a larger project.

I wasn't expecting this to be merged, it as more of a PoC to show what I was thinking and my reasoning behind it.

I have sent a email as you have requested, and I am happy for you to close this PR off as you see fit if there is no plans to implement something similar.

PR #1260, opened 2024-11-25
Callbook refactor

AndreasK79:
Started on a refactor of how we do the callbook lookup.

The reason is simply that is should be much easier to add other callbooks, and maintain the code.

QRZCQ is tested working.
QRZ and HamQTH still works.


int2001 (approved):
Really great improvement. much cleaner code than before.
i tested a lot of calls and found no issue.

tbh: one cannot test every call-constellation, but i tried to test as much as possible.

i like

phl0 (approved):
Tested around 30 callsigns which all parsed correctly. This looks fine to me. :)

PR #1265, opened 2024-11-26
Display a numeric keyboard for number input fields

phl0:
This changes the numeric input fields so that on mobile / touch input platforms a simplified numeric keyboard is shown instead of a full blown text input one.

Before:

20241126_215629000_iOS

After:

20241126_215557000_iOS

### DIFF
--- a/application/views/qso/index.php
+++ b/application/views/qso/index.php
@@ -171,10 +171,10 @@
                 <div class="mb-3 col">
                   <label for="frequency"><?= __("Frequency"); ?></label>
                   <div class="input-group input-group-sm">
-                    <input type="number" inputmode="decimal" tabindex="3" class="form-control form-control-sm" id="freq_calculated" name="freq_calculated" value="0" />
+                    <input type="text" tabindex="3" class="form-control form-control-sm" id="freq_calculated" name="freq_calculated" value="0" />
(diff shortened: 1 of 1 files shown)


HB9HIL:
This is on conflict with the new qrg input where you can enter 'khz', 'mhz' and so on to change the unit...

HB9HIL:
See also #1266

phl0:
> This is on conflict with the new qrg input where you can enter 'khz', 'mhz' and so on to change the unit...

Rolled back the changes for the frequency input fields. Ok to merge?

HB9HIL (approved):

HB9HIL (approved):

int2001 (approved):
works on those field (Az/Ele/Power).

phl0:
Idea for the two different approaches to frequency input fields: Make a (user) option to choose between old style (with nuneric keyboard options) and the advanced input options with auto format? @HB9HIL @int2001

int2001:
Both options (user option to deactivate the “intelligent parsing” of the qrg OR setting the field to number via JS when working mobile) means: Both branches must be maintained.

KISS (for user AND code) would be here: Use the button for the unit and put the field into num-type.

just my 2ct

PR #1266, opened 2024-11-26
QRG Unit - allow short input aswell

HB9HIL:
The new QRG input allows to change the QRG unit by input. This is an addition to #1196 

`14.2mhz` changes the unit from whatever to MHz.

Testusers wished also to enter the short variant of it

New:
`14.2m` OR `14.2mhz`  changes the unit from whatever to MHz.

same for `hz`/`h`,  `khz`/`k` and `ghz`/`g` 

Note:
If no qrg unit is entered the currently used QRG unit does not change


int2001 (approved):
works

PR #1270, opened 2024-11-27
fix quicklog feature

HB9HIL:
When I refactored the qso.js I broke the quicklog feature. This fixes it.


int2001 (approved):
Fix works

HB9HIL:
catched NULL/empty in a direct commit https://github.com/wavelog/wavelog/commit/ea0079917e66bac9e46502a59c48f56ec480e86d

PR #1271, opened 2024-11-27
QSL Tab Fix in edit modal

HB9HIL:
Should fix 1269 

issue here aswell the identical ID's just as in #1237


phl0 (approved):
Works for me.

PR #1272, opened 2024-11-27
Numblock for mobile

HB9HIL:
for #1265 

Now aswell on mobile for the frequency inputs. In QSO and Contesting

credits to @phl0


phl0:
Will test (later).

int2001 (approved):
works.
tested:
- on mobile (iPhone) with and without CAT
- on desktop with full-size browser-window and small-window and CAT/no CAT

phl0 (commented):
Tested with Android and iOS clients. Works even in combination with active CAT to update frequencies. Nice solution. Suggest to apply the adv. input stuff to RX frequency as well.

PR #1275, opened 2024-11-27
Callbooks updated in installer

HB9HIL:
Addition to #1260  and #1274


phl0 (approved):
Good point. Code reading looks ok.

int2001 (approved):
couldn't really test, lack of other subscription.
but from cross-reading code this looks ok.

PR #1283, opened 2024-11-29
Fixes JS-Bug when clicked on fresh spot

int2001:
Should fix https://github.com/wavelog/wavelog/discussions/1281


HB9HIL (approved):
tested working

PR #1284, opened 2024-11-29
Exceptionhandling for missing/wrong mode at QRZ

int2001:
QRZ emits the following message, if the mode wasn't set properly (due to wrong import or whatever reason):

`STATUS=FAIL&RESULT=FAIL&REASON=QRZ Internal Error: Unable to add QSO to database.&EXTENDED=qso_hash: required field missing mode1, qso_hash: required field missing mode1,`

this one handles the error and marks qrz as "invalid"

this one belongs to the Log-Sync. Not to the Callbook-Lookup (tnx fer hint @phl0 )


phl0 (approved):
This works and catches the qrz.com error as desired. Tested with mode being '0' as reported and also '' and null. All cases catched:

Screenshot from 2024-11-29 08-35-51

Screenshot from 2024-11-29 08-39-31

Screenshot from 2024-11-29 08-39-46

PR #1286, opened 2024-11-29
Compact date+time and RST QSO fields on mobile

filipmelik:
Sometimes I use Wavelog for logging random ad-hoc contacts via my mobile phone. But when you open Live QSO page on mobile, the whole page is unnecessary tall - I made quick and dirty mod on my instance, to make date&time + RST fields fit on the one line, which in turn saves 2 lines of precious space, and make me scroll a little less to reach the Save button

Before:
before

After: 
after

### DIFF
--- a/application/views/qso/index.php
+++ b/application/views/qso/index.php
@@ -68,7 +68,7 @@
                   <input type="text" class="form-control form-control-sm input_date" name="start_date" id="start_date" tabindex="4" value="<?php echo date('d-m-Y'); ?>" <?php echo ($manual_mode == 0 ? "disabled" : "");  ?> required pattern="[0-3][0-9]-[0-1][0-9]-[0-9]{4}">
                 </div>
 
-                <div class="mb-3 col-sm-6 col-md-6 col-lg-4 col-xl-4 col-4">
+                <div class="mb-3 col-sm-6 col-md-6 col-lg-4 col-xl-4 col-4 ps-0 pe-0">
                 <label for="start_time"><?= __("Time on"); ?></label>
                   <div class="input-group">
                     <input type="text" class="form-control form-control-sm input_start_time" name="start_time" id="start_time" tabindex="5" value="<?php echo $manual_mode == 0 ? date('H:i:s') : date('H:i'); ?>" size="7" <?php echo ($manual_mode == 0 ? "disabled" : "");  ?> required pattern="[0-2][0-9]:[0-5][0-9]">
--- a/application/views/qso/index.php
+++ b/application/views/qso/index.php
@@ -99,12 +99,12 @@
 
               <?php } else {?>
               <div class="row">
-                <div class="mb-3 col-md-6">
+                <div class="mb-3 col-6">
                   <label for="start_date"><?= __("Date"); ?></label>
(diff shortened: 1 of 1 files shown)


phl0 (changes requested):
I like the idea basically. But it has two issues:

The advanced frequency input field is going missing.
current state:
image

with your feature branch:
image

Cannot explain why though. Maybe @HB9HIL has an idea.

With separate QSO times it still has separate lines for time and date:
Screenshot_20241129_080518_Firefox

Maybe you can handle this as well?

HB9HIL:
> I like the idea basically. But it has two issues:
> 
> The advanced frequency input field is going missing. 

I updated the branch. No issue in his PR, just the branch was not on latest dev

HB9HIL (approved):
I like

HB9HIL (changes requested):
oh yeah. but @phl0 is right. please handle "Separat QSO times aswell". Enable in your user settings "Log End Times for QSOs Separately"

filipmelik:
> oh yeah. but @phl0 is right. please handle "Separat QSO times aswell". Enable in your user settings "Log End Times for QSOs Separately"

Thanks for noticing @phl0 , good catch. Fixed. It was kind of tricky to get it right on all the breakpoints, but this looks better now imho :)

HB9HIL:
@filipmelik you need to update your branch to the latest dev

```bash
git fetch 
git merge upstream/dev
```

filipmelik:
> @filipmelik you need to update your branch to the latest dev
> 
> ```shell
> git fetch 
> git merge upstream/dev
> ```
> 
> you force pushed your branch to the pre-version again.
> 
> Alternativly hit the button "Update branch" and perform a `git pull`

sorry, forgot the specify upstream while pulling. should be ok now (forcepushed again), sorry for troubles 🌹

int2001 (approved):
i really like it.

application/views/qso/index.php
```diff
@@ -63,12 +63,12 @@
...
               <?php if ($this->session->userdata('user_qso_end_times')  == 1) { ?>
               <div class="row">
-                <div class="mb-3 col-md-3">
+                <div class="mb-3 col-sm-12 col-md-12 col-lg-4 col-xl-4 col-4">
                   <label for="start_date"><?= __("Date"); ?></label>
                   <input type="text" class="form-control form-control-sm input_date" name="start_date" id="start_date" tabindex="4" value="<?php echo date('d-m-Y'); ?>" <?php echo ($manual_mode == 0 ? "disabled" : "");  ?> required pattern="[0-3][0-9]-[0-1][0-9]-[0-9]{4}">
                 </div>
 
-                <div class="mb-3 col-md-4">
+                <div class="mb-3 col-sm-6 col-md-6 col-lg-4 col-xl-4 col-4">
```
HB9HIL, line 71: to gain a little bit more space for the time we could remove the padding here. Bootstrap comes with some margin so it would be fine. On Desktop and mobile

```suggestion
                <div class="mb-3 col-sm-6 col-md-6 col-lg-4 col-xl-4 col-4 ps-0 pe-0">
```
image

image

HB9HIL (changes requested):

application/views/qso/index.php
HB9HIL: Before:

image

HB9HIL (approved):
great!

phl0 (approved):
Looks fine with end times on mobile now:

photo_5355306666419152414_y

PR #1288, opened 2024-11-29
[Advanced Logbook] More icons

AndreasK79:
Added a few more icons.
Changed to btn-success class on search and action button.
image

Adjusted invalid sql to also check for col_mode = 0


HB9HIL (approved):
I like :-) makes it more clear for the user

int2001 (approved):

PR #1291, opened 2024-11-30
[Gridmap] Added fields to the legend

AndreasK79:
Added fields to the gridmap:
image


phl0 (approved):
Works as expected on the first glimpse.

PR #1294, opened 2024-12-01
[WIP] Extend Anaytics --> Stats with choser for year to be reported

int2001:
It's end of the year. Time to fetch the summary of #QSOs and so on.
Therefore i extended the Views with a dropdown where one can chose the year to be reported.

image


phl0 (approved):
Nice one. Works as expected.

AndreasK79 (approved):

PR #1296, opened 2024-12-02
Restore uppercase transformation for inputs

phl0:
In https://github.com/wavelog/wavelog/pull/1162 we broke CSS text-transformation to upper case for inputs like grid, VUCC grids, callsign and some more. Example:

image

This restores the text-transformation by adding class "uppercase". On the pro side we can get rid of various id-based CSS styles in general.css.


int2001 (approved):
working again

HB9HIL (approved):

PR #1297, opened 2024-12-02
Set QRZ only to "M" if uploaded/invalid before

int2001:
Small bug in LBA.

QRZ was set to "M" even if it was never Uploaded before


phl0 (changes requested):
This does not work on my end. Sets status to modified still even if not uploaded before. Investigating ...

int2001:
> This does not work on my end. Sets status to modified still even if not uploaded before. Investigating ...

pse check agn.

phl0 (approved):
Works for me now :)

PR #1300, opened 2024-12-02
[Activated Gridsquares] Fix after fields was added

AndreasK79:
I broke the activated gridsquares map after I added fields to the gridsquare map.
Here is the fix.


phl0 (approved):
Works (again).

PR #1301, opened 2024-12-02
Country sat records

phl0:
Fix country name in details view (missing DXCC name in SQL query).
Bug happens in SAT distance detail view if one user-definec column is set to Country.


int2001 (approved):
good find. this bug seems to be there for a long time ;)

PR #1302, opened 2024-12-02
Fix country name in timeline views

phl0:
Fixes a bug where country name shows the satellite name in timeline views:

image


AndreasK79 (approved):

int2001 (approved):
cnfm. fixed

PR #1303, opened 2024-12-02
Make orbit dropdown aka. fences everywhere

phl0:
Make orbit type hardcoded.

PR #1304, opened 2024-12-02
LoTW SAT name should not be required

phl0:
The LoTW name should not be required. Based on https://github.com/wavelog/wavelog/pull/1303 which should be merged first.


assets/js/sections/satellite.js
```diff
@@ -108,7 +108,7 @@ function editSatelliteDialog(id) {
 
 function saveUpdatedSatellite(form) {
 	$(".alert").remove();
-	if (form.nameInput.value == "") {
+	if (form.displayNameInput.value == "") {
 		$('#edit_satellite_dialog').prepend('<div class="alert alert-danger" role="alert"><a href="#" class="close" data-dismiss="alert" aria-label="close">&times;</a>Please enter a name!</div>');
```
int2001, line 112: This `&times;` produces a "x" in front of the alert-text. There's no spacing. Maybe add one?!

And: It has no effect in chrome (perhaps i am using it wrong - from reading the code i assume it should close the message?!)

Same at L59

int2001 (changes requested):
Works and is logical. See comment (Cosmetics)

assets/js/sections/satellite.js
phl0: Added a space. The "X" closes the popup (on my Firefox system).

int2001:
Looks better now. Closing the popup isn't working here (seems to be local problem)

phl0:
> Looks better now. Closing the popup isn't working here (seems to be local problem)

Tnx. No issues with closing on Firefox here. Is a separate issue with Chrome which existed before obviously. Will merge after approval of https://github.com/wavelog/wavelog/pull/1303.

Issue #1011, opened 2024-09-28
Installation fails with nginx webserver


HB9HIL:
The Wavelog Installer fails with nginx webserver. 

It's basically the database migration which fails due to a infinite redirect which only appears in default nginx configs. 

Was already mentioned in #917

HB9HIL:
I was able to reproduce this issue. But currently I have no idea how to fix it. Maybe I have some time in the next two weeks to dig into this.

violarulan:
Tested with fresh install on my test machine and it works. The installer can finished by itself with all green.

I used rewrite config as follows

```
    # wavelog.conf
    # some static assets
    location /assets {
        try_files $uri $uri/;
    }
    # ....
    location / {
        try_files $uri $uri/ /index.php?$args;    # after installing, leave $config['index_page'] as empty in application/config.php
        include fastcgi_params;
        fastcgi_pass			unix:/run/php-fpm/php-fpm.sock;
        fastcgi_index			index.php;
        fastcgi_buffers			8 16k;
        fastcgi_buffer_size		32k;
        fastcgi_param DOCUMENT_ROOT	$realpath_root;
        fastcgi_param SCRIPT_FILENAME	$realpath_root$fastcgi_script_name;
    }    
```

HB9HIL:
Okay so probably not a bug and just a configuration issue. Thanks!

hellofinch:
Hello, this is my nginx's configuration file. Are there some wrong? I also leave $config['index_page'] as empty in application/config.php. But it still failed.
@violarulan 
```
server {
    listen 80 default_server;

    client_max_body_size 100M;

    root /app/public;
    index index.php index.html index.htm;

    location /pma/ {
        root /var/www/html;
        index index.php;

        location ~ \.php {
            fastcgi_buffer_size 128k;
            fastcgi_buffers 4 256k;
            fastcgi_busy_buffers_size 256k;

            fastcgi_pass unix:/run/php/php8.3-fpm.sock;
            fastcgi_param SCRIPT_FILENAME $document_root$fastcgi_script_name;
            fastcgi_param SCRIPT_NAME $fastcgi_script_name;
            include /etc/nginx/fastcgi_params;
            fastcgi_read_timeout 300;
        }
    }

    location /wavelog/ {
        root /var/www/html;
        # index index.php;
        try_files $uri $uri/ /index.php?$args;

        location ~ \.php {
            fastcgi_buffer_size 128k;
[...]

PR #1307, opened 2024-12-03
[Installer] better debugging

HB9HIL:
Added a lot of log_messages for better debugging in the installer and fixed the database version check which was accidentially a hard fence. Should be soft instead.

No other logic changes. Review with whitespace hidden.

Helps debugging in #1011 or https://github.com/wavelog/wavelog/discussions/847

### DIFF
--- a/application/controllers/Update.php
+++ b/application/controllers/Update.php
@@ -211,7 +211,7 @@ class Update extends CI_Controller {
         $gz = gzopen($url, 'r');
         if ($gz === FALSE) {
             $this->update_status("FAILED: Could not download from clublog.org");
-            log_message('error', 'FAILED: Could not download exceptions from clublog.org');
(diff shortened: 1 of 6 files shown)


HB9HIL:
Tested on:
- Debian 12
- Linux Mint 21

There was an issue with the database version check. In some circumstance at mariadb the command `$link->server_info` returns not only the version, but also which version the MariaDB does imitate to be backwards compatible. So on Linux Mint 21 with the MariaDB version `10.6.18` which returns on server_info: `5.5.5-10.6.18-MariaDB-0ubuntu0.22.04.1`

This breaks the Version Parser. Therefore I changed this to `SELECT VERSION()` which should work much better on MySQL and MariaDB.

HB9HIL:
Tested successfully on

Debian 12
Linux Mint 21
Docker

phl0 (approved):
Works here.

PR #1308, opened 2024-12-03
Fixes Bug where Modes are not shown correct

int2001:
e.g.: A year (or whole logbook) where only a AM-QSO exists.
Stats-page (Modes) fails with NaN (see picture)

image

This was because AM was never reported.


phl0 (approved):
Works fine for me.

PR #1312, opened 2024-12-04
Check for wrong credentials when doing cron-download / LoTW

int2001:
When invoking LoTW-Download via cron and wrong credentials were provided, the logic tries to parse the HTML returned by LoTW.
This one fixes it.

### DIFF
--- a/application/controllers/Lotw.php
+++ b/application/controllers/Lotw.php
@@ -785,7 +785,7 @@ class Lotw extends CI_Controller {
 				curl_setopt($ch, CURLOPT_RETURNTRANSFER, true);
 				curl_setopt($ch, CURLOPT_CONNECTTIMEOUT, 30);
 				$content = curl_exec($ch);
-				if(!curl_errno($ch)){
+				if(!curl_errno($ch) || str_contains($content,"Username/password incorrect</I>")){
 					file_put_contents($file, $content);
 
 					ini_set('memory_limit', '-1');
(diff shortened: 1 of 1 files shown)


HB9HIL (approved):
Did not test but code looks good. The only question  I had is if it needs the `</I>` in the check string. But I guess you tested it and it's fine :-)

int2001:
The `</I>` is for safety reasons.

application/controllers/Lotw.php
```diff
@@ -696,7 +696,7 @@ function lotw_download($sync_user_id = null) {
 				curl_setopt($ch, CURLOPT_RETURNTRANSFER, true);
 				curl_setopt($ch, CURLOPT_CONNECTTIMEOUT, 30);
 				$content = curl_exec($ch);
-				if(curl_errno($ch)){
+				if(curl_errno($ch) || str_contains($content,"Username/password incorrect</I>")) {
```
phl0, line 699: This leads to a misleading display in case of wrong credentials (because of the added result code of curl which is "no error"):

Screenshot from 2024-12-04 09-30-16

application/controllers/Lotw.php
```diff
@@ -785,7 +785,7 @@ public function import() {	// Is only called via frontend. Cron uses "upload". w
 				curl_setopt($ch, CURLOPT_RETURNTRANSFER, true);
 				curl_setopt($ch, CURLOPT_CONNECTTIMEOUT, 30);
 				$content = curl_exec($ch);
-				if(!curl_errno($ch)){
+				if(!curl_errno($ch) && !str_contains($content,"Username/password incorrect</I>")){
```
phl0, line 788: Same as above but here $result will even be overwritten later.

phl0 (changes requested):
Fix for the mentioned issues in https://github.com/int2001/wavelog/pull/15.

phl0 (approved):
Fine with me now.

PR #1313, opened 2024-12-04
ADIF Standard Version 3.1.5

HB9HIL:
Since there is a new release of the ADIF Standard from Version 3.1.4 to Version 3.1.5 we want to support that in Wavelog. 

This PR prepares the necessary changes and will be worked on by the whole team. 

**Todo:**

- [x] Write a migration to make the necessary database changes
- [x] Set the correct migration version in `application/config/migration.php` (will happen just before merge)
- [x] Adjust ADIF Import
- [x] Adjust ADIF Export
- [x] QSO.php


int2001 (approved):
tested:
- import new Fields
- import with new checkbox "mark DCL as uploaded"
- export

since this one affects a core function, i'd appreciate a double/triple-check by @phl0 and @AndreasK79 as well

AndreasK79:
We need to up the ADIF version:

> Wavelog ADIF export
<ADIF_VER:5>3.1.4
<PROGRAMID:7>WaveLog
<PROGRAMVERSION:3>1.9
<EOH>

HB9HIL:
> We need to up the ADIF version:
> 
> > Wavelog ADIF export
> > <ADIF_VER:5>3.1.4
> > PROGRAMID:7WaveLog
> > PROGRAMVERSION:31.9

Oh good point. missed that one

AndreasK79:
We should also fix the TODO for QSLmsg.
Seems it can mostly be fixed by placing QSLmsg in import_bulk instead.

And if simplefle's save_qsos use import_bulk instead, I think that issue should be fixed.

PR #1318, opened 2024-12-05
Don't save and process empty files for QRZ

int2001:
observed hundreds of `qrzcom_download_report_[hash].adi`-Files which were in a "staled"-state and digged in.
Content: `RESULT=FAIL&COUNT=`. not more, not less.

this one prevents saving and parsing them.

Repro: Put in wrong creds for QRZ and let cron run for QRZ-download run.


AndreasK79 (approved):

phl0 (approved):
Works here.

PR #1323, opened 2024-12-05
Resolve PHP errors and display angles correctly

phl0:
Az/El as well as elevation arrows are not displayed in the correct angle due to some PHP code error:

image

image

Broken since https://github.com/wavelog/wavelog/pull/889 with commit https://github.com/wavelog/wavelog/pull/889/commits/07edb146b747a7e72d031eaca0894471b169a459


HB9HIL (approved):
works. good find

int2001 (approved):
this brings the arrows back to function. works

PR #1331, opened 2024-12-07
[Distance unit] Dashboard and logbook overview shows distance in set unit

### LINKED ISSUE #1324: User distance setting set to miles but overview distance column shows kilometers
I am using the latest version of the container under docker compose, which reports version as 1.8.8 in the debug menu. 

Server Software | Apache/2.4.62 (Debian)
PHP Version | 8.3.12 OK
MySQL Version | 11.3.2-MariaDB-1:11.3.2+maria~ubu2204
Codeigniter Version | 3.2.0-dev

**Steps to Reproduce**
Account -> Wavelog preferences
  Select Distance as a logbook column. 
  Select Miles as measurement distance.
  Scroll to bottom and click save.
Return to Logbook->Overview, observe distance column in km.

**Expected Behavior**
I expect this column to be displayed in miles.

**Additional Context**
I also noticed the account date format does not carry over to the Past QSO entry page which is where it is most needed but I found a separate bug for this and it looks like a more complicated issue.

AndreasK79:
Fixed #1324

### DIFF
--- a/application/locale/bg_BG/LC_MESSAGES/messages.po
+++ b/application/locale/bg_BG/LC_MESSAGES/messages.po
@@ -6,7 +6,7 @@
 msgid ""
 msgstr ""
 "Report-Msgid-Bugs-To: translations@wavelog.org\n"
-"POT-Creation-Date: 2024-12-06 09:23+0000\n"
+"POT-Creation-Date: 2024-12-06 12:16+0000\n"
 "PO-Revision-Date: 2024-11-01 08:53+0000\n"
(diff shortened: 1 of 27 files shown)


phl0 (changes requested):
Works fine for the Logbook overview. However discovered that QSO view could also need a fix here:

Screenshot from 2024-12-07 21-25-34

Fix in https://github.com/AndreasK79/wavelog/pull/39

phl0 (approved):

PR #1332, opened 2024-12-07
Satpass date

AndreasK79:
Added date so that you can run predictions on different dates:
image

Added TLE column to show if sat has TLE in sat table:
image

Added update button for TLE under debug (and cron job):
image

### DIFF
--- a/application/config/migration.php
+++ b/application/config/migration.php
@@ -22,7 +22,7 @@ $config['migration_enabled'] = TRUE;
 |
 */
 
-$config['migration_version'] = 229;
+$config['migration_version'] = 230;
 
 /*
 |--------------------------------------------------------------------------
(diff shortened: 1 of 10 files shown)


phl0 (changes requested):
image

I'm sorry ... :D

phl0 (approved):
It works fine here.

int2001 (approved):
i like it.
perhaps add a `*` in one of the next PRs to plan the day by seeing all passes? :)

phl0:
> i like it. perhaps add a `*` in one of the next PRs to plan the day by seeing all passes? :)

And maybe set default times to 00:00 and 24:00?

phl0 (approved):
Latest version running fine now.

PR #1336, opened 2024-12-09
[Sat Sked] Can now predict sat skeds

AndreasK79:
This is a work in progress


phl0 (approved):
Tested and works fine here.
(Min / Max time need some love though) :)

PR #1338, opened 2024-12-09
Function to add/update SAT data from tqsl/LoTW config file

phl0:
This adds an update function to read valid SAT names from tqsl config file. This contains a list of valid SAT names and their acceptance status by LoTW. We parse the gzipped XML file and:
- Insert SAT if not contained in local database or
- Update LoTW status if SAT is present but set to not accepted by LoTW
Other SATs are ignored for now.


application/models/Update_model.php
```diff
@@ -359,13 +359,87 @@ function tle() {
...
+		curl_setopt($curl, CURLOPT_BINARYTRANSFER,true);
+		curl_setopt($curl, CURLOPT_TIMEOUT, 10);
+
+		$response = curl_exec($curl);
+		if (curl_errno($curl)) {
+			echo __('cURL error:').' '.curl_error($curl).' ('.curl_errno($curl).')';
+			return;
+		}
+		curl_close($curl);
+		$xml = simplexml_load_string(gzdecode($response));
```
int2001, line 387: perhaps adding a try/catch around it?
don't know what happens if there's a 200 but with some kind of crap/empty body?

application/views/lotw/satupdate.php
```diff
@@ -0,0 +1,13 @@
+<div class="container">
+<h2><?php echo $page_title; ?></h2>
+
+<?php
+print '<table class="table table-striped table-hover">';
```
int2001, line 5: Minor thing:
why print where we use echo all times?

afaik print returns a boolean if printed while echo doesn't. But can't see a logic where the result is needed...

application/views/lotw/satupdate.php
```diff
@@ -0,0 +1,13 @@
...
+print '<tr><th>'.__('Name').'</th><th>'.__('Display Name').'</th><th>'.__('Start Date').'</th><th>'.__('End Date').'</th><th>'.__('Status').'</th></tr>';
+foreach ($satupdates as $sat) {
+   print('<tr><td>'.$sat['name'].'</td><td>'.$sat['displayname'].'</td><td>'.$sat['startDate'].'</td><td>'.$sat['endDate'].'</td><td>'.$sat['status'].'</td></tr>');
+}
+print '</table>';
+?>
```
int2001, line 11: maybe add a link to TLE-Update here? just as an idea to make it clearer...

int2001 (changes requested):
functionality is given.

Some minor-changes requested ;)

int2001 (approved):
works. very nice

PR #1342, opened 2024-12-11
Skip Users without any station_id at lotw_download

int2001:
Otherwise a SQL which checks the QSOs for existance will faili


HB9HIL (approved):
should work

phl0 (approved):
Tested with empty profiles/locations and works. Even reduces runtime as without this patch the PHP would run for like 5m on my end. Now instant result. Very nice :)

PR #1343, opened 2024-12-12
Fix correct check for empty date

phl0:
Check for last LoTW date fails if none can be found. $row is set in any case and `$row->MAXDATE`  is NULL. So `isset($row)` will always be true.

Added debug output shows:

```
object(stdClass)#47 (1) { ["MAXDATE"]=> NULL }
```

image

So we need to check `$row->MAXDATE` for being != NULL.


int2001 (approved):
tested and works

PR #1334, opened 2024-12-08
Clubstations for Wavelog

HB9HIL:
# Clubstations

This is probably one of the most requested features in Wavelog. Now it's becoming reality. 

> [!IMPORTANT]
> This is still a work in progress. You can test it but it can change at any time.

Everything also in the first draft of a Wiki: https://github.com/wavelog/wavelog/wiki/Clubstations

## Clubstations and it's Members

image

You can enable this feature with the config item `special callsign`. It will allow the administrator of the Wavelog to create new Clubstations. Also he can add users of the local Wavelog Instance to this Club. I implemented two types of ..

## Permission Levels

Currently implemented are two permission Levels

- Club Member
   This user can log QSOs and edit QSO which were done by himself. He can see other QSOs but can't edit or delete them. No access to the QSL Management.
- Club Officer
   Officers can manage Club Members and their permissions, also he can edit/delete all QSOs and has access to the QSL Management.

This is how it's currently planned. (There is nothing fixed yet, we can discuss about it. Also not everything is implemented yet)

image

## Restrictions

Next to the restrictions mentioned above about editing QSOs, there are (unfortunately) some more. One big restriction is that Contesting is not available (yet) for Clubstations. Reason is our Architecture and that it's not possible at the moment to have browser-session based "contest sessions". We will working on that but this will take some more time. 

## How it works

Clubstations are basically normal user-accounts with a special flag. So users which are members of this clubstation can "impersonate" this user account. There is button in the header menu when the clubstation feature is enabled. Is the user an Officer of the club he see's also the button for the club permission management otherwise only the call of the clubstation. 

image

Once switched to the clubstation it's also super easy to go back into the personal account.

image

PR #1349, opened 2024-12-13
First Login Wizard

HB9HIL:
If you create a new user after a successfull installation we want to guide this user for the first login. It's about the stationsetup which has to be done before any QSOs can be logged. For this we introduce the First Login Wizard

It's only shown if
- The user is no admin
- The user is currently not impersonated 
- It's no Clubstation
- and a entry in the database is not 1 (true = was already shown before)

If the database entry does not exists we check if the user has any station locations. if not we show him the wizard.

image

This PR depends on #1334 why it's actually drafted.


phl0 (approved):
Works fine. Some hints for zone loookups could be useful however: https://github.com/HB9HIL/wavelog/pull/124

phl0 (approved):
Looks very fine to me now :)

PR #1357, opened 2024-12-17
Add option to enable eqsl massdownload

phl0:
The eqsl.cc massdownload function is not thread safe. So we added a config switch to disabled it. Default setting is disabled. Should not be enabled on multi-user instances.


int2001 (approved):
Tested:
- without option at config.php --> Works
- with false option at config.php --> Works
- with true option at config.php --> Works

PR #1360, opened 2024-12-19
Adding ability to check LoTW-Credentials

int2001:
Added REST-Service which checks the validity of the user-provided LoTW-Credentials.

Goal:
Implement Check-Button like for QRZ as well for lotw at user-settings

### DIFF
--- a/application/controllers/Lotw.php
+++ b/application/controllers/Lotw.php
@@ -754,7 +754,6 @@ class Lotw extends CI_Controller {
 		$lotw_url .= "login=" . $data['user_lotw_name'];
 		$lotw_url .= "&password=" . $data['user_lotw_password'];
 
-		log_message("Error",$lotw_url);
 		$ch = curl_init();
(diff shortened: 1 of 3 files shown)


application/controllers/Lotw.php
```diff
@@ -723,6 +723,88 @@ function lotw_download($sync_user_id = null) {
...
+			exit();
+		}
+		$ret=[];
+		$ret['status']='';
+
+
+		$raw = file_get_contents("php://input");
+		try {
+			$obj = json_decode($raw,true);
+		} catch (e) {
```
HB9HIL, line 740: This will probably not work as expected as `e` is not defined. Should be defined as Exception event


```suggestion
		} catch (Exception $e) {
```

int2001: Done too much JS the last time. Tnx

phl0 (approved):
Works for me (corrected some wording and translation).

phl0 (approved):

PR #1361, opened 2024-12-19
Filter on valid continents only

AndreasK79:
Only show statistics results from valid continents in the database.

#fenceseverywhere


phl0 (approved):
Works fine.

PR #1363, opened 2024-12-19
Improve import regarding the QSLMSG

HB9HIL:
We need to reduce the amount of querys in the import function due the changes of adif 3.1.5


AndreasK79 (approved):

AndreasK79:
Always nice to get rid of one query for each qso import :)

PR #1367, opened 2024-12-21
Fixes "sticking" Mail

int2001:
Fixes Bug with Mails as mentioned in https://github.com/wavelog/wavelog/discussions/1366

was introduced in #1248


HB9HIL (approved):
fix confirmed

PR #1369, opened 2024-12-23
WAE award

AndreasK79:
This PR adds the WAE award to Wavelog.

I am drafting it, since we do not have a way of setting region (needed for WAE countries) in the GUI.


phl0 (approved):
Tested and works fine (with my latest commit) :)

int2001 (approved):
checked - works

PR #1370, opened 2024-12-23
[Advanced Logbook] Added region to columns

AndreasK79:
Added column region.
Added batch edit for region.


phl0 (approved):
Works (now). Nice addition :)

PR #1374, opened 2024-12-23
[Logbook Advanced] Added a dropdown for region edit

AndreasK79:
Turns out we can use a dropdown for region:
image


phl0 (approved):
Works a treat :)

PR #1378, opened 2024-12-24
CAT radio timeout alert fix

filipmelik:
Due to typo(?) in the /radio API response, the timeout alert was never displayed, as the JS code rely on `updated_minutes_ago` key instead of `update_minutes_ago`

Before fix:
Screenshot 2024-12-23 at 20 57 42

After the fix:
Screenshot 2024-12-23 at 20 52 35

### DIFF
--- a/application/controllers/Radio.php
+++ b/application/controllers/Radio.php
@@ -238,7 +238,7 @@ class Radio extends CI_Controller {
 					if (isset($cat_url) && ($cat_url != null)) {
 						$a_ret['cat_url'] = $cat_url;
 					}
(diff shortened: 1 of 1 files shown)


phl0 (approved):
Nice find. The bug has been sitting there since Nov 2023. Introduced in https://github.com/wavelog/wavelog/commit/cce52c126c3f48f0821d39aaae477a0e41b848eb#diff-6bba3d2684c123fdf4da934ef6db36eba2115144a67de6558034dacdec9c9bc9R170.

PR #1379, opened 2024-12-24
Fix cat radio UI

filipmelik:
In case only some of the optional/complementary CAT information was displayed in the CAT alert, it might end up in extra spaces, because the JS code relied on fact, that all the previous properties will also be filled. This should fix the extra space and clean up the code little bit.

Before:
 
Screenshot 2024-12-23 at 20 57 42

After:

Screenshot 2024-12-24 at 8 27 46


phl0 (approved):
Tested and works. Tnx for the fix.

PR #1380, opened 2024-12-25
Fixes bug at JS when trying to log call with multiple zeros

int2001:
B4:
Try to log OZ400POST - won't work / won't lookup

after this patch:
Works again.


phl0 (approved):
Works as designed 😊👍
Tnx for fixing my mistakes 😬

Issue #1384, opened 2024-12-26
Post Bluesky Tweets, Like Mastodon


ethancedwards8:
**What Feature is Missing?**
The ability to tweet on Bluesky about QSOs, just like the current Mastodon functionality.

**Why is This Feature Important?**
It can bring more exposure to wavelog and amateur radio. 

**Is This Feature Personal or Beneficial to Others as Well?**
It would benefit me as I mainly use Bluesky, but I think it would help others.

**Implementation** 
I don't think it would be too challenging to implement. Will look into it in the coming days.
https://github.com/wavelog/wavelog/blob/c0fd01e0010fd63fde9efd4efd341df34e11fba1/application/views/view_log/qso.php

PR #1385, opened 2024-12-26
views: log: qso: add post to bluesky button

ethancedwards8:
Added post to Bluesky button, hopefully encrouaging more people to use wavelog. SVG is used because the Unicode Character used by font-awesome does not render on most modern browsers for some reason. SVG works pretty universally.

Fixes: https://github.com/wavelog/wavelog/issues/1384


HB9HIL:
Hello @ethancedwards8 

reason why the bluesky icon of fontawesome is not rendered ist because Wavelog uses still fontawesome version 5 at the moment. Bluesky is not implemented in V5. (we did not upgrade to v6 yet because this would change a lot of icons)

so it's fine to use the SVG 👍

HB9HIL (approved):

HB9HIL:
@ethancedwards8 

Thanks for your contribution and welcome to Wavelog :)

PR #1392, opened 2024-12-26
[Advanced Logbook] Show name of region

AndreasK79:
Now shows name in LBA instead of just the region abbreviation.

image


phl0 (approved):
Works for me.

PR #1399, opened 2024-12-27
Share Modal

HB9HIL:
After a button for Bluesky was added in #1385  the amount of buttons in the QSO details view is kinda high. I bult a modal which is easy to scale and does not clutter the QSO details modal with social network buttons

image

### DIFF
--- a/assets/js/sections/common.js
+++ b/assets/js/sections/common.js
@@ -1159,6 +1159,31 @@ function enableMap() {
     map.keyboard.enable();
 }
 
+function shareModal(qso_data) {
+    console.log(qso_data);
+    $.ajax({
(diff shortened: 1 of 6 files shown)


assets/js/sections/common.js
```diff
@@ -1159,6 +1159,31 @@ function enableMap() {
     map.keyboard.enable();
 }
 
+function shareModal(qso_data) {
+    console.log(qso_data);
```
int2001, line 1163: if removed, i am fine ;)

int2001 (changes requested):
tested - works. pse remove debug-output

int2001 (approved):

PR #1400, opened 2024-12-27
show eqsl images in LBA in the same way as in Logbook Overview

### LINKED ISSUE #1398: View eQSL Card on Advanced Logbook Page
**What Feature is Missing?**
The down arrow for eQSL received should display the eQSL card on the advanced logbook page like it does for the general logbook page.

**Why is This Feature Important?**
It feels unintuitive for the feature to be on one page but not the other.

**Is This Feature Personal or Beneficial to Others as Well?**
I think it would benefit others as well.

HB9HIL:
fix #1398

### DIFF
--- a/src/QSLManager/QSO.php
+++ b/src/QSLManager/QSO.php
@@ -726,7 +726,7 @@ class QSO
 
 		$eqslstring .= '>&#9650;</span>';
 
-		$eqslstring .= '<span ';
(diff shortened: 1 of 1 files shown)


int2001 (approved):
good one. hopefully people are getting more silent now about the missing "pseudo-download"

PR #1401, opened 2024-12-27
Put some dropdowns side by side

AndreasK79:
QSO entry before:
image

QSO entry after:
image

QSO edit before:
image

QSO edit after:
image


int2001 (approved):
checked - is working and looks much better now

PR #1402, opened 2024-12-27
CleanUp primary_subdivisions and adif_modes before doing mig

int2001:
primary_subdivision-table has no PK.

this is no problemm but in combination with no downgrade here it could(!) produce dupes.
This one solves the problem, by:

- [x] Cleaning up table from (technical) dupes
- [x] Cleaning up table from logical dupes
- [x] Adding a UNIQUE_KEY
- [x] Take care of the UNIQUE_KEY

Same problem at adif_modes. This PR cleans up the table and adds a proper unique_key

### DIFF
--- a/application/migrations/231_adif_315.php
+++ b/application/migrations/231_adif_315.php
@@ -145,6 +145,7 @@ class Migration_adif_315 extends CI_Migration {
 		$this->db->query("DELETE FROM `primary_subdivisions` WHERE `adif`=15 and `state`='PM' and `subdivision`='Permskaya Kraj'");
 		$this->db->query("DELETE FROM `primary_subdivisions` WHERE `adif`=206 and `state`='BM' and `subdivision`='Bruck/Mur'");
 
+		// Add Unique IDX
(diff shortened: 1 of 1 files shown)


application/migrations/231_adif_315.php
```diff
@@ -8,542 +8,574 @@
...
+		// ############################################################################################################
+
+		/**
+		 * 2. Primary Subdivisions
+		 * 
+		 * https://adif.org/315/ADIF_315_annotated.htm#Primary_Administrative_Subdivision
+		 */
+
+		// Clean Up primary_subdivisions from dupes, since there's no PK!!
+		$this->db->query("CREATE TABLE tmp_psd AS SELECT DISTINCT ps.`adif`,ps.`state`,ps.`subdivision`,ps.`deprecated` FROM `primary_subdivisions` ps");
```
HB9HIL, line 141: At this point in Code the column `ps.deprecated` does not exist. It will be created in Line 156

HB9HIL (approved):
tested sucessfully in docker

AndreasK79:
Tested with no errors here. Set migration to 230, then upgraded. No errors in the log, and indexes were created.

AndreasK79 (approved):

phl0 (approved):
Works as designed here (not sure why I had errors before) ...

PR #1407, opened 2024-12-29
If there was an error at inserting QSO via API, return 400

int2001:
Reproduce:

insert a QSO with a different station_callsign than at station_profile. This one returns (in latest master/dev) a 201 (insert) but with "adif_error"-count > 0. And QSO wasn't inserted!!

With this PR a 400 would be returned and the messages-object will contain further information.

Part of resolving https://github.com/wavelog/WaveLogGate/issues/26

Testing: Try to log a QSO with the API as described here: https://github.com/wavelog/wavelog/wiki/API#apiqso
Vary the station_callsign within the payload

### DIFF
--- a/application/controllers/Api.php
+++ b/application/controllers/Api.php
@@ -240,7 +240,7 @@ class API extends CI_Controller {
 			} else {
 				$return_msg[]=$custom_errors;
 				http_response_code(400);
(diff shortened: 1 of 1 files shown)


HB9HIL:
will check later today

phl0 (approved):
Good idea. Works fine on my end.

PR #1411, opened 2024-12-30
[Advanced Logbook] Added automatic select when clicking on text input

AndreasK79:
Added automatic select of text when clicking on the inputs having a * in them.
This makes it easier searching for what you want, since you don't have to erase or mark and erase before writing.


int2001 (approved):
good one. sppeds up handling and works

PR #1415, opened 2024-12-31
Update DXCC coming from LoTW to Log

int2001:
Take DXCC which was provided from LoTW and Update Log-Entry on confirmation with it.

Same logic for DXCC as for Grids


phl0 (approved):
Works. Before:

image

After:

image

Issue #1403, opened 2024-12-27
Saved Preferences on Advanced Logbook Page


ethancedwards8:
**What Feature is Missing?**
Advanced logbook preferences (like station location, # of results, etc) should be saved. Each time I go off the page these preferences reset. 

**Why is This Feature Important?**
It is annoying and error prone to have to set all the preferences each time.

**Is This Feature Personal or Beneficial to Others as Well?**
This feature will be beneficial to all concerned.

ethancedwards8:
I'd be happy to do the legwork on this PR, I'm just not sure where to start.

HB9HIL:
I justed discussed this with @AndreasK79. One way would be to store the current values in `localStorage`. Would be reset on page clearing Browser Cache or hit "Reset". 

Tbh I don't see this stored in database. Only in browser. 

Questionable is: Store everything? Like
also filters or only certain stuff. 

Implementation won't be that hard but we should think that through.

ethancedwards8:
Good question. I'm not sure. Definitely the number of QSOs and the station location at a minimum. Maybe not filters. 

Hopefully others will provide some feedback.

AndreasK79:
Never thought about adding preferences to the advanced logbook, but I don't see why not.
For now, I guess #results and locations are enough. I really don't see the point in storing filters, as that usually changes.

If you would like to try to add this, go ahead. As @HB9HIL mentioned, use localStorage to save this.

ethancedwards8:
Solved by: https://github.com/wavelog/wavelog/pull/1576

PR #1419, opened 2025-01-01
DRAFT: Add Persistent Preferences On Advanced Logbook Page

ethancedwards8:
This is in response to issue: https://github.com/wavelog/wavelog/issues/1403

I have encountered an issue, however.

### DIFF
--- a/application/views/logbookadvanced/index.php
+++ b/application/views/logbookadvanced/index.php
@@ -18,11 +18,6 @@
 	var lang_gen_hamradio_ituzone = '<?= __("ITU Zone"); ?>';
 	var lang_gen_hamradio_cqzone = '<?= __("CQ Zone"); ?>';
 
-	/*Pull from localStorage to set form input value*/
-	if (localStorage.hasOwnProperty('qsoresults')) {
-		document.getElementById('qsoResults').value = localStorage.getItem('qsoresults');
-	}
-
     <?php
     echo "var homegrid ='" . strtoupper($homegrid[0]) . "';";
     if (!isset($options)) {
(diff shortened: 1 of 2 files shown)


ethancedwards8:
I would think what I have would work, but for some reason when I run this code I am unable to retrieve any QSOs. In any case, it is 2AM here so I am sleeping, but I am making progress (or some semblance of it).

ethancedwards8:
Oh, I think it may be because the element is not yet created, so it cannot find it and then errors. I have an idea.

ethancedwards8:
Moving the script to the bottom of the file did in fact not work. Sleeping now. Let me know if y'all think of something.

AndreasK79:
Remove the js from index.php and put it in $(document).ready(function () { in logbookadvanced.js like this:

image

ethancedwards8:
> Remove the js from index.php and put it in $(document).ready(function () { in logbookadvanced.js like this:
> 
> image

That worked. Thanks. 

I'm having trouble getting selectedlocations to work. I cannot dedicate much time to this the next few days as I have college applications due.

AndreasK79:
Locations are a bit more tricky, since we use a plugin for multiselection.
However, I got this to work:

> 	if (localStorage.hasOwnProperty('selectedlocations')) {
		const selectedLocations = localStorage.getItem('selectedlocations');
		const locationsArray = selectedLocations ? selectedLocations.split(',') : [];
		$('#de').multiselect('select', locationsArray);
	}

ethancedwards8:
> Locations are a bit more tricky, since we use a plugin for multiselection. However, I got this to work:
> 
> > if (localStorage.hasOwnProperty('selectedlocations')) {
> > const selectedLocations = localStorage.getItem('selectedlocations');
> > const locationsArray = selectedLocations ? selectedLocations.split(',') : [];
> > $('#de').multiselect('select', locationsArray);
> > }

Unfortunately, that does not work for me on Firefox. Are you using a Chromium engine? Would that even make a difference?

AndreasK79:
Yeah, it's based on Chromium, but I tried in Firefox, and they behave the same. The active location is always set in the dropdown. If you modify the dropdown in index.php, the active location will not be set, but the javascript will do it. Problem is that if you don't have localStorage set, no location will be set. So a fallback is needed there.

How is it not working for you?

ethancedwards8:
When I go to the advanced logbook page I get an alert that says: `You need to select at least 1 location to do a search!`.

PR #1382, opened 2024-12-25
views: debug: add System Time

ethancedwards8:
Added system time column to debug page. Useful in debugging system time information and contextualizing cron jobs.

Fixes: https://github.com/wavelog/wavelog/issues/1381

PR #1420, opened 2025-01-01
Add custom date format for system time display

HB9HIL:
@ethancedwards8 added in https://github.com/wavelog/wavelog/pull/1382 the server system time to the debug view. This is great for debugging, but to be consistent we should show the date in the user's date format


int2001 (approved):
works...

PR #1421, opened 2025-01-01
Remove doubling WWFF-Ref to SIG-Fields

int2001:
Special-Treatment for WWFF-References has been introduced in 2022.
Reason was - according to the PR - "some third party tools need the WWFF-Information at SIG-Fields".

Since we try to be as near as possible to the ADIF-Standard, and WWFF itself uses its own fields, we should drop this treatment after 2 years.

Testing:
- log a QSO with WWFF-Reference
- Export it
- You'll see WWFF-Reference at WWFF-Field as well as at SIG_ Fields

after applying patch:
- log a QSO with WWFF-Reference
- Export it
- You'll see WWFF-Reference at WWFF-Field.
- No more redundancy with SIG-Field

### DIFF
--- a/application/libraries/AdifHelper.php
+++ b/application/libraries/AdifHelper.php
@@ -234,25 +234,11 @@ class AdifHelper {
 
 		$line .= $this->getAdifFieldLine("MY_CNTY", $county);
 
-		$stationsSig = $qso->station_sig;
-		// If MY_SIG is WWFF or "" and there's a station_wwff set, use data from station_wwff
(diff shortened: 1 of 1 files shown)


dg2ron:
that's great. 
This redundancy is completely unnecessary. There is the WWFF field and it should be used.
A nice side effect is that when importing, the WWFF information in the SIG field is also removed.
No more duplicate entries.

AndreasK79 (approved):

PR #1422, opened 2025-01-01
[QSO View] Show confirmation info even if date is not set.

### LINKED ISSUE #1356: QSL/eQSL status not always shown in QSO detail view
This is for reporting bugs with the code, if this issue is regarding setting up on your local server please use the discussions area

Describe the bug
When opening the QSO detail view by clicking on the corresponding callsign, the line "This QSO was confirmed on..." either by eQSL or paper QSL is not always shown, even though the QSO has been confirmed by one or both QSL types which is correctly shown in the list by the green up/down arrows.
I suspect this is for QSOs that I have imported from another logbook software via ADIF file, where only the status of QSL/eQSL was present, but not the date when QSL has been sent/received.

Version Information

    OS: Ubuntu Server 24.04.1 LTS
    PHP: 8.3.6
    MariaDB: 10.11.8
    Wavelog: 1.9

Steps to Reproduce

Click on any QSO entry where QSL/eQSL is marked as sent and received, but (presumably) no date information is present.

Expected Behavior
I think it would be good if the fact that the QSO has been confirmed by QSL/eQSL is displayed regrdless of the presence of date info. Perhaps the line could be changed to simply "This QSO was confirmed." and in case a date is available it could be appended in brackets or in a second line like "Date: 3/3/24" or something like that.

Desktop Information:

    OS: Windows 11 Pro
    Browser: Firefox
    Version: 132.0.2

Best 73 and thank you for wavelog, it's awesome!
Sebastian DL2KSE

AndreasK79:
If no date is set, confirmation will be displayed like this:
image

Fixes #1356.

### DIFF
--- a/application/views/view_log/qso.php
+++ b/application/views/view_log/qso.php
@@ -531,9 +531,19 @@
                     <p><?= __("This QSO was confirmed on"); ?> <?php $timestamp = strtotime($row->COL_LOTW_QSLRDATE); echo date($custom_date_format, $timestamp); ?>.</p>
                     <?php } ?>
 
+					<?php if($row->COL_LOTW_QSL_RCVD == "Y" && $row->COL_LOTW_QSLRDATE == null) { ?>
+                    <h3><?= __("LoTW"); ?></h3>
+                    <p><?= __("This QSO is confirmed on LoTW."); ?></p>
+                    <?php } ?>
+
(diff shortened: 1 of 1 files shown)


int2001 (approved):
works.
tested with altering the status at DB

PR #1426, opened 2025-01-01
Trigger recheck work/cnf state on modechange

int2001:
Bug at Worked/Confirm-check.

Example:

- Go to live-QSO
- Set Mode manual to LSB and Band to 40m
- Enter DJ7NT
- save it
- do the same - Callframe becomes yellow/orange because worked.
- Enter DJ7NT
- Set Mode manual to CW and Band to 40m
- Callframe stays orange, even if DJ7NT was never worked in CW on that band.

This one fixes it.


HB9HIL (approved):
works. good find!

PR #1433, opened 2025-01-02
[WIP] Cluboptimizing

int2001:
Some optimizations around the Club-Feature:

- [x] Remove fetching of all users, when showing permission-page (it is never used)
- [x] Fixes missing Radios, because statement of fetching them was wrong.
- [x] Added Radio-Logic as well to Bandmap
- [x] Catch Error if OP isn't filled when logging via API
- [x] Make migration downgrade failsafe(r) by adding if exists for columns


HB9HIL:
With permission of @int2001 we merge this one to have the most serious fixes in dev

PR #1437, opened 2025-01-02
Also fill email upon single callbook update

phl0:
Guess we should also process the email address here.


phl0:
Also changed the update of locator field as this seems to have broken in https://github.com/wavelog/wavelog/commit/96ba26d

int2001 (approved):
works.
tested:
- prefilled Grid (doesn't update - as designed)
- empty Grid (does update - as designed)

Issue #1394, opened 2024-12-26
POTA References Aren't Imported when Importing from ADI files


ethancedwards8:
**Describe the bug**
When attempting to upload a POTA activation the QSOs do not retain POTA reference information for either operators (activator (me) or hunter for p2p contacts)

**Version Information**
We need some information about your versions. You can get all Information in the Admin > Debug Menu
- OS:               Fedora Linux + Docker
- PHP:              8.3.15
- MySQL/MariaDB:    11.3.2-MariaDB-1:11.3.2+maria~ubu2204
- Wavelog:          1.9 (dev branch)

**Steps to Reproduce**
1. Perform POTA activation in Hamrs
2. export ADI file
3. Import file to wavelog
4. check QSO log. POTA references are not there.

**Expected Behavior**
POTA references should be imported form ADI files. Specifically my_sig_info (activator) and sig_info (hunter) fields.

**Screenshots**
Include screenshots if applicable to illustrate the issue.

**Desktop Information:**
- OS:               MacOS
- Browser:        Brave

phl0:
This is a known issue. Wavelog uses the POTA specific ADIF fields for that information. But up to now even POTA itself does not make use of those and stil relies on SIG amd SIG_INFO. From my point of view POTA and corresponding software should make use of the specifically foe them introduced fields here.

ethancedwards8:
What do you suggest users do in the meantime?

phl0:
> What do you suggest users do in the meantime?

I will look into it if we implement a (temporary) workaround.

phl0:
I had a play with hamrs. To be honest the export from the app is a bit confusing:

```
<ADIF_VER:5>3.1.4
<CREATED_TIMESTAMP:15>20250102 214915
<PROGRAMID:5>HAMRS
<PROGRAMVERSION:6>2.25.6
<EOH>

<RST_RCVD:2>59
<RST_SENT:2>59
<CALL:6>DN1YWF
<OPERATOR:5>DF2ET
<FREQ:2>14
<BAND:3>20m
<MODE:3>SSB
<TX_PWR:2>10
<POTA_REF:7>DE-0795
<COMMENT:4>test
<TIME_ON:6>214857
<QSO_DATE:8>20250102
<NAME:7>Florian
<QTH:11>Tippelsberg
<STATE:2>NW
<GRIDSQUARE:6>JO31om
<SIG_INFO:7>DE-0795
<MY_SIG_INFO:0>
<EOR>
```

[...]

PR #1443, opened 2025-01-03
Skip importing sig_info if sig is not set

phl0:
During hacking around with https://github.com/wavelog/wavelog/issues/1394 I discovered that hamrs would export POTA references in POTA and SIG_INFO field but without SIG itself being set to "POTA". So we import a SIG_INFO without knowing which activity program/group this belongs to. IMHO we should skip SIG_INFO in case SIG is not set. See https://github.com/wavelog/wavelog/issues/1394#issuecomment-2568429230.

### DIFF
--- a/application/models/Logbook_model.php
+++ b/application/models/Logbook_model.php
@@ -4222,6 +4222,14 @@ class Logbook_model extends CI_Model {
 				$qslmsg = $station_qslmsg;
 			}
 
+			// Only import SIG_INFO and SIG_INFO_INTL if SIG is set. Discard otherwise as we do not now which activity group the reference belongs to
+			if (!isset($record['sig']) || $record['sig'] == '') {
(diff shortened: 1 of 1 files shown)


application/models/Logbook_model.php
```diff
@@ -4222,6 +4222,14 @@ function import($record, $station_id = "0", $skipDuplicate = false, $markClublog
 				$qslmsg = $station_qslmsg;
 			}
 
+			// Only import SIG_INFO and SIG_INFO_INTL if SIG is set. Discard otherwise as we do not now which activity group the reference belongs to
+			if (!isset($record['sig']) || $record['sig'] == '') {
```
HB9HIL, line 4226: maybe bit cleaner and easier to read :)

```suggestion
			if (empty($record['sig'])) {
```

HB9HIL:
Nice find!! I really like

PR #1444, opened 2025-01-03
Fix CORS error with installer

patrickrb:
Resolves #1376

Updated the migrate and dxcc urls in the installer to point to the same host that the user navigated to.  This keeps the user on the same URL/host/domain that the user originally navigated to, preventing the dreaded CORS error.

Original CORS error:

image

Updated installer working

### DIFF
--- a/install/run.php
+++ b/install/run.php
@@ -297,7 +297,7 @@
 
         return new Promise((resolve, reject) => {
             $.ajax({
-                url: url: `${window.location.origin}/index.php/update/dxcc`,
+                url: `${window.location.origin}/index.php/update/dxcc`,
                 success: async function(response) {
                     if (response == 'success') {
                         running(field, false);
--- a/install/run.php
+++ b/install/run.php
@@ -268,7 +268,7 @@
 
         return new Promise((resolve, reject) => {
             $.ajax({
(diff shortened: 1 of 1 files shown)


install/run.php
```diff
@@ -297,7 +297,7 @@ function init_read_log() {
 
         return new Promise((resolve, reject) => {
             $.ajax({
-                url: "<?php echo $_POST['websiteurl'] ?? $websiteurl; ?>" + "index.php/update/dxcc",
+                url: url: `${window.location.origin}/index.php/update/dxcc`,
```
HB9HIL, line 300: copy copy copy paaaste xD

HB9HIL, line 300: Remove that c+p mistake :)
```suggestion
                url: `${window.location.origin}/index.php/update/dxcc`,
```

HB9HIL (changes requested):
Tested in various configurations, and with reverse proxys. By that I've learned something new about Codeigniters redirect behaviour.. 

Remove that c+p mistake and we're good to go

HB9HIL (approved):

install/run.php
patrickrb: Doh, that's embarrassing. Nice catch!

PR #1450, opened 2025-01-03
Don't disable button if no file was chosen

int2001:
If no file was chosen at ADIF Upload, and User presses upload the button is disabled.
this one simply returns if no file was chosen, so user can try again.


AndreasK79 (approved):

phl0 (commented):
Good idea!

PR #1452, opened 2025-01-03
Fallback for DXCC data

HB9HIL:


patrickrb:
Looks good to me!

int2001 (approved):
tested:
- DNS Failure --> Works
- Connectionproblems --> Works

missing: Automatic workflow to update the gh-mirror from time to time

PR #1454, opened 2025-01-04
Remove old-stuff ("Custom-Stat")

int2001:
12ys ago someone started to build a "custom-stat-page" and never finished it.
this page used evil SQLs (count over the complete table instead limiting it for the user).

PR removes that old stuff


AndreasK79 (approved):

phl0 (approved):
"Hau wech die Scheiße" :)

PR #1462, opened 2025-01-05
Fix advanced search

phl0:
Was b0rken in https://github.com/wavelog/wavelog/pull/1412/commits/37e05649a6d0f4a16b9ba00f00b5d9922f6760a1 resp. https://github.com/wavelog/wavelog/pull/1412 where the JSON parameter was renamed from "search" to "method" for some reason.

### DIFF
--- a/application/controllers/Search.php
+++ b/application/controllers/Search.php
@@ -98,7 +98,7 @@ class Search extends CI_Controller {
 	}
 
 	function search_result() {
-        	$sstring = str_replace('Ø', "0", $this->input->post("method", TRUE) ?? '');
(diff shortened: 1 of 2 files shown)


phl0:
Also removed old? debug part of the code. Please verify @AndreasK79 

Screenshot from 2025-01-05 09-51-04

int2001 (approved):
huh. how did this happen? was i drunk?
good find.

ahhh. it wasn't obvious because in 99,999% JS already replaced the zero....

AndreasK79 (approved):

AndreasK79:
@phl0 can't remember, but no need for the testvar :)

phl0:
> @phl0 can't remember, but no need for the testvar :)

So lets see where it breaks :D

PR #1465, opened 2025-01-06
Skip LoTW-Input for records without a station_callsign

int2001:
if a user tries to upload a LoTW-Report which doesn't include the station_callsign, skip this record.


phl0 (approved):
#fencestotheuser

Works. Though I would like to know how $user achieved that :)

int2001:
that'd be interesting for me also

PR #1467, opened 2025-01-06
Keep SAT frequencies on mode change

phl0:
If we post log SAT QSOs and choose SAT and SAT_MODE and then change mode from SSB to CW the default SAT frequencies are overwritten by the default CW frequencies of the band plan. This should not happen if we have selected a satellite before (and use those frequencies). Demo:

https://github.com/user-attachments/assets/2fc39c24-ec65-4c1b-b162-a2d4fe3bc115

Fix: Trigger mode changing of frequencies only if SAT is not set aka. empty.

### DIFF
--- a/assets/js/sections/qso.js
+++ b/assets/js/sections/qso.js
@@ -882,7 +882,7 @@ $('#start_date').on('change', function () {
 
 /* on mode change */
 $('.mode').on('change', function () {
-	if ($('#radio').val() == 0) {
+	if ($('#radio').val() == 0 && $('#sat_name').val() == '') {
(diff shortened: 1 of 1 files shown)


int2001:
TX-TRG stays as it was after this patch.
But RX is still overwritten. intended?

phl0:
> But RX is still overwritten. intended?

Nope. Just forgot to look at RX frequency after patch. Investigating :)

phl0:
Fix for RX freq pushed.

int2001 (approved):
looks good for me

phl0:
> looks good for me

tnx fer testing.

PR #1468, opened 2025-01-06
[Advanced Logbook] Added antenna azimuth and elevation columns

AndreasK79:
Added columns to display antenna azimuth and elevation.


phl0 (approved):
Nice addition for SAT ops. Works as designed.

PR #1472, opened 2025-01-07
show no clubs available only for admin

HB9HIL:
Show the info, that no clubstations are available in the header menu only for admins


int2001 (approved):
works

PR #1473, opened 2025-01-07
remove club permissions on conversion

HB9HIL:
fixes a bug where club permissions weren't removed after user conversion


int2001 (approved):
wks. tnx!

PR #1474, opened 2025-01-07
allow searching for first and lastname at the same time at club permissions

HB9HIL:
May an admin want to search for "Firstname Lastname"

this allows a search like this


phl0 (approved):
Works. Good stuff.

PR #1476, opened 2025-01-07
[Advanced Logbook] Added Clublog status in batch edit

AndreasK79:


int2001 (approved):
tested and works. good one

phl0 (approved):
Tested. Works

PR #1479, opened 2025-01-08
[Adif Export] Added info that the advanced logbook can be used for export as well

AndreasK79:
image

Also removed an unused function in the controller. (Adif/Main.php view doesn't exist either)


phl0 (approved):
Works fine

PR #1480, opened 2025-01-08
add .user.ini to .gitignore

DB4SCW:
Some hosters require a ".user.ini" file to store differing php settings usually found in .htaccess like memory_limit or max_execution_time.

To prevent this file from being overwritten on update or creating error messages, this should be included in .gitignore.
There is a possibility for each individual user to add this to the local ".git/info/exclude" file on every install to get the same result, but I think the global .gitignore makes more sense.

I can't think of any blockers, but if someone can think of some, happy to discuss and change.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,4 @@ sync.sh
 .maintenance
 .demo
 .htaccess
+.user.ini


HB9HIL (approved):

phl0:
> To prevent this file from being overwritten on update or creating error messages, this should be included in .gitignore. 

Could you please outline where this file would be overwritten? It is an untracked file (as not part of Wavelog). So I guess it cannot cause error messages as well. Pls shed some light 😇

DB4SCW:
If you run git pull it could potentially output "you have untracked files, do a git stash, etc." which is fine for the more technically inclined users, but could irritate other folk. Not really error messages, you are right.

phl0:
> If you run git pull it could potentially output "you have untracked files, do a git stash, etc." 

This message appears if you have modified files of the git repo which are also modified upstream. But this message will not appear if you just have an untracked file(s) in the directory. I am still trying to understand the problem ...

DB4SCW:
Yeah you are right, git pull will only act up if a .user.ini where to be integrated into upstream. You are completely right, this is not 100% necessary.
But maybe we could see this as a purely prophylactic measure to make sure we'll never trigger that in the future.
Happy to close it without merging if you like.

phl0:
> Yeah you are right, git pull will only act up if a .user.ini were to be integrated into upstream. You are completely right, this is not 100% necessary. But maybe we could see this as a purely prophylactic measure to make sure we'll never trigger that in the future. Happy to close the PR without merging if you like.

Na, we are going to merge it. We just don't want to spill the .gitignore with lots of filenames not being part of our repo :)

PR #1481, opened 2025-01-08
Dashboard improvements

filipmelik:
Implementation of a behavior explained in this [discussion](https://github.com/wavelog/wavelog/discussions/1410) except the user-configurable QSO count.

I have also updated QSO widget, so the number of QSOs are configurable using URL. 
Plus some minor cleanup.

NEED TO UPDATE WIKI ON MERGE TO MASTER

### DIFF
--- a/application/controllers/Dashboard.php
+++ b/application/controllers/Dashboard.php
@@ -3,6 +3,8 @@
 class Dashboard extends CI_Controller
 {
 
+	const LAST_QSOS_COUNT = 18; // max number of most recent qsos to be displayed on a dashboard
+
 	public function index()
 	{
 		// Check if users logged in
@@ -115,7 +117,10 @@ class Dashboard extends CI_Controller
 		$data['qrz_sent_today'] = $QSLStatsBreakdownArray['QRZ_Sent_today'];
 		$data['qrz_rcvd_today'] = $QSLStatsBreakdownArray['QRZ_Received_today'];
 
-		$data['last_qsos_list'] = $this->logbook_model->get_last_qsos('18', $logbooks_locations_array);
+		$data['last_qsos_list'] = $this->logbook_model->get_last_qsos(
+			Dashboard::LAST_QSOS_COUNT, 
+			$logbooks_locations_array,
+		);
 
 		$data['vucc'] = $this->vucc->fetchVuccSummary();
 		$data['vuccSAT'] = $this->vucc->fetchVuccSummary('SAT');
--- a/application/controllers/Dashboard.php
+++ b/application/controllers/Dashboard.php
@@ -1,7 +1,6 @@
 <?php if (!defined('BASEPATH')) exit('No direct script access allowed');
 
-class Dashboard extends CI_Controller
-{
+class Dashboard extends CI_Controller {
 
 	const LAST_QSOS_COUNT = 18; // max number of most recent qsos to be displayed on a dashboard
 
--- a/application/controllers/Dashboard.php
+++ b/application/controllers/Dashboard.php
@@ -115,7 +115,7 @@ class Dashboard extends CI_Controller
 		$data['qrz_sent_today'] = $QSLStatsBreakdownArray['QRZ_Sent_today'];
 		$data['qrz_rcvd_today'] = $QSLStatsBreakdownArray['QRZ_Received_today'];
 
-		$data['last_five_qsos'] = $this->logbook_model->get_last_qsos('18', $logbooks_locations_array);
+		$data['last_qsos_list'] = $this->logbook_model->get_last_qsos('18', $logbooks_locations_array);
 
 		$data['vucc'] = $this->vucc->fetchVuccSummary();
 		$data['vuccSAT'] = $this->vucc->fetchVuccSummary('SAT');
--- a/application/controllers/Dashboard.php
+++ b/application/controllers/Dashboard.php
@@ -2,7 +2,8 @@
 
 class Dashboard extends CI_Controller {
 
-	const LAST_QSOS_COUNT = 18; // max number of most recent qsos to be displayed on a dashboard
+	const DEFAULT_QSOS_COUNT = 20;
+	const MAX_QSOS_COUNT_LIMIT = 50; 
 
 	public function index()
 	{
@@ -116,10 +117,20 @@ class Dashboard extends CI_Controller {
 		$data['qrz_sent_today'] = $QSLStatsBreakdownArray['QRZ_Sent_today'];
 		$data['qrz_rcvd_today'] = $QSLStatsBreakdownArray['QRZ_Received_today'];
 
-		$data['last_qsos_list'] = $this->logbook_model->get_last_qsos(
-			Dashboard::LAST_QSOS_COUNT, 
-			$logbooks_locations_array,
-		);
+		// Determine last (recent) QSO count to be fetched
+		$last_qso_count_opt = $this->user_options_model->get_options(
+			'dashboard', 
+			array('option_name' => 'last_qso_count', 'option_key' => 'count'), 
+			$this->uri->segment(3)
+		)->result();
+		if (count($last_qso_count_opt) > 0) {
+			// value found in user options - use it
+			$last_qso_count = $last_qso_count_opt[0]->option_value;
+		} else {
+			// value not found in user options - use default value
+			$last_qso_count = \Dashboard::DEFAULT_QSOS_COUNT;
+		}
+		$data['last_qsos_list'] = $this->logbook_model->get_last_qsos($last_qso_count, $logbooks_locations_array);
 
 		$data['vucc'] = $this->vucc->fetchVuccSummary();
 		$data['vuccSAT'] = $this->vucc->fetchVuccSummary('SAT');
--- a/application/controllers/Qso.php
+++ b/application/controllers/Qso.php
@@ -1,7 +1,6 @@
 <?php if ( ! defined('BASEPATH')) exit('No direct script access allowed');
 
-class QSO extends CI_Controller 
-{
+class QSO extends CI_Controller {
 
     const LAST_QSOS_COUNT = 5; // max number of most recent qsos to be displayed on a qso page
 
--- a/application/controllers/Qso.php
+++ b/application/controllers/Qso.php
@@ -1,6 +1,9 @@
 <?php if ( ! defined('BASEPATH')) exit('No direct script access allowed');
 
-class QSO extends CI_Controller {
+class QSO extends CI_Controller 
+{
+
+    const LAST_QSOS_COUNT = 5; // max number of most recent qsos to be displayed on a qso page
 
 	function __construct()
 	{
@@ -34,7 +37,7 @@ class QSO extends CI_Controller {
 		$data['stations'] = $this->stations->all_of_user();
 		$data['radios'] = $this->cat->radios(true);
 		$data['radio_last_updated'] = $this->cat->last_updated()->row();
-		$data['query'] = $this->logbook_model->last_custom('5');
+		$data['query'] = $this->logbook_model->last_custom(self::LAST_QSOS_COUNT);
 		$data['dxcc'] = $this->logbook_model->fetchDxcc();
 		$data['iota'] = $this->logbook_model->fetchIota();
 		$data['modes'] = $this->modes->active();
@@ -601,7 +604,7 @@ class QSO extends CI_Controller {
       $this->load->model('logbook_model');
       session_write_close();
 
-      $data['query'] = $this->logbook_model->last_custom('5');
+      $data['query'] = $this->logbook_model->last_custom(self::LAST_QSOS_COUNT);
 
       // Load view
       $this->load->view('qso/components/previous_contacts', $data);
--- a/application/controllers/Widgets.php
+++ b/application/controllers/Widgets.php
@@ -40,7 +40,7 @@ class Widgets extends CI_Controller {
 				show_404(__("Unknown Public Page."));
 			}
 
-			$data['last_five_qsos'] = $this->logbook_model->get_last_qsos(15, $logbooks_locations_array);
+			$data['last_qsos_list'] = $this->logbook_model->get_last_qsos(15, $logbooks_locations_array);
 			
 			$this->load->view('widgets/qsos', $data);
 		}
--- a/application/controllers/Widgets.php
+++ b/application/controllers/Widgets.php
@@ -9,6 +9,7 @@
 class Widgets extends CI_Controller {
 
 	const LAST_QSOS_COUNT = 15; // number of of qsos to be displayed in the last qsos widget
+	const LAST_QSOS_MAX_LIMIT = 40; // if user requests more than this limit, qsos will be capped to this number 
 
 	public function index()
 	{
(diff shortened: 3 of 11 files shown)


phl0 (approved):
Tested and works fine on my end.

int2001:
works. but there's no difference in the behaviour.
Is there a reason for splitting the feature and not doing it in one PR?

application/controllers/Dashboard.php
```diff
@@ -3,6 +3,8 @@
 class Dashboard extends CI_Controller
 {
 
+	const LAST_QSOS_COUNT = 18; // max number of most recent qsos to be displayed on a dashboard
+
 	public function index()
```
int2001, line 8: Minor one. Pse don't mix the code-intenting here.
>90% of our code is implemented with
`function xyz() {`
and not with
```
function xyz()
{
```

application/controllers/Widgets.php
```diff
@@ -40,7 +40,7 @@ public function qsos($logbook_slug = null) {
 				show_404(__("Unknown Public Page."));
 			}
 
-			$data['last_five_qsos'] = $this->logbook_model->get_last_qsos(15, $logbooks_locations_array);
+			$data['last_qsos_list'] = $this->logbook_model->get_last_qsos(15, $logbooks_locations_array);
```
int2001, line 43: Y not using a variable here as well and pin it (as before) to 15?

application/controllers/Dashboard.php
filipmelik: I am not sure I follow here. The line you are referencing here was not changed by me at all. Could you pls elaborate what do you mean? Or did you try to say that if I find the "bad" code style, I should correct it? 🤔

application/controllers/Widgets.php
filipmelik: I was not paying too much attention here to the rest of the code, just to the key name. Will do.

filipmelik:
> works. but there's no difference in the behaviour. Is there a reason for splitting the feature and not doing it in one PR?
Well, the thing is that nobody commented on the linked discussion, so I was not sure if you, maintainers, are ok with this idea. And I do not want to put big effort into something that might not be used in the end, so I was waiting some kind of "formal" approval there 😄  So I started with the small thing, that I believed could do no harm and can be merged even without this approval 😊  But since you are asking now where is the rest of the code, I will also change the map to plot the QSOs from the table below, OK?

application/controllers/Widgets.php
int2001: you're right. not here, but at qso.php

int2001:
would be interesting, what you're planning about configuring the amount of QSOs

filipmelik:
> would be interesting, what you're planning about configuring the amount of QSOs

I am not 100% sure about this one tbh - if implemented, it is another field that can be configured by user. And I do not think that all must be neccesarilly configurable 😊  It was just an idea to discuss. What do you guys think about this being a field configurable somewhere in settings? It would also complicate the code a little, so there are pros/cons, as always :)  am actually OK with the current value of 18 QSOs, I was just trying to find out what you guys think about this 😊

int2001:
since this one was requested several times, i'd vote for a user-setting.

More opinions @AndreasK79 / @phl0 / @HB9HIL ?

phl0:
> More opinions @AndreasK79 / @phl0 / @HB9HIL ?

Agree

HB9HIL:
> since this one was requested several times, i'd vote for a user-setting.

I'm not against an user-option.  The question I currently asking myself is, do we need to limit it? 

The inital limit was because of space which is used if there are a lot of previous QSOs. Wouldn't it be easier and more "intuitive" if we just show all previous QSOs in a box with scrollable content? Just an idea which came to my mind. So we could save the extra code for a user-option and just display every previous QSOs with the same callsign. And as mentioned without using to much space in a box with scrollable content. 

If you more want a user-option I'm also fine with it.

int2001:
Absolutely against a "no-limit". 

Reason: Without a Call entered it would show the full log (too much! - performance!). Currently it shows the last 5 contacts, if no call is entered at the logging page.

HB9HIL:
> Absolutely against a "no-limit".
> 
> Reason: Without a Call entered it would show the full log (too much! - performance!). Currently it shows the last 5 contacts, if no call is entered at the logging page.
> 
> And even if we only show the 5 ones (without a call prepared to log): There are OPs who work some stations every day. So list can grow until it explodes.

Don't forget to limit the user option then ;-)

int2001:
Those were the reasons (so far) we didn't implemented a variable limit ;)
too many side-effects where one has to take care of.

application/controllers/Widgets.php
filipmelik: fixed

application/controllers/Qso.php
```diff
@@ -1,6 +1,9 @@
 <?php if ( ! defined('BASEPATH')) exit('No direct script access allowed');
 
-class QSO extends CI_Controller {
+class QSO extends CI_Controller
```
int2001, line 3: i meant this intention.
pse collapse to one row

filipmelik: Ahh, I understand now :) You were referring to `function` in your previous comment so I was confused. 
I have originally changed it, because I had seen same codestyle in `Dashboard.php`, but apparently, that was a wrong example to take :) Fixed

filipmelik:
So I tried to look into it and I think I will need some guidance here, as I am not sure how the settings system works. 

1) @int2001 suggested that I should use `user-option` for that, which means that this value will be stored in `user_options` table. Correct?
2) Where should I put this setting to be configurable? I looked into `views/user/edit.php` and at first glance this seemed like the place. However, after examining how where the values presented ther eare stored, it looks like they are pulled from table `users` and not `user_options`. If the value should be in `users`, it would need DB migration, as this would add a new column, right? What is the correct approach here please?
3) How about the case of default value, when user has has fresh install/not touched the value yet? Assuming the value will be in `user_options`, how does it get there in this case? Or should I use some "hardcoded default" somewhere in the code for this case?

Sorry for so many questions :)

int2001:
> 1. @int2001 suggested that I should use `user-option` for that, which means that this value will be stored in `user_options` table. Correct?

Yes, correct.

> 2. Where should I put this setting to be configurable? I looked into `views/user/edit.php` and at first glance this seemed like the place. However, after examining how where the values presented ther eare stored, it looks like they are pulled from table `users` and not `user_options`. If the value should be in `users`, it would need DB migration, as this would add a new column, right? What is the correct approach here please?

Half correct. First one: Yes `views/user/edit.php` is correct. Second one: storing user-attributes directly into the users-table is deprecated. it should(!) only hold the essentials values. Everything else is stored in user_options. To achieve that see - e.g. - the hams.at-settings - they're stored here: https://github.com/wavelog/wavelog/blob/e88a5ac220266b46c04b8d0428859b9ff95c188b/application/models/User_model.php#L291 

One of the main-reasons for introducing the user_options-table was: No need for a migration, no growing the amount cols at user_table any longer. It is already used by a lot of settings.

There's also a complete model to handle the user_options. See https://github.com/wavelog/wavelog/blob/e88a5ac220266b46c04b8d0428859b9ff95c188b/application/models/User_options_model.php

> 3. How about the case of default value, when user has has fresh install/not touched the value yet? Assuming the value will be in `user_options`, how does it get there in this case? Or should I use some "hardcoded default" somewhere in the code for this case?

absolutely right. this should be done at user_create. (See add_user functions at controller/model)
Please add some logic, to limit the amount. for example: 100?
The User shouldn't be able to input a number higher than this limit. Backend should check as well (Paradigm: "Never trust Userinput")

> Sorry for so many questions :)

You're welcome. i hope it helps a little. If you've further questions, don't hesitate to ask them ;)

filipmelik:
I have added 2 more commits. First one is hopefully straightforward, the second one is something I need help with.
There are 2 issues I do not know how to deal with:

1. **Where to put constants**, so they are widely available, across controllers and views? Specifically, I need 2 constants in code, first one being hard limit in case somebody is trying to change the number of QSOs displayed on dashboard in the "save settings" request to some absurd value (aka do not trust the user input). Second one being Default number of QSO's displayed, in case user is upgrading from previous version. @int2001 suggested that the default value should be put into `user_create`, which makes sense, but that won't cut it in case of upgrade from previous version where this function is not called (if I looked up correctly) and thus it will not be available in the user settings table.
2.Due to missing default value outlined in bullet 1, I had to do this code duplication (marked with comments in last commit), which is ugly and definitely not DRY. It works, but I do not think this is the way. I must be missing something here, do you see any other way how to do that?

Thanks for help

int2001:
> 1. **Where to put constants**, so they are widely available, across controllers and views? Specifically, I need 2 constants in code, first one being hard limit in case somebody is trying to change the number of QSOs displayed on dashboard in the "save settings" request to some absurd value (aka do not trust the user input). Second one being Default number of QSO's displayed, in case user is upgrading from previous version. @int2001 suggested that the default value should be put into `user_create`, which makes sense, but that won't cut it in case of upgrade from previous version where this function is not called (if I looked up correctly) and thus it will not be available in the user settings table.

ok. since it is now (or soon) configurable by user, pse simply hardcode the old limit (5 QSOs) at the Model. So Application will behave like before if nothing is changed. Best place to put it, is where you get the options. "Not set ==> Hardcoded default value"

The ultimate limit (no value above that possbile) could also be hardcoded - i'd do that at the controller while collecting the parameters. See "form_validation". If you like you can add it to config.php. Please consider a fallback, if the config.php-item wasn't set. And yes: that one should be hardcoded as well. i'd suggest 50 QSOs (which is very very much)

>    2.Due to missing default value outlined in bullet 1, I had to do this **code duplication** (marked with comments in last commit), which is ugly and definitely not DRY. It works, but I do not think this is the way. I must be missing something here, do you see any other way how to do that?

you're right. duplication should be minimized. imho it is possible to hardcode only both values (default-limit / ultimate-limit) once with the described way.

int2001:
GM.
This looks very good. Tnx for investing the effort.

i tested:
- Settingpage: Works
- Dashboard: Works
- Last-QSOs at qso-form: Setted value seems to have no effect?!
- QSO-Widget: Setted value seems to have no effect?!

Perhaps i am doing something wrong. Did not read the entire code, yet.

filipmelik:
GM Joerg, 
you summarized it well - this is exactly the behavior I was seeking for.

The main effort around this PR was to make **Dashboard** QSO count configurable and making sure, that map displays the number of QSOs configured. Hence The Settings Page + Dashboard works.

The [Last-QSOs at qso-form]((https://github.com/wavelog/wavelog/pull/1481/commits/e6fcd611b4719e4cafec7ce756a012e515da14b1)) and [QSO-Widget](https://github.com/wavelog/wavelog/pull/1481/commits/9b55aab3e2c202de15ac100d9fc54e3c03850a75) - The effort here was just making sure, that both places that deals with the last QSO count value will use common constant, to prevent situations that in case the value ever changes, the author will not forget updating value in both places, since up until now, it was hardcoded on those places (note: QSO widget value was in the end changed so the number of QSO's are [configurable via parameter](https://github.com/wavelog/wavelog/pull/1481/commits/95a595aa66b90a0032b5bf6dfcc434cc668e0252)).

To summarize:
- you are doing nothing wrong :)
- Dashboard QSO count is configurable now
- Widget QSO count is configurable now
- in case somebody will need making the Last-QSO count at qso-form configurable, it would be better to do it in separate PR anyways imho

int2001:
ok. misunderstood. i thought the whole time it was about the last qsos at QSO-Form ;)
however. i was confused about the adjustments at QSO-Form and Widget without any effect. You explained the reasons...

tbh: i am unsure about the current PR.
On one hand you added configure for last-qsos at settings and that they're taking effect at the dashboard.
on the other i see upcoming requests/confusions regarding "But my last-qsos at QSO-Form behave the same as before".

really unsure.

need opinion from @phl0, @AndreasK79 and @HB9HIL here.

Regarding Code/PR (in your described scope) i'd approve it - well done.

filipmelik:
>  i see upcoming requests/confusions regarding "But my last-qsos at QSO-Form behave the same as before"

I see, however, the setting I have added looks exactly like on the screenshot below. The section is called `Dashboard` and the hint text also describes that this is intended to change behavior only on the dashboard page. I tried to make it crystal clear that this is just dashboard thing. So if somebody will complain that QSO list in QSO page behaves the old way, then he perhaps did not read carefully what the setting really is about 😉  
 
Screenshot 2025-01-28 at 7 17 45

phl0 (approved):
I am fine with the current state of the PR. It works fine on my end. The setting clearly says it affects the dashboard. However if you think about adding a separate setting for "Previous QSOs" I'd vote for a separate setting. I like to have 50 QSOs on the dashboard. But I guess 50 would be too much for the previous list and would move the DXCC summary too far down.

phl0:
> tbh: i am unsure about the current PR. On one hand you added configure for last-qsos at settings and that they're taking effect at the dashboard. on the other i see upcoming requests/confusions regarding "But my last-qsos at QSO-Form behave the same as before".

> need opinion from @phl0, @AndreasK79 and @HB9HIL here.

My view: Can be added in a separate PR. I would vote for a separate setting to make the previsous QSOs list independent from the dashboard.

int2001 (approved):

HB9HIL:
@filipmelik we reverted this PR in #1568 .. Sorry for that but I found something which is not okay

image

1. You now include the Dashboard Controller every time the user model is loaded. This is not only bad practice as you start mixing the MVC pattern, also this is an totally unnecessary load.

2. You are using a Class Loading /Dashboard which is not used in the entire codebase. It would be much better to stick with the overall code style to keep maintenance level low and the code style consistent. Please reopen the PR and we will fix that together

filipmelik:
I have somehow fucked up my git and the commits are gone, lol, I will need some more time to reopen. However, I guess in the meantime, could you point me towards a way how would you like me to implement it? 

My ultimate goal was to use one constant defined in one place. Since the constant is related to setting on Dashboard, I have put it to Dashboard.php. I do not really care if the constant will live elsewhere, but I did not find any such place where this could be defined (but bear with me - I do not know CodeIgniter framework at all). I really want to avoid hardcoding the same value on 4 different places 😊  So, any pointers please how to proceed here?

HB9HIL:
> I have somehow fucked up my git and the commits are gone, lol, I will need some more time to reopen. However, I guess in the meantime, could you point me towards a way how would you like me to implement it?
> 
> My ultimate goal was to use one constant defined in one place. Since the constant is related to setting on Dashboard, I have put it to Dashboard.php. I do not really care if the constant will live elsewhere, but I did not find any such place where this could be defined (but bear with me - I do not know CodeIgniter framework at all). I really want to avoid hardcoding the same value on 4 different places 😊 So, any pointers please how to proceed here?

The point is: We don't use constants anywhere. Even if we would it would be really bad to load a controller file inside a big model. 

The best way to solve that - which is also the way we usually solve these things - is to store it in the session. 

You already store it in the user options table. which is awesome. Now check out the `update_session()` method in theUser_model. Here all relevant data get written into the php session. Sure you could also load the value from the database but using the session is much better in terms of overall performance by reducing the amount of db querys. 

Always stick to the MVC pattern. Controller controls and model handles data (session or database). 

Sorry for this inconvience. I really don't want to be the bad cop here but including a controller in the user model to use one constant is definitly not the way we should go.

PR #1482, opened 2025-01-08
Make DARC DOK and other refs selectize input uppercase

phl0:
As DARC DOK value is upper case in any case we should make the input uppercase as well. (CSS text-transform only applies to the shown value).

Before:
image

After:
image

Drawback is that "ADD" is now also shown uppercase but I guess this is a minor issue :)


filipmelik:
Nice one! Could we also add this behavior to WWFF, POTA and maybe to the rest of the relevant fields too? I was having this idea in my pipeline too and wanted to take a look at that at some point 😊

phl0:
For sure. Should be easy. Will do

int2001 (approved):

phl0:
Added uppercase-handling for other refs as well.

int2001 (approved):

PR #1483, opened 2025-01-08
Implement contest srx/stx and comments in SimpleFLE

DJ3CE:
DF3CB has specified `,***` and `.***` for contest exchange-received and sent, as well as `<...>` for comments.

Comments may contain spaces and are handled ahead of splitting rows by spaces.

Additionally, you may enter an arbitrary ADIF-field(\*) in the form of '<tx_pwr: 50>' or '<notes: Uses a straight key>', as long as the key matches '[a-z_]*:' and is implemented in wavelog's 'import()' in the Logbook-Model.

I'd be happy, if you'd a look into, do some tests and give some feedback :). Thanks!

### DIFF
--- a/assets/js/sections/simplefle.js
+++ b/assets/js/sections/simplefle.js
@@ -470,8 +470,8 @@ function handleInput() {
 			<td>${callsign}</td>
 			<td><span data-bs-toggle="tooltip" data-placement="left" title="${freq}">${band}</span></td>
 			<td>${mode}</td>
-			<td>${rst_s}${stx_info}</td>
-			<td>${rst_r}${srx_info}</td>
+			<td>${rst_s} ${stx_info}</td>
+			<td>${rst_r} ${srx_info}</td>
 			<td>${gridsquare}</td>
 			<td>${sotaWwffText}</td>
 			</tr>`);
--- a/assets/js/sections/simplefle.js
+++ b/assets/js/sections/simplefle.js
@@ -242,15 +242,17 @@ function handleInput() {
 		var add_info = {};
 
 		// First, search for <...>-Patterns, which may contain comments (... or additional fields)
-		let addInfoMatches = row.matchAll(/<([^>]*)>/g);
+		let addInfoMatches = row.matchAll(/<([^>]*)>|\[([^\]]*)\]/g);
 		addInfoMatches.forEach((item) => {
-		  row = row.replace(item[0], "");
-		  let kv;
-		  if (kv = item[1].match(/^([a-z_]+): *(.*)$/)) {
-		    add_info[kv[1]] = kv[2];
-		  } else {
-		    add_info.comment = (('comment' in add_info)?add_info.comment+' ': '')+item[1];
-		  }
+			row = row.replace(item[0], "");
+			let kv;
+			if (item[0][0] == '<' && (kv = item[1].match(/^([a-z_]+): *(.*)$/))) {
+				add_info[kv[1]] = kv[2];
+			} else if (item[0][0] == '[') {
+				add_info.qslmsg = item[2];
+			} else {
+				add_info.comment = (('comment' in add_info)?add_info.comment+' ': '')+item[1];
+			}
 		});
 
 		// Now split the remaining line by spaces and match patterns on those
@@ -315,9 +317,9 @@ function handleInput() {
 				callsign = item.toUpperCase();
 				call_rec = true;
 			} else if (
-				item.match(/^[A-R]{2}[0-9]{2}([A-X]{2}([0-9]{2}([A-X]{2})?)?)?$/i)
+				parts = item.match(/(?<=^#?)[A-R]{2}[0-9]{2}([A-X]{2}([0-9]{2}([A-X]{2})?)?)?$/i)
 			) {
-				gridsquare = item.toUpperCase();
+				gridsquare = parts[0].toUpperCase();
 			} else if (itemNumber > 0 && item.match(/^[-+]\d{1,2}$|^\d{1,3}$|^\d{1,3}[-+]d{1,2}$/)) {
 				if (rst_s === null) {
 					rst_s = item;
@@ -339,6 +341,8 @@ function handleInput() {
 						srx = parts[5];
 					}
 				}
+			} else if (itemNumber > 0 && (parts = item.match(/(?<=^@)[A-Za-z]+/))) {
+				add_info.name = parts[0];
 			}
 
 			itemNumber = itemNumber + 1;
--- a/assets/js/sections/simplefle.js
+++ b/assets/js/sections/simplefle.js
@@ -224,6 +224,8 @@ function handleInput() {
 	var callsign = "";
 	var gridsquare = "";
 	var sotaWwff = "";
+	var srx = "";
+	var stx = "";
 	qsoList = [];
 	$("#qsoTable tbody").empty();
 	errors = [];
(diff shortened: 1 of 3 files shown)


HB9HIL (commented):
image

Hello @DJ3CE 

Thanks for your effort. I just started the review and the first thing I noticed is the UI. When entering contest serial numbers the data is shown in the refs column which is usually for xOTA references. We may have to find another solution here. Just thinking loud: One possibility would be to change that column to "Data" and place an icon in there which shows a Tooltip with additional data. If no data is there it probably should be hidden. 

Nothing crucial but UI/UX is also important here. The functionality of entering Contest Serial Numbers works! Continue testing...

HB9HIL:
Another one:

```
20m ssb
2304 hb9hjq jn46ri ,47 .37
5 hb9hgg .38
6 hb9hab .372
```

This results in:

image


Imho the sent serial number should either increase or be hidden... But not the same if not provided. Maybe increase?

HB9HIL:
Another one:

If "No Contest" is selected in SFLE, the Contest should set to `Other` since an empty contest with serial numbers seems strange

DJ3CE:
Thanks for your notes and for looking into it.

I'd go with the information provided by the user, so I'd be fine with an empty 'contest' field. Seems to be the most predictable behavior for the user, but I have no strong opinion on that. How about a notice box or an notice on uploading? Not sure if I'm the right one to implement the latter, but :woman_shrugging: ;) .

Sent contest exchange: There are contests (YOTA, e.g.) where you send a constant exchange, so it is convenient to keep it, like we keep band and mode. You may increment easily by suffixing the received exchange by ',' and you may erase sent exchange with a single ', ' (read 'set it to empty').

If these arguments don't catch we could also go with ', ' for keeping the previous stx  (dropping otherwise), and ',+' for incrementing previous stx.

Thinking about transferring a contest, the implemented way seems easier and shorter, not sacrificing intuitivity too much - at least to me ;).

I second the large room for improvement in the UI: How about adding stx/srx to the reports, as it is done in the log-views anyway? Would be nice to have a tooltip showing all information taken from the input, additionally.

Other UI ideas?

Although FLE-documentation does not mention non-serial contest exchange I'd additionally implement that once we sorted out, how to proceed.

HB9HIL:
> I'd go with the information provided by the user, so I'd be fine with an empty 'contest' field. Seems to be the most predictable behavior for the user, but I have no strong opinion on that. How about a notice box or an notice on uploading? Not sure if I'm the right one to implement the latter, but 🤷‍♀️ ;) .

The Cabrillo Exporter is built to search in the Database for Contests. If this field is empty you can log as many SFLE QSOs as you want. You wouldn't be able to export them as cabrillo. Therefore the field can't be empty. The solition is `Other` as this fits any other contest which is available in Wavelog's database. 

> Sent contest exchange: There are contests (YOTA, e.g.) where you send a constant exchange, so it is convenient to keep it, like we keep band and mode. You may increment easily by suffixing the received exchange by ',' and you may erase sent exchange with a single ', ' (read 'set it to empty').

You implemented only a serial number as an exchange. In those contests you mention you usually sent always the same "exchange" which is in this contest the exchange string which is different from the serial number exchange. Can you refer to a contest where is always sent the same **serial number**? I'm pretty sure it's the exchange string which was not implemented here. (inserting `,test .abd2233` will not write this data into the database as only integer are allowed in `stx/srx`. 
Strings live in `stx_string/srx_string` which is the exchange (like your age, the cq zone you live in and so on)

Your supposed solution with 
`,` for keeping the last sent serial number and
`,+` increasing the last sent serial number by one 
seems a good idea to me. 

>[!NOTE]
> But in any case. In the documentation need to be clearly communicated that only serial NUMBERS are supported. There is no implementation for the "exchange" (string) yet

> I second the large room for improvement in the UI: How about adding stx/srx to the reports, as it is done in the log-views anyway? Would be nice to have a tooltip showing all information taken from the input, additionally.

I like this idea. You mean like this
image

DJ3CE:
> The Cabrillo Exporter is built to search in the Database for Contests. If this field is empty you can log as many SFLE Contest QSOs as you want. You wouldn't be able to export them as cabrillo. Therefore the field can't be empty. The solution is `Other` as this fits any other contest which is available in Wavelog's database.

In my opinion that's two different things: I may choose not to set the contest in first place (maybe I have taken part in different contests and want to sort that out later in the log). If we assume 'Other' we may mess up exports of what a user has already logged, in my opinion. As I said, no strong opinion here - and a notice would be be a really good thing anyway.

> Your supposed solution with `,` for keeping the last sent serial number and `,+` increasing the last sent serial number by one seems a good idea to me.

 :+1:

> 
> Note
> 
> But in any case. In the documentation need to be clearly communicated that only serial NUMBERS are supported. There is no implementation for the "exchange" (string) yet

I had already written about non-numeric exchange aka `stx_string` in my reply and you could already set `stx_string` with `<stx_string:Bob>` btw.
With the (parsed) QSO-data in the table it should be clear what happens while I agree that some additional documentation would be nice, especially since there are aspects not following the original FLE-documentation.
 
Regarding contest data in the table: Your screenshot shows exactly what I meant.

DJ3CE:
Some changes here, moving forward :):

* Now also non-numeric exchange possible,
* added documentation,
* incrementing `stx` now can be enabled by `,++` and disabled by `,+0` (closer to SimpleFLE documentation),
* aligning with the SimpleFLE principle only to note data, that has changed, `stx` and `stx_string` are applied from the previous QSO, if a received contest-exchange is detected,
* wiping `stx` and `stx_string` (together with previous `stx`/`stx_string`) can be done with `,-`,
* show contest-exchange beneath RST(R)/-(S), like in the log,
* checking and warning if contest-data-present and no-contest-selected resp. vice versa

I'd be happy for some testing and opinions/remarks/suggestions regarding the documentation :).

p.s. Already spotted a mistake in the documentation in the last `<pre>`-block (some parts missing), will fix in a second...

assets/js/sections/simplefle.js
```diff
@@ -363,14 +446,32 @@ function handleInput() {
...
 			<td>${extraQsoDate}</td>
 			<td>${qsotime}</td>
 			<td>${callsign}</td>
 			<td><span data-bs-toggle="tooltip" data-placement="left" title="${freq}">${band}</span></td>
 			<td>${mode}</td>
-			<td>${rst_s}</td>
-			<td>${rst_r}</td>
+			<td>${rst_s}${stx_info}</td>
+			<td>${rst_r}${srx_info}</td>
 			<td>${gridsquare}</td>
```
HB9HIL, line 475: Please add some spacing here
```suggestion
			<td>${rst_s} ${stx_info}</td>
			<td>${rst_r} ${srx_info}</td>
```

HB9HIL (changes requested):
Works pretty good now. Great improvement amd good work! Change the one suggestion and we are good to go

assets/js/sections/simplefle.js
HB9HIL: And you may want to update the SimpleFLE documentation when this PR is merged ;-) but its fine after merge. Would really appreciate it :)

https://github.com/wavelog/wavelog/wiki/SimpleFLE

HB9HIL (approved):
to make some progress here I added the spacing myself. :-)

Imho this is ready 2 merge

phl0 (approved):
Tested and works fine on my end.

PR #1484, opened 2025-01-09
fix reset_time button to actually reset the time

### LINKED ISSUE #1451: Reset QSO start time button sometimes not working
The 'Reset Start Time' button sometimes fails to work in manual QSO logging page (`/qso?manual=1`). 

If we set some random time and click the button, it won't reset the time.

chrome_Puc5OKPuTI

I think the bug was introduced in [1.8.8](https://github.com/wavelog/wavelog/commit/3ad31dffcd229023b535faff2958b13037877293#diff-15d104bd141cae4f18c96e6791c4f262d0e05450b17d6b5d239102914cc4986aR158), where we use `.attr('value', ...)` instead of `.val(...)`. 

Any thoughts?

patrickrb:
Resolves #1451

Fixes an issue where the reset time button wasn't working

# Before fix:

# After fix:

### DIFF
--- a/assets/js/sections/qso.js
+++ b/assets/js/sections/qso.js
@@ -154,26 +154,43 @@ $('#reset_time').on("click", function () {
 	});
 });
 
-$('#reset_start_time').on("click", function () {
-	var now = new Date();
-	$('#start_time').attr('value', ("0" + now.getUTCHours()).slice(-2) + ':' + ("0" + now.getUTCMinutes()).slice(-2));
-	$("[id='start_time']").each(function () {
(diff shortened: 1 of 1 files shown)


int2001:
Good finding.

But: the Reset-Button for the end-time seems to have the same bug.

Enable `Log End Times for QSOs Separately` at Account-Menu. Would be great if this one can be fixed within this PR, too.

int2001 (approved):
wks. tnx

phl0 (approved):
Checked and works

PR #1489, opened 2025-01-10
[QSO Edit] Added Clublog under QSL

AndreasK79:
Added Clublog under QSO edit:
image

Also added all statuses under LBA batch edit:
image


phl0 (approved):
Tested. Works :)

PR #1490, opened 2025-01-12
[DXCC Award] Made the contact popup more future safe

AndreasK79:
Don't know why I didn't fix this before, but it is a code going several years back.
Better to use the ADIF number for the DXCC than the name.


int2001 (approved):
vry good finding!

tested:
- DXCC-Award.
- DXCC-Ovierview at logging

works

PR #1492, opened 2025-01-12
[Advanced Logbook] Added flag and clickable callsign in map popup

AndreasK79:
image


int2001 (approved):
tested on: Mac and Windows (Chrome). Works and looks nice

PR #1495, opened 2025-01-13
Brute force protection

int2001:
Implemented a Login-Failure-Counter.

After 3 failed attempts no login is possible. User has to reset its credentials OR Admin needs to set new PW.

Testing:
- Login multiple times with wrong password
- Reset your PW via "forgot password" OR Reset the PW by setting a new one via User_edit.

Some more background info on this:
As WL is becoming more and more pupular, it's only a question of time for the 1st Brute-Force-Attack.
If the admin locks hisself out of the system, he can simply raise the `$config['max_login_attempts'] = 3;` to a higher value. we should document this at the wiki

### DIFF
--- a/application/controllers/User.php
+++ b/application/controllers/User.php
@@ -1055,7 +1055,7 @@ class User extends CI_Controller {
 				$this->session->set_flashdata('warning', __("You can't login to a clubstation directly. Use your personal account instead."));
 				redirect('user/login');
 			} else if ($login_attempt === 3) {
-				$this->session->set_flashdata('warning', __("Your account is locked, due to too many failed login-attempts. Please reset your Password."));
+				$this->session->set_flashdata('warning', __("Your account is locked, due to too many failed login-attempts. Please reset your password."));
 				redirect('user/login');
 			} else {
 				if(ENVIRONMENT == 'maintenance') {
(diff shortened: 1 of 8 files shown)


application/controllers/User.php
```diff
@@ -1023,6 +1023,9 @@ function login($firstlogin = false) {
 			} else if ($login_attempt === 2) {
 				$this->session->set_flashdata('warning', __("You can't login to a clubstation directly. Use your personal account instead."));
 				redirect('user/login');
+			} else if ($login_attempt === 3) {
```
ethancedwards8, line 1026: Could we change this to be configurable? IMO 3 could easily be reached pretty quickly just by fat fingering your password a couple times.

int2001: good point. thought also about it.
@HB9HIL wants to make a few adjustments as well. after that has been done i'll implement it with default "3", but configurable at config.php (No, @HB9HIL - no UI-Config 😜)

int2001: done

int2001:
tnx for adding the UI-Unlocker, @HB9HIL - everything seems to work now.
i'd appreciate if @AndreasK79 or @phl0 can check/approve this one as well

phl0:
> i'd appreciate if @AndreasK79 or @phl0 can check/approve this one as well

Will do.

AndreasK79 (approved):

AndreasK79:
Tested working with entering password wrong, then logging in as admin and resetting. That worked.

phl0 (approved):
Works (fine) for me.

PR #1497, opened 2025-01-14
[Advanced Logbook] Enhance sat edit and fix sat mode

AndreasK79:
The current batchedit overwrite SAT MODE when setting SAT NAME. Problem is that it should not be done when SAT MODE is blank. Now checks if value is blank before setting.

Also an enhancement for SAT is added so that you can set BAND TX And RX:
image

@phl0 for you to test and break ;)


phl0 (approved):
Nothing to break. Works fine here :)

phl0 (approved):
Works. We may need to take care of frequencies here as well though

PR #1498, opened 2025-01-14
Lotw update Fix

HB9HIL:
Fixing the LoTW Update logic which updates the QSO in the database. 

Explaination:
In the `import_check()` we check for matching QSOs within an 15 Minute range. The LoTW Update Logic does the same check again but this time without the 15 Minute range and expects a strikt match of the Time ON. 

Since the method has the QSO ID (COL_PRIMARY_KEY) available , it's not necessary to query for the other parameters aswell. So we can get rid of them.

Smaller Code but better results :)


int2001:
good finding.
imho it is okay. waiting for our LoTW-Expert ;)

phl0 (approved):

PR #1499, opened 2025-01-14
Add units and make sortable

phl0:
This adds units to az/el and station power and makes those columns sortable numerically (agn).


AndreasK79:
Tested working. A nice addition!

PR #1500, opened 2025-01-14
Fallback to short-living-popup if fetch fails (cors-workaround)

int2001:
See https://github.com/wavelog/wavelog/discussions/1487

Under normal circumstances QSY is working as before.
If one has WavelogGate or sth. different (Microcontroller) somewhere else than on local computer, this one falls back to open a popup (and close it instantly) to push the QSY-Trigger to the service.


phl0 (approved):
Works here. Maybe remove the debug output?

PR #1501, opened 2025-01-15
Fix for Last-QRZ-QSL

int2001:
before patch:
- if triggered by cron, every QSO will be fetched OR even worse: last-qsl stuck at first synced station/user

after patch:
- if triggered by cron the last received "QSL" from qrz will be taken for each user/sync


phl0 (approved):
works for me and fixes the bug.

PR #1509, opened 2025-01-16
Import and show incoming (e)QSL messages

phl0:
This adds importing the incoming QSL messages from eqsl and shows them on QSO view and in eQSL overview table.


AndreasK79:
The LBA part works. Can't really test the eQSL stuff.

int2001 (approved):
good teaming ;)

PR #1510, opened 2025-01-16
Dxcc update tweak

AndreasK79:
Tweaking the DXCC updater to use less time and resources.

### DIFF
--- a/application/controllers/Update.php
+++ b/application/controllers/Update.php
@@ -31,138 +31,73 @@ class Update extends CI_Controller {
 	}
 
 	/*
- * Load the DXCC entities
- */
-public function dxcc_entities($xml_data = null) {
-    // Ensure the Paths library is loaded
-    if (!$this->load->is_loaded('Paths')) {
-        $this->load->library('Paths');
-    }
-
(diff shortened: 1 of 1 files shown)


phl0 (approved):
Good to go imho

phl0 (changes requested):
Should compare before and after. Needs some fix on dates :-)

AndreasK79:
After some investigation, it turns out that the 1970 is in fact coming from the file:

```
 	<exception record='327'>
		<call>KG4W</call>
		<entity>UNITED STATES OF AMERICA</entity>
		<adif>291</adif>
		<cqz>5</cqz>
		<cont>NA</cont>
		<long>-77.54</long>
		<lat>37.69</lat>
		<start>1970-01-01T00:00:00+00:00</start>
		<end>2020-12-31T23:59:59+00:00</end>
	</exception>
```

The old code sets these as blank on start, so the code is updated to do the same.

phl0 (approved):
Latest fix fixes the start date correctly. Works fine imho

int2001 (approved):
great performance-hub. i like

PR #1511, opened 2025-01-17
60m was missing for GetBand

int2001:


AndreasK79 (approved):

phl0:
Good point. Intresting that this has not been raised earlier .. =)

PR #1518, opened 2025-01-19
[Callsign Statistics] Made a view to see how many times you worked a callsign

AndreasK79:
This was just an idea I got to see how many times a callsign was worked within the filters available.

### DIFF
--- a/application/controllers/Callstats.php
+++ b/application/controllers/Callstats.php
@@ -1,0 +1,122 @@
+<?php
+defined('BASEPATH') OR exit('No direct script access allowed');
+
+class Callstats extends CI_Controller
+{
+
+    function __construct()
+    {
+        parent::__construct();
+
+        $this->load->model('user_model');
+        if (!$this->user_model->authorize(2)) {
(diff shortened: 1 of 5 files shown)


int2001:
Nice feature. Played a bit with it. Problems i found:
- Once selected a SAT, the filter isn't working. Even if setting to a HF Band, SAT is still filtered. Reproduce: Select SAT/SO-50, filter. Go back to HF (SAT disappears) filter again.
- Not 100% sure about sanitizing. you use xss_clean at details and binding (Great!). At index you're collecting the data "plain". Didn't look into the view deeply if you use the values only for comparison there or for output as well. If used for output, pse sanitize. 
 
BTW: Shorthand for xss_clean while collecting is -e.g.: `$propagation = $this->input->post('propagation',true);` instead of `$propagation = $this->input->post('propagation');`

phl0 (approved):
Works for me after implementing the GEO/LEO/MEO stuff. @int2001 Would you please re-/cross-check? mni tnx

int2001 (approved):
yeah. Tested. MEO now possible as well

PR #1519, opened 2025-01-19
Fixes for Wamp server

AndreasK79:
While troubleshooting WAMP server, I found these 2 issues.

The migration is probably related to the MySQL config. I don't see any harm in changing it.

The tag in the version gives an error. Seems to be some issues with permission, but I don't see the harm in defaulting to unknown if latest version can't be grabbed from Github. This also made the debug page crash without it.

### DIFF
--- a/application/migrations/134_create_label_paper_types_table.php
+++ b/application/migrations/134_create_label_paper_types_table.php
@@ -21,7 +21,7 @@ class Migration_create_label_paper_types_table extends CI_Migration {
 
 				'paper_name' => array(
 					'type' => 'VARCHAR',
-					'constraint' => '250',
+					'constraint' => '191',
 				),
 
 				'metric' => array(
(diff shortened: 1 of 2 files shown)


int2001:
Nice feature. Played a bit with it. Problems i found:
- Once selected a SAT, the filter isn't working. Even if setting to a HF Band, SAT is still filtered. Reproduce: Select SAT/SO-50, filter. Go back to HF (SAT disappears) filter again.
- Not 100% sure about sanitizing. you use xss_clean at details and binding (Great!). At index you're collecting the data "plain". Didn't look into the view deeply if you use the values only for comparison there or for output as well. If used for output, pse sanitize. 
 
BTW: Shorthand for xss_clean while collecting is -e.g.: `$propagation = $this->input->post('propagation',true);` instead of `$propagation = $this->input->post('propagation');`

phl0:
Tried to test the migration but failed due to so much that would have to be turned back. Reading the code it should be ok however :)

AndreasK79:
@patrickrb one thing I didn't fix here, is that the migration and dxcc update does not work in the installer.
The problem is the url resolving in run.php. 

If I hardcoded the url, the migration worked. If I changed the url to the way it was before, the installer also worked.

I don't know why this is a problem in the WAMP server, but perhaps you have an idea?

PR #1520, opened 2025-01-19
Deactivate Clublog if auth-error

int2001:
If a user provided wrong-creds for Clublog, his sync will be disabled if the cronjob detects that.

at Realtimelogging there was no such mechanism. This one fixes that.

Important: Because clublog puts the whole instance on its blacklist otherwise


AndreasK79 (approved):

AndreasK79:
Can't test, but code looks ok.

phl0 (approved):
Works ok here in a few tests.

PR #1521, opened 2025-01-19
[Advanced Logbook] Fixing empty text field search

AndreasK79:
Seems my logic was a bit flawed regarding removing * to search for empty results.
This PR should fix that.


int2001 (approved):
works. good find

PR #1528, opened 2025-01-20
Fix "my_country" in ADIF export from advanced search

DJ3CE:
**Describe the bug**
Exporting ADIF from the "advanced search" leads to corrupted 'MY_COUNTRY'-data in the resulting ADIF-file.

**Version Information**
- OS: Linux
- PHP: PHP 7.4
- MySQL/MariaDB:  MariaDB 10.3
- Wavelog: 1.9.1 / current dev-Branch, 94821b7d0

**Steps to Reproduce**
1. Go to 'Advanced Search'
2. Click on 'Search' and 'Export to ADIF'
4. Open the ADIF
5. Look for the 'MY_COUNTRY'-Line of a QSO with some other country than yours

**Expected Behavior**
'MY_COUNTRY' should contain your (station's) country, aligning with 'MY_DXCC'.
Instead, 'MY_DXCC' shows the correct DXCC (in my case 230) and 'MY_COUNTRY' shows the QSO-partners country as well (i.e. uppercased form of 'COUNTRY').

**Additional Info**
This PR should fix the described bug.

As in Line 300 in `Search.php` the `dxcc_entities` column 'name' is explicitly aliased as 'station_country', the join of `dxcc_entities` in line 308 should reference `station_profile.station_dxcc`.

p.s. ... I took over the changes regarding the 0 here as well.

### DIFF
--- a/application/controllers/Search.php
+++ b/application/controllers/Search.php
@@ -104,7 +104,8 @@ class Search extends CI_Controller {
 	}
 
 	function export_to_adif() {
-		$data['qsos'] = $this->fetchQueryResult(($this->input->post('search', TRUE) ?? ''), FALSE);
+		$sstring = str_replace('Ø', "0", $this->input->post("search", TRUE) ?? '');
+		$data['qsos'] = $this->fetchQueryResult($sstring, FALSE);
 		$this->load->view('adif/data/exportall', $data);
 	}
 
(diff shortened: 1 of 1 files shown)


HB9HIL (approved):
nice find

phl0:
Cannot reproduce here for some reason though reading the code it looks ok. Any (special) prerequisites needed to provoke the bug?

phl0 (approved):
Sorry I used the wrong search. Fix confirmed. Good work 👍

AndreasK79:
@DJ3CE thanks for the fix! I know we had several of these bugs that were fixed other places in the code, but this one was obviously missed until now!

PR #1529, opened 2025-01-20
Show Update Results in Debug Page as Flash Messages

HB9HIL:
if there is a session available, we can show the results of the update buttons in the debug page as flash messages. This leads to a much better UX instead echoing out the plain result in a blank page. In CLI the results are still normally printed out. 

image


phl0 (approved):
Good idea. Works well here.

AndreasK79 (approved):

PR #1531, opened 2025-01-21
Make Old CSV-Export injection-safe (binding)

int2001:
CSV-Thing seems to be absolutely forgotten at refactoring to bind. this one fixes it.

Still draft, because there seems to be another issue with this one. (wink to @phl0 - be my guest, if you found sth.)


HB9HIL (approved):
not tested but code looks good

phl0 (approved):
Works. No other changes needed in this model imho.

PR #1533, opened 2025-01-21
[Installer] Fix if Wavelog is installed in a sub folder

AndreasK79:
Did some more digging into WAMP server, and found that the current running of the ajax call does not work since the url is in a sub folder.

By doing this change, it works in WAMP, however, I am not able to test all variants here, so I don't know if this is safe for all variants.

Xampp works.
Wamp works.
Docker does not work.

Not sure what to do here.

### DIFF
--- a/install/run.php
+++ b/install/run.php
@@ -28,7 +28,7 @@
                 </div>
 
                 <?php
-                // we can easily add more steps here if necessary 
+                // we can easily add more steps here if necessary
                 ?>
 
                 <div class="mb-3" id="installer_lock" style="opacity: 50%;">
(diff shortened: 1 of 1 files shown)


HB9HIL:
This is related to the change in #1444 by @patrickrb 

will dig into that

int2001 (changes requested):
just for protocol-reasons:

losing docker-users to make WAMP working is no option for me.
i try to find a solution together with you, @AndreasK79 - but i don't like the idea to disable docker/make docker disfunctional

AndreasK79:
> losing docker-users to make WAMP working is no option for me. i try to find a solution together with you, @AndreasK79 - but i don't like the idea to disable docker/make docker disfunctional

And that was never the point. I fixed Wamp, then tested Docker, and found out that the fix doesn't work there.

This is all related to folders.

The reason it failed in Wamp, was that the migration is called with a url like this: http://127.0.0.1/index.php/migrate, but it should be http://127.0.0.1/wavelog/index.php/migrate.

When I tested Docker, it failed because the migration was called with http://localhost:8086/install/index.php/migrate, and should be http://localhost:8060/index.php/migrate.

HB9HIL:
Got a replacement for this PR. See 1540

PR #1540, opened 2025-01-22
[Installer] Fixes

HB9HIL:
Replaces #1533 

Thanks @AndreasK79 for your debugging. But it seems that get rid of the configurable directory in the installer is the more safe way since you can change that stuff afterwards in the config.php. The installer needs to be bullet proof in the context of URL and directory. 

Please also test this stuff against WAMP / XAMPP

I tested docker and it works

@int2001 I also adressed the `.lock` issue in this PR which was already mentioned. The check is now extended to also check for config files and if there is something existing the visitor gets redirected (which was also broken) :D

### DIFF
--- a/application/config/config.sample.php
+++ b/application/config/config.sample.php
@@ -109,7 +109,7 @@ $config['auth_level'][99] = 'Administrator';
 | a PHP script and you can easily do that on your own.
 |
 */
-$config['base_url'] = 'http://localhost/logbook';
+$config['base_url'] = 'http://localhost/logbook/';
 
 /*
 |--------------------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


HB9HIL:
Also a note for @patrickrb: I removed the changes of your PR #1444 since the url and directory is no longer configurable. In this case the solution before your fix is more secure (decision is made in PHP). The installer HAS to be called in the correct URL. Switching the URL during the installation just causes issues and the url can be safely changed after the installation in the config.php

AndreasK79:
Can confirm this working in Xampp. I removed Wamp earlier, so I am not set up for it, but the issue were the same in both.

AndreasK79 (approved):

PR #1550, opened 2025-01-24
[Advanced Logbook] Query fix for getting band/mode/sat to the dropdowns

AndreasK79:
This PR includes fixes for the following:
1. If your logbook did not contain any locations, BAND/MODE/SAT dropdown would be blank.
2. Since the advanced logbook can filter on locations, it could happen that you had a mode/band/sat in a location that would not be fetched, since the fetch only fetched in the active logbook. Queries now fetches from locations based on user_id instead.
3. Since continents stats used get_modes, I had to swap out the function from another place.
4. A few users reported that column/map options would not save. Probably because the nature of JS. The location.reload would trigger before the ajax call was finished. Reworked the JS to close the dialog and refresh the page once the ajax call is finished.

### DIFF
--- a/application/controllers/Continents.php
+++ b/application/controllers/Continents.php
@@ -7,9 +7,10 @@ class Continents extends CI_Controller {
         $this->load->model('user_model');
         $this->load->model('bands');
         $this->load->model('logbookadvanced_model');
+		$this->load->model('gridmap_model');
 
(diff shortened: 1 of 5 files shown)


int2001:
Tested 1,2 and 3. 4 wasn't reproduceable.

1,2,3 are working as expected with this PR

PR #1558, opened 2025-01-27
Clublog gui

AndreasK79:
PR is now ready for test.
A new menu item under third party services is available.
You can trigger manually upload and download there.

I deleted the clublog library. I could not find that it was in use.
Refactored a little bit by moving a function from logbook_model to clublog_model as well.


phl0 (changes requested):
Code is working fine so far. Two issues though:

- Result of upload is shown and remaining on download tab (and upload tab)
- Only result of last location is shown on download. I.e. previous messages are overwritten

phl0 (approved):
Tested and works fine.

phl0 (approved):
Re-tested:
- cron job works across two users
- config switch to disable also works

PR #1560, opened 2025-01-27
Add QSOs per operator to statistics view

DB4SCW:
In light of the addition of club stations to wavelog and my own experiences running event callsigns, a piece of data that is highly requested most of the time is "QSOs per Operator".

I added this stat with supporting visualization to the statistics view.

There are 2 decisions I made which maybe are debatable:

1. Display this stat even for "non-clubstation" users
Of course, this stat is only hugely interesting for club stations with multiple operators. But there are some edge cases like training calls where the callsign itself does not use the club station feature (because all operators must be supervised by the callsign holder so they may not have their own login), but because of the trainer/trainee setup, there are still different operators for each QSO.

2. Handling of empty COL_OPERATOR column
I decided to set an empty or NULL operator column to the corresponding station callsign of the QSO, mostly because an empty row in the data and on the tooltips of the chart looked "wrong". Furthermore, on callsigns which are strictly personal calls and do not log operator, this strat will display the station callsign and thus, the reality of the operation.

How to test: Have QSOs in your data with at least 2 different operators (or change at least one QSO to have a different operator call to the station callsign), then click on the operators tab in the statistics view

### DIFF
--- a/application/controllers/Statistics.php
+++ b/application/controllers/Statistics.php
@@ -110,6 +110,35 @@ class Statistics extends CI_Controller {
 		echo json_encode($bandstats);
 	}
 
+	public function get_operators() {
+
+		//load logbook model
+		$this->load->model('logbook_model');
+
+		//define stats array
+		$operatorstats = array();
+
+		//get year if present
+		$yr = xss_clean($this->input->post('yr')) ?? 'All';
+		
+		//load stats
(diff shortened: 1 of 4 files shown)


phl0 (approved):
Tested and works here

AndreasK79 (approved):

PR #1562, opened 2025-01-27
Handle 2-char grids (prevent filling the error log)

phl0:
QSOs with only 2 chars as grid fill the error log:

image

This fixes by filling the 3rd and 4th position with 55 (center of grid).


int2001 (approved):
wks. but who tf is using 2char-grids?

phl0:
> wks. but who tf is using 2char-grids?

Me. :-)

Tbh: ir was a SAT QSO where I only understood the firdt two chars. The rest was eaten by QRM. So Inlogged only 2 chars hoping to get the rest from QSL (which has not arrived yet).

PR #1563, opened 2025-01-27
Add a link to VUCC modal for worked grids

phl0:
This adds a link to the VUCC popup for worked grids on the hams.at page.


int2001 (approved):
Added a QSO with DF2ET on RS-44 located at FM03
Called hams.at page within WL
pressed Grid et voila QSO shows up. Works

PR #1570, opened 2025-01-28
Use SAT export name if differs from name

phl0:
Would like to have it "AO-7" rather than "AO-07"


int2001:
Was unable to reproduce the "Problem" here.

At controller - where json is produced - satellite->name and sat_export_name are the same (with one exception: Greencube).

maybe i am missing something here. Help ;)

phl0:
You are right. Seems I missed to commit two lines of my changes. Now added. Please try again.

int2001 (approved):
works now.

phl0:
Tnx fer review.

PR #1571, opened 2025-01-28
Housekeeping

filipmelik:
Pulling out housekeeping commits from #1481 into separate PR

### DIFF
--- a/application/controllers/Dashboard.php
+++ b/application/controllers/Dashboard.php
@@ -115,7 +115,7 @@ class Dashboard extends CI_Controller
 		$data['qrz_sent_today'] = $QSLStatsBreakdownArray['QRZ_Sent_today'];
 		$data['qrz_rcvd_today'] = $QSLStatsBreakdownArray['QRZ_Received_today'];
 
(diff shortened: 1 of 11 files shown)


phl0 (changes requested):
If we make QSO count a constant we should use its value in the message text. Made a PR to achieve that: https://github.com/filipmelik/wavelog/pull/1

phl0:
The rest is working for me btw :)

phl0:
Made another PR to adjust translation.

phl0:
The code still has an issue:

image

The constant was set to 7 as you can see at the bottom. But after entering a callsign an tabbing out of the input field it still lists last 5 QSOs of that call. Created another PR https://github.com/filipmelik/wavelog/pull/3 that load the constance into the session settings so that it can be used from there.

phl0:
This works fine for me now. Are you planning to have this merged as intermediate stage and then continue on moving the variable to a user setting?

int2001 (approved):
tested:
- Dashboard
- Widget
- Log QSO

works

phl0:
So merge time and start for @filipmelik to go ahead with user setting :)

PR #1576, opened 2025-01-29
[Advanced Logbook] Saves number of results and selected locations

AndreasK79:
Based on the idea by @ethancedwards8.

Selected number of results are saved in local storage when a search is done.
Same with the selected locations.

Settings are loaded when the page is loaded.

### DIFF
--- a/application/views/logbookadvanced/index.php
+++ b/application/views/logbookadvanced/index.php
@@ -1,4 +1,6 @@
 <script type="text/javascript">
+    var user_id = <?php echo $this->session->userdata('user_id'); ?>;
+
     /*
      * Custom user settings
      */
--- a/assets/js/sections/logbookadvanced.js
+++ b/assets/js/sections/logbookadvanced.js
@@ -439,6 +439,20 @@ $(document).ready(function () {
 		'position': 'sticky', 'top': '0px', 'z-index': 1, 'background-color':'inherit', 'width':'100%', 'height':'37px'
 	})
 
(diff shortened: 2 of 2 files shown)


phl0 (approved):
This works fine. Good improvement of LBA.

ethancedwards8 (approved):

ethancedwards8:
Thanks!

int2001 (changes requested):
Works on a single-User.
But when doing impersonate for Clubstations it fails, because LocalStorage is connected to Browser and not to logged in User/Session.

Perhaps adding some kind of user_id may help here

phl0:
> Works on a single-User. But when doing impersonate for Clubstations it fails, because LocalStorage is connected to Browser and not to logged in User/Session.

Guess that fails for all the other settings then as well?

AndreasK79:
> Guess that fails for all the other settings then as well?
The column and map options are user_options, so no problem detected there.

phl0:
> The column and map options are user_options, so no problem detected there.

Great! And sorry, I didn't read the code here. Just came to my mind ...

int2001:
> Guess that fails for all the other settings then as well?

it's only failing here, because localstorage is persistant when doing the impersonate

ethancedwards8:
Maybe just invalidate local storage when switching between accounts/club stations?

phl0:
Could you please re-verify @int2001? 😇

int2001 (approved):

PR #1580, opened 2025-01-30
Take Lat/Long from exceptions if exceptions are taken

int2001:
Bug: If a DXCC was derivated by its exception, lat/long were empty.
Result was: Wrong display for last-qsos / qrz-lookup to the right of QSO-Form

this one fixes it.

testcall: JD1/JG8NQJ


AndreasK79 (approved):

AndreasK79:
@int2001 that fixed it :)

PR #1581, opened 2025-01-30
As shortcut replace dot by /P on callsign input

phl0:
Noted by DJ9BX some loggers replace a dot by "/P" suffix on callsign input. This implements this behaviour in Wavelog for QSO callsign input.


int2001 (approved):
i like. nice addiion

PR #1583, opened 2025-01-30
Users Table

HB9HIL:
if you have some users on your instance and those users have some qso's the users table becomes pretty slow..

I tried to enhance that. It should be much faster now. Obviously a dedicated counter table would be the fastest option but also a pretty uncomfortable one. So let's see how this one performs..


HB9HIL:
Will add a config switch here to reduce the stats if necessary

phl0 (approved):
1t w0rx!

int2001 (approved):
works. good improvment for medium-sized installations

PR #1593, opened 2025-01-31
More safe Club Log Handling to prevent being blacklisted

int2001:
Nothing huge, but more/better handling of Results from Club Log.

For documentation:
Club Log sent a 403 during the past, if there was a dupe within a file.
Many 403s lead to trigger their firewall.
They changed it to 200

This code handles the 200 and is a bit more verbose on other things.


phl0 (approved):
Working fine. Tested:
- Regular manual uploads
- Upload by cron
- Upload for wrong/not configured callsign
- Timeout on upload

PR #1599, opened 2025-02-01
On-Air widget

filipmelik:
See it in action [here](https://www.qrz.com/db/OK1GOD).

1] I left some questions/ideas in the code. Please, chime in with your opinions. 
2] The `get_by_callsign` I  have added might be unnecessary in case we do not want callsign in the URL. If we are OK with URL containing only `user_id` we can remove it.


int2001 (changes requested):
Good morning, @filipmelik 
Great idea.

But:
1. You already mentioned it within the code as a comment: It'll be possible for anybody to see on which QRG you're QRV. Please couple this thing to the slug-feature or make an extra-Option where the User explicit enables this one.

Reasons for that: Sensitive Data. Not every OP wants to share that information. See GDPR as well.

Furthermore:

2. If you couple it to a explicit granting, you don't need to lookup the User-ID by the Call. simply use the slug which was provided. There are already functions for that.

3. You're selecting everything of the user (cat). If he has 10 Radios you're going to show those ones. Better approach: Only select the most recent updated radio (via SQL) and process it. In 99% of all cases that is the right one. MultiBand-Ops are very rare. imho no need to display all radios which were updated within the timeout.

I like the idea of this widget.
73, DJ7NT

filipmelik:
Hey Joerg, 
thanks for coming back with your opinions!

> 1. You already mentioned it within the code as a comment: It'll be possible for anybody to see on which QRG you're QRV. Please couple this thing to the slug-feature or make an extra-Option where the User explicit enables this one.

At first, I wanted to couple this to the slug, as you mention. However, as far as I can see (both code-wise and UI-wise), it seems that slug is strongly coupled with concrete **logbook** (in DB `station_logbooks.public_slug`) and not the user (radios are coupled to user - in DB `cat.user_id`). That's the reason, I have ditched this idea and I personally vote for introducing explicit opt-in in the account settings and then using just the user ID in the URL, instead of slug or fetching user by callsign (as it is in my current draft code). Does it makes sense to you going this way?

> 2. If you couple it to a explicit granting, you don't need to lookup the User-ID by the Call. simply use the slug which was provided. There are already functions for that.

Addressed in my comment above ^^^

> 3. You're selecting everything of the user (cat). If he has 10 Radios you're going to show those ones. 

This is not exactly true. The way I have written it in the current draft is this:
- Fetch all radios of the user (not limiting just to most recently updated radio)
- Iterate over all radios and display only those, that were updated earlier than "cat radio timeout" (which is configurable by get parameter, or global wavelog setting is used if this parameter is not present). In other words, even if the user has 10 radios, only those that are within "recently updated" interval are shown, not all of them.

Reason for this behavior - I personally tend to have both of my radios turned at the same time, one radio for HF and other for UHF/VHF but sometimes I am just listening, not actively making contacts. So this way, the widget will report both my QRGs and somebody can possibly contact me on the band of choice 😊 

So this in not really designed for "multiband ops", but rather as a reminder where I can be contacted when I have my radios turned on next to me, but not currently making contacts at all. 

I can image this could be configurable by GET param - somebody could want the behavior I am proposing, and somebody just could want to have the 1, most recently updated radio at all times.  

WDYT?

int2001:
> At first, I wanted to couple this to the slug, as you mention. However, as far as I can see (both code-wise and UI-wise), it seems that slug is strongly coupled with concrete **logbook** (in DB `station_logbooks.public_slug`) and not the user (radios are coupled to user - in DB `cat.user_id`). That's the reason, I have ditched this idea and I personally vote for introducing explicit opt-in in the account settings and then using just the user ID in the URL, instead of slug or fetching user by callsign (as it is in my current draft code). Does it makes sense to you going this way?

Ok, you're right. It should be coupled to the user, not to location or log. Some kind of user-slug i think. I don't like the idea of the User_id as part of the URL. Should be something individual.

Reason: You can simply iterate through the IDs from outside. Even with explicit granting (which has to be implemented) not really good.


> * Fetch all radios of the user (not limiting just to most recently updated radio)
> * Iterate over all radios and display only those, that were updated earlier than "cat radio timeout" (which is configurable by get parameter, or global wavelog setting is used if this parameter is not present). In other words, even if the user has 10 radios, only those that are within "recently updated" interval are shown, not all of them.
> 
> Reason for this behavior - I personally tend to have both of my radios turned at the same time, one radio for HF and other for UHF/VHF but sometimes I am just listening, not actively making contacts. So this way, the widget will report both my QRGs and somebody can possibly contact me on the band of choice 😊

Roger. You're right. if you're going to add the setting for granting/revoking the feature pse add a switch "Show all recent updated radios or only last one?" - i think both usecases should be covered with that.

btw: doing the same here. but i can only work with one radio. so if i turn to my HF-TRX, i switch off / turn vol to 0 at the UHF/VHF-TRX.

filipmelik:
> > At first, I wanted to couple this to the slug, as you mention. However, as far as I can see (both code-wise and UI-wise), it seems that slug is strongly coupled with concrete **logbook** (in DB `station_logbooks.public_slug`) and not the user (radios are coupled to user - in DB `cat.user_id`). That's the reason, I have ditched this idea and I personally vote for introducing explicit opt-in in the account settings and then using just the user ID in the URL, instead of slug or fetching user by callsign (as it is in my current draft code). Does it makes sense to you going this way?
> 
> Ok, you're right. It should be coupled to the user, not to location or log. Some kind of user-slug i think. I don't like the idea of the User_id as part of the URL. Should be something individual.
> 
> Reason: You can simply iterate through the IDs from outside. Even with explicit granting (which has to be implemented) not really good.
> 

OK, can you suggest where is the best place to put the "slug"? I thought about user options, but that is not accessible, unless user is logged in, so I think we need a migration here. Nor it is maybe the best place to put this. Any idea? But the rest of the configuration widget parameters should be saved in user setting I think, just the "slug" does not fit there for me. It needs to be queried "from outside" to create "slug" -> user_id conversion. 

> > * Fetch all radios of the user (not limiting just to most recently updated radio)
> > * Iterate over all radios and display only those, that were updated earlier than "cat radio timeout" (which is configurable by get parameter, or global wavelog setting is used if this parameter is not present). In other words, even if the user has 10 radios, only those that are within "recently updated" interval are shown, not all of them.
> > 
> > Reason for this behavior - I personally tend to have both of my radios turned at the same time, one radio for HF and other for UHF/VHF but sometimes I am just listening, not actively making contacts. So this way, the widget will report both my QRGs and somebody can possibly contact me on the band of choice 😊
> 
> Roger. You're right. if you're going to add the setting for granting/revoking the feature pse add a switch "Show all recent updated radios or only last one?" - i think both usecases should be covered with that.

OK, will do

int2001:
> OK, can you suggest where is the best place to put the "slug"? I thought about user options, but that is not accessible, unless user is logged in, so I think we need a migration here. Nor it is maybe the best place to put this. Any idea? But the rest of the configuration widget parameters should be saved in user setting I think, just the "slug" does not fit there for me. It needs to be queried "from outside" to create "slug" -> user_id conversion.

yes. won't work without a migration. New field at user-table? slugname (uniqueness!). if present: widget is enabled, if not: disabled. UI at account-settings


> > Roger. You're right. if you're going to add the setting for granting/revoking the feature pse add a switch "Show all recent updated radios or only last one?" - i think both usecases should be covered with that.
> 
> OK, will do

great. tu!

int2001:
Last-QSO is without Timestamps. Of course we can discuss to add some right-management here too. unsure.
you're absoluelty right regarding the mess. it's historically grown ;)

as an impulse i like your proposal. but we need a way to migrate existing settings here, if we chose that route.

need opinion from rest of core-team here, @AndreasK79 / @phl0 / @HB9HIL

filipmelik:
> Last-QSO is without Timestamps. Of course we can discuss to add some right-management here too. unsure. you're absoluelty right regarding the mess. it's historically grown ;)
> 

yes yes, I am of course aware of that, I did not mean to blame anybody, I know how those things work 😊 just sharing my thoughts as a user here. Sometimes just the “legacy” things get out of hand over time, and would benefit from a rework. But I know this is tough cookie to crack 😁

int2001:
> yes yes, I am of course aware of that, I did not mean to blame anybody, I know how those things work 😊 just sharing my thoughts as a user here. Sometimes just the “legacy” things get out of hand over time, and would benefit from a rework. But I know this is tough cookie to crack 😁

we refactored a lot of things since start of Wavelog. fully aware of that there is a lot of legacy-code. You're welcome to help. Only guidance is:
- should be consistent for user
- easy understandable
- respect the code-style (we know - it's not optimal, but it is like it is)
- always keep the userbase in mind (from very-old-man up to nerd)
- migrations should be there

phl0:
I would (also) like to see this handled by a slug/random value for identification purposes. Rather not good to have this guessable.

HB9HIL:
> > Just an idea:
> > Why not simply using the user callsign in the url and a disable switch in the user settings? No need for an extra "user slug" or an additional column. The user callsign is no sensitive information.
> 
> That is the way the current draft is currently implemented 😊 I would also prefer going this way, as it is the easiest

I just had a phone call with @int2001 and I know understand better the concerns of having the callsign in the URL, expecially in this case. Therefore an easy suggestion:

1. Make the widget turned of per default. A user has to enable the the widget in the user settings. 
2. I would not make the slug configurable, as 99% of the users will use their callsign as slug which is exactly what we don't want. I would use the user_id, encrypt it with the encrypt methods of codeigniter and use this value as url parameter. 

```php
/** 
 * Something like this. Just an example.
 */

// Load the encryption lib like this
if (!$this->load->is_loaded('encryption')) {
	$this->load->library('encryption');
}

// Encrypt the string (you have to load the user before)
$url_slug = md5($this->encryption->encrypt($user_id));

```

This $url_slug should be stored somewhere. Another way could be to not use md5 but `urlencode()` the encryption string and `$this->encryption->decrypt()` it to get the corresponding user. The encryption lib is using the `$config('encryption_key')` to encrypt the stuff.

int2001:
> 1. Make the widget turned of per default. A user has to enable the the widget in the user settings.
> 2. I would not make the slug configurable, as 99% of the users will use their callsign as slug which is exactly what we don't want. I would use the user_id, encrypt it with the encrypt methods of codeigniter and use this value as url parameter.

Fine with that idea. tnx

filipmelik:
Hey guys! 
just had some spare time to think this through again and I have some questions/concerns here:

I kind of feel that the suggestion with the encryption is complicated and adds a lot of complexity to the code (you must ensure user has not default encryption key, you must ensure that this slug will be generated for new instances + already running instances, you must store the "encrypted slug" somewhere, and probably some more we do not know yet...)

> I just had a phone call with @int2001 and I know understand better the concerns of having the callsign in the URL

Can you please elaborate a little bit more what the concerns with the callsign are please? I also do not feel callsign is something secret. So I cannot seem to understand, why the "callsign-in-url" + "off-by-default" + "explicit-opt-in" is not enough. The user is in charge of deciding whether he wants to present this information or not. To be honest, I do not see how the encrypted URL slug should add to the overall security/privacy 🤔

int2001:
The callsign itself isn't secret.
Background:
- if you put the callsign as part of the URL, it could be an invitation to some crawlers to test every callsign for the enabled widget. Once found, one is able to derive to Operatortimes. And that's not good (Even if someone enabled the widget). So it's more for preventing script-kiddies to crawl huge instances (Hugest known instance has more than 2k of Users)
- Explicite Enabling (Opt-In): As described before the data "is currently QRV at XYZ MHz" is a dataset which is sensible in sense of GDPR/DSGVO. Please don't ask for the reasons. of course a QSO is always public. But that's the law.

I hope we get more clarity on this now.

filipmelik:
@int2001 @HB9HIL Hi guys, please check the proposed solution.

- I have decided to go with saving the slug in the users table, since it is strongly coupled to the user. The naming is of the field (slug) is generic on purpose, since it might be useful for other tasks than widgets too.
- I think the migration is a good way to go, as this will add the column and generate the slugs for new installations and for the existing users too
- I just realized that there is one case I missed, and that is when adding new user to already existing instance - I forgot to generate the slug, there, but I will do it in next commit later on, I am going to sleep now
- Another thing to do is write docs (probably on merge?)

application/migrations/236_user_callsign_public_slug.php
```diff
@@ -0,0 +1,52 @@
...
+        if (!$this->load->is_loaded('encryption')) {
+            $this->load->library('encryption');
+        }
+
+		$fetch_result = $this->db->get($this->config->item('auth_table'));
+        if ($fetch_result && $fetch_result->num_rows() > 0) {
+            foreach ($fetch_result->result_array() as $user_row) {
+                if ($user_row["slug"] === null) {
+                    $user_id = $user_row["user_id"];
+                    $user_slug_base = md5($this->encryption->encrypt($user_id));
```
int2001, line 34: concatenate the user_name to it, and i think we'll be fine regarding the obfuscation.
md5(user_id) is nothing else than $user_id - predictable.

filipmelik: You probably missed the inside of the md5 - the user id is “encrypted”, and since the encryption uses random salt, the result is different each time 😊 should be ok then wdyt?

int2001: ahhh. you're right.
so this one should indeed be fine.

int2001 (changes requested):
Tested with an existing User.
Works. Great!!!

But: I created a new User and tried to activate the widget // there was no slug. Could you please changee it / add the part for User-Creation?

filipmelik:
> Tested with an existing User. Works. Great!!!
> 
> But: I created a new User and tried to activate the widget // there was no slug. Could you please changee it / add the part for User-Creation?

Yes, I mentioned this in https://github.com/wavelog/wavelog/pull/1599#issuecomment-2667091274

I just realized this flaw yesterday in bed, will fix soon 😊

filipmelik:
> Tested with an existing User. Works. Great!!!
> 
> But: I created a new User and tried to activate the widget // there was no slug. Could you please changee it / add the part for User-Creation?

Fixed. I have used the username as a seed for user slug when creating user, since the user_id is not known until database insert. It does not matter that the seed is not the user id - we need just some semi-random value, so it shpuld be OK. I could also do it after the db insert, but that would make the code less readable.

Plus I noticed little bug with `$dashboard_map` not being taken into account when creating an account, Fixed that too.

phl0 (changes requested):
It works fine on my end. Due to v2.0.1 being released on Friday we need to move the migration to 237_user_callsign_public_slug.php and adapt application/config/migration.php accordingly. If this is done it is ready to be merged from my point of view.

phl0:
Oh and one more thing. We should display a proper error mesdage on the widget if not CAT is configured. We do not want the CI error message to be shown in the infrage like this:

IMG_20250223_091802_443

filipmelik:
Migration updated, error is now explicit instead just generic "not found".

Screenshot 2025-02-23 at 13 36 20

phl0 (approved):
Fine with me.

int2001 (approved):

int2001:
One more @filipmelik : Can u pse add a proper wiki-page for it with parameters and best-practise for an qrz-iframe and so on?

PR #1607, opened 2025-02-02
Again Zero-Problems - now at Userlist

int2001:
Type in DL0XX in search-field at useradmin. No results, even if station exists.

this one fixes it


phl0 (approved):
Works fine. Also input is now upper case. I like ✌️

PR #1609, opened 2025-02-03
Improvement for huge Usertables (See #1607)

int2001:
On huge Usertables #1607 doesn't help, because the search-field doesn't exist while trying to bind i (it is created after full render).
This one fixes it.


phl0 (approved):
Workd

PR #1610, opened 2025-02-03
[Callbook] Added qrz.ru support

AndreasK79:
In the config file you need to set callbook = qrzru

`$config['callbook']` = "qrzru"; // Options are hamqth, qrz, qrzcq or qrzru`

Username and password needs to be added as well:
```
$config['qrzru_username'] = "";
$config['qrzru_password'] = "";
```

### DIFF
--- a/application/libraries/Callbook.php
+++ b/application/libraries/Callbook.php
@@ -148,17 +148,12 @@ class Callbook {
 
 		if (!$this->ci->session->userdata('qrzru_session_key')) {
 			$result = $this->ci->qrzru->session($username, $password);
-			if ($result[0] == 0) {
-				$this->ci->session->set_userdata('qrzru_session_key', $result[1]);
-			} else {
-				$data['error'] = __("QRZ.RU Error").": ".$result[1];
-				return $data;
-			}
+			$this->ci->session->set_userdata('qrzru_session_key', $result);
 		}
 
 		$callbook = $this->ci->qrzru->search($callsign, $this->ci->session->userdata('qrzru_session_key'));
 
-		if ($callbook['error'] ?? '' == 'Invalid session key') {
+		if ($callbook['error'] ?? '' == 'Session does not exist or expired') {
 			$qrzru_session_key = $this->ci->qrzru->session($username, $password);
 			$this->ci->session->set_userdata('qrzru_session_key', $qrzru_session_key);
 			$callbook = $this->ci->qrzru->search($callsign, $this->ci->session->userdata('qrzru_session_key'));
--- a/application/libraries/Callbook.php
+++ b/application/libraries/Callbook.php
@@ -40,6 +40,13 @@ class Callbook {
 				}
 				return $this->hamqth($this->ci->config->item('hamqth_username'), $this->ci->config->item('hamqth_password'), $callsign);
 				break;
+			case 'qrzru':
+				if ($this->ci->config->item('qrzru_username') == null || $this->ci->config->item('qrzru_password') == null) {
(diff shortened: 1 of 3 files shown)


R1BLH:
I tried to test this PR (switched to branch qrzru_callbook_support).

While filling out the New QSO form I got an error
```
Callbook Search for R1BLH
QRZ.RU Error: 
```
Messages in the log:
```
ERROR - 2025-02-06 09:25:23 --> Severity: Warning --> Uninitialized string offset 0 /var/www/wavelog.cats-home.net/application/libraries/Callbook.php 151
ERROR - 2025-02-06 09:25:23 --> Severity: Warning --> Uninitialized string offset 1 /var/www/wavelog.cats-home.net/application/libraries/Callbook.php 154
```

AndreasK79:
@R1BLH thanks for testing. I don't think that is the problem though. It would only lead to the library always being loaded, and then give an error in the log that the library is always loaded. Anyway, nice find.

Seems I need to debug, to figure more out. I've sent you an email to your qrz address.

R1BLH:
> Seems I need to debug, to figure more out. I've sent you an email to your qrz address.

OK.

AndreasK79:
@R1BLH now I think it's ready for you to test.

R1BLH:
> now I think it's ready for you to test.

It works!!! Thank you very much!
Is it possible to add another improvement? If the <esurname> and <ename> tags are empty, then fill the name field of the QSO with the content of the <surname> and <name> tags.

AndreasK79:
@R1BLH ok, made the change. If all good, I'll merge it.

R1BLH:
> @R1BLH now I think it's ready for you to test.

It occurred to me that when querying data from national callbooks, the data in the national languages should be used as primary data. This means that the data in English is secondary in this case.
Of course, there is the question of the correctness of exporting to an ADIF file, because according to the specification only ASCII characters are allowed. However, national systems support UTF-8 characters in ADIF text fields. 

To get everything right, we need fields in the table that correspond to the *_INTL tags of the ADIF specification (QTH_INTL, NAME_INTL, etc. according to ADIF specification 3.1.5). As well as the corresponding fields in the QSO input/edit interface. And not only export to ADIF files, but also to ADX files.

R1BLH:
> ok, made the change. If all good, I'll merge it.

Perfect! Than you!

AndreasK79:
> It occurred to me that when querying data from national callbooks, the data in the national languages should be used as primary data. This means that the data in English is secondary in this case. Of course, there is the question of the correctness of exporting to an ADIF file, because according to the specification only ASCII characters are allowed. However, national systems support UTF-8 characters in ADIF text fields.

Should I swap then to take ename last?

> To get everything right, we need fields in the table that correspond to the *_INTL tags of the ADIF specification (QTH_INTL, NAME_INTL, etc. according to ADIF specification 3.1.5). As well as the corresponding fields in the QSO input/edit interface. And not only export to ADIF files, but also to ADX files.

*_INTL already exists in the DB, but I think this should be moved to a new discussion in that case.

AndreasK79:
> Code reading ok. All my picky suggestions resolved :D

Always good with more eyes checking the code :)

R1BLH:
> Should I swap then to take ename last?

I think it would be better (name and qth would be filled in the same language).

R1BLH:
> *_INTL already exists in the DB, but I think this should be moved to a new discussion in that case.

Yes, of course.

AndreasK79:
@R1BLH done. Merge time.

PR #1612, opened 2025-02-03
Added Timeout to HRD-Pusher

int2001:
If HRD is not available the curl hangs for about 600s.
not good.

this one adds a timeout of 10s to it


phl0 (approved):
works

PR #1613, opened 2025-02-03
Replace 0 by Ø in awards table search input

phl0:
Makes searching for slashed zero callsign in awards tables work again. Identified by @dg2ron in DXCC table on QSO logging showing worked slots for DXCC.


int2001 (approved):
tested with small DXCC (less QSOs) as well as with huge DXCCs.

works!

PR #1615, opened 2025-02-04
[Bandmap List] Added continent to columns

AndreasK79:
image


phl0 (approved):
👍

int2001 (approved):
good idea to add the continent. i made the col a bit smaller to save space (if someone works `/p` in the field)

PR #1617, opened 2025-02-04
[Advanced Logbook] Defaults were mixed, would lead to columns showing wrong

AndreasK79:
image

Since the column was defaulted to false one place, it needs to be false here too.


HB9HIL (approved):
makes sense

int2001 (approved):

PR #1624, opened 2025-02-06
Fixes two issues at DOK-Lookup. Pre/Suffix and crappy Zero(tm)

int2001:
Once a call has been logged  with its DOK the next time you work it the DOK is taken from the lastQSO.
this works in general.

But:
- If there was a Zero in the Call, it didn't work --> Fixed
- If the station had a pre- or suffix it didn't work --> Fixed.


phl0 (approved):
Nice find :)

PR #1628, opened 2025-02-08
Add link to QSO view to eQSL card download page

phl0:
This adds a link to QSO view for the eQSL card download page.


int2001 (approved):
works

PR #1629, opened 2025-02-08
Trim \r\n\t on ADIF before parsing JSON

phl0:
I use the API from the command line and use an ADIF file within the curl call. This breaks the import because ADIF file contains newlines after each <eor>. So It would make sense to strip newlines before parsing JSON.

Example command:

```
curl -X POST https://<WAVELOG_URL>/index.php/api/qso -H "Content-Type: application/json" -H "Accept: application/json" -d "{\"key\":\"<API_KEY>\",\"station_profile_id\":\"<STATION_PROFILE_ID>\",\"type\":\"adif\",\"string\":\"$(tail -n27 some_log.adi)\"}"
```

### DIFF
--- a/application/controllers/Api.php
+++ b/application/controllers/Api.php
@@ -166,6 +166,7 @@ class API extends CI_Controller {
 
 		// Decode JSON and store
 		$raw = file_get_contents("php://input");
+		$raw = $raw = preg_replace('#[\r\n\t]+#', '', $raw);
 		$obj = json_decode($raw,true);
 		$raw='';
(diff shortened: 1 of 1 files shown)


int2001 (approved):
tested with WavelogGate which uses the API - works.

i assume you tested it as well with WSJTX-Improved and so on?!

phl0:
> i assume you tested it as well with WSJTX-Improved and so on?!

Correct. Sending a single record does not have a newline (at least what I implemented). No issues detected here.

PR #1640, opened 2025-02-12
Qso edit gui button move

AndreasK79:
Moved the update from callbook button.
Before:
image

After:
image

Reason:
3 people reported on discord that they managed to wrongfully click it.


phl0 (approved):
Good idea.

int2001 (approved):
good one.

PR #1641, opened 2025-02-12
Allow freetext for operator

int2001:
Allow Free-Text for Operator when importing for a clubcall.

Scenario: Clubofficer receives an ADIF from a station which doesn't log via Wavelog. The only chance to set the OP correctly is: mass- or single-edit the QSOs.

Since the Clubmanager knows who sent him the adif, he can simply type it in.

i thought as well about selectize, @HB9HIL but dropped it, because it'll be overengeneered here in my opinion.
Tried some other approach with "datalists" here too, but dropped them also

### DIFF
--- a/application/views/adif/import.php
+++ b/application/views/adif/import.php
@@ -79,7 +79,7 @@
                         $show_operator_question = true;
                         if ($this->config->item('special_callsign') && (!empty($club_operators))) {
                             $show_operator_question = false; ?>
-                            <div class="small form-text text-muted"><?= __("Type in the operator of the imported QSOs") ?></div>
+                            <div class="small form-text text-muted"><?= __("Type in the operators callsign of the imported QSOs") ?></div>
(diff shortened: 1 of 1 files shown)


application/views/adif/import.php
```diff
@@ -79,17 +79,8 @@
...
-                                <?php foreach ($club_operators as $operator) { ?>
-                                    <option value="<?php echo $operator->user_callsign; ?>"
-                                        <?php if ($operator->user_callsign == $this->session->userdata('cd_src_call')) {
-                                            echo ' selected="selected"';
-                                        } ?>>
-                                        <?php echo $operator->user_callsign; ?>
-                                    </option>
-                                <?php } ?>
-                            </select>
+                            <div class="small form-text text-muted"><?= __("Type in the operator of the imported QSOs") ?></div>
```
HB9HIL, line 82: I'd use another text then here. 


```suggestion
                            <div class="small form-text text-muted"><?= __("Type in the operators callsign of the imported QSOs") ?></div>
```

HB9HIL (approved):
Good for me. Change the text and merge that thing :)

PR #1645, opened 2025-02-12
Add prop mode to check

int2001:
E.g.: If done 2 QSOs in a short time. One via SAT one Terrestic - both confirmed via LoTW or any other 3rd-party only one will be marked as confirmed.

this one takes care of the PROP_MODE


phl0 (approved):
This works fine and now does confirm multiple QSOs within the QSL window independently.

phl0 (approved):
Like the eQSL dupe check refactor. Less code are more specific :)

PR #1646, opened 2025-02-12
Fix some errors after data table update

AndreasK79:
@int2001 here is the fix


int2001:
Seems like we catched every bug.
Who wants to approve?

AndreasK79:
I say hopefully😊

phl0 (approved):
Works for me :)

PR #1652, opened 2025-02-13
Prevent mysql error on editing

int2001:
If someone edits a qso and posts an empty End-Time a DB-Bug is coming up.
This prevents saving the qso, if the endtime isn't given.

PS: Yes, this needs some love on top (show the user whats wrong), but for now the PR prevents some 500er Errors thrown by Webserver.

PS: compare / check with "whitespace ignore" (Tried to tidy up the qso-controller in a 2nd commit)


phl0 (approved):
Works. Good job.

PR #1659, opened 2025-02-14
Edit fence

int2001:
Collecting adjustments for edit_qso.

Mainly to make it "dummy-safe".

- [x] Check Times for validity
- [x] Implement try/catch logic at edit() and return success/no_success at least to Ajax-Save-Request
- [X] Implement checks for DXCC to be numeric
- [X] Implement return details on "No Call given"
- [X] Implement return details on "No access to station_id"
- [x] Give User Feedback whats perhaps wrong

### DIFF
--- a/application/controllers/Qso.php
+++ b/application/controllers/Qso.php
@@ -191,14 +191,15 @@ class QSO extends CI_Controller {
 		$data['iota'] = $this->logbook_model->fetchIota();
 		$data['modes'] = $this->modes->all();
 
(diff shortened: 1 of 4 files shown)


phl0 (changes requested):
Empty end time seems to not be allowed?

image

int2001:
> Empty end time seems to not be allowed?

going to check. tnx

phl0:
DXCC numeric check works:

image

phl0 (approved):
Works for me :)

PR #1661, opened 2025-02-14
Reverse-proxy install fix

zxalexis:
Fixed install under docker behind reverse-proxy. Needs additional parameter to be set up in reverse-proxy header: 
X-Request-URI $request_uri;
Still have to manually start installation invoking .../path/install/ but installation creates correct config in-place and app is working as supposed.
[R1CDY]


ethancedwards8:
I can vouch. I run in docker behind nginx-proxy-manager and have 0 issues.

zxalexis:
> Hello @zxalexis
> 
> I don't really understand why there is a fix needed. We know about a lot of installations behind all kind of different reverse proxys which do not have any issues.
> 
> Can you describe which Issue do you want to fix here?

Install behind a reverse-proxy needs manual intervention to config.php to fix URL at the end  - it is not editable in install form.  My install also did not start ./install/run.php as $directory was detected as '/'.
This fix detects reverse-proxy difference in base URL automatically.
As I see you have messed up with directories which are not always the same in URL and in php environment (and should not treated as equal in any circumstaces), so the most convinient way to clean things would be to allow to fix manually base URL in preinstall form (and $website_url in turn which goes to config.php)

HB9HIL:
which reverse proxy are you using?

zxalexis:
> which reverse proxy are you using?

Nginx native on aarch64 as reverse-proxy. Docker is on x64 machine bihind the net.

HB9HIL:
To be honest: There is probably a configuration issue with your nginx that there was a manual intervention necessary. 

I will test this setup personally and will report back okay? But it's the first time I hear the issue that there is a manual intervention necessary

HB9HIL:
I tested our current master and dev branches for this. Docker works without any issues with our provided configurations
docker-compoye.yml -> https://github.com/wavelog/wavelog/wiki/Installation-via-Docker#docker-composeyaml
nginx reverse proxy konfiguration -> https://github.com/wavelog/wavelog/wiki/Installation-via-Docker#reverse-proxy

I don't see the need to change something on such a sensitive file the the index.php

sorry

HB9HIL:
Let me say one more comment on this. I think what you meant to fix was the combination of docker, nginx as reverse proxy and Wavelog installed in a subdirectory. 

This combination is really really difficult to make work properly in the installer and this for more then one reason:
1. PHP Frameworks like Codeigniter, especially if they are a bit older like our Codeigniter 3 won't work pretty well in subdirectorys. I DO NOT recommend to install Wavelog like that. Choose a subdomain instead and work with server_name in nginx if you can't have more then one machine / IP Adress. 
2. We recently disabled the possibility to manually edit the directory in the installer. Reason for that was the high support volume for people which did not understand what to do here. More simple -> less support volume.

I'm sure this fix works for your but we stick with our recommended configurations in the wiki which work pretty well in the current code state - without subdirectories.

PR #1664, opened 2025-02-16
Order station-locations by call and description

int2001:
This one orders the station-locations within LBA, ADIF-Import, Station-chooser at logging and several other places by Call and Description


AndreasK79 (approved):

AndreasK79:
Works for me :)

phl0:
Good works. Had bothered me also. But not enough yet :D

PR #1665, opened 2025-02-16
[Contesting] moved the setsession code to document ready

AndreasK79:
While doing some testing, I found that it seems to be better to move and activate setsession on document ready.
At least while I tested, I did not have the double reload problem anymore.


int2001 (approved):
tested:
- reload
- changing views / coming back to contest
- logging out and in again

all working (for now)

PR #1668, opened 2025-02-17
Make dashboard-map user-configurable

int2001:
This one makes the Map at Dashboard User-configurable.

Global setting has priority! so if diasbled at global setting, the user-setting has no effect.
Default is "Y"

### DIFF
--- a/application/controllers/User.php
+++ b/application/controllers/User.php
@@ -200,7 +200,7 @@ class User extends CI_Controller {
 				$data['user_callsign'] = $this->input->post('user_callsign');
 				$data['user_locator'] = $this->input->post('user_locator');
 				$data['user_timezone'] = $this->input->post('user_timezone');
-				$data['user_measurement_base'] = $this->input->post('user_measurement_base') ?? 'Y';
+				$data['user_measurement_base'] = $this->input->post('user_measurement_base') ?? 'K';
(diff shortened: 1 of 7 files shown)


AndreasK79:
Tested. User options has highest priority.
Also missed one option:
image

I don't see the point in keeping the one in global options though. I say remove it, and just keep the one in user options.

AndreasK79 (approved):

AndreasK79:
I am good with these changes now.

application/controllers/User.php
```diff
@@ -200,7 +200,8 @@ function add() {
 				$data['user_callsign'] = $this->input->post('user_callsign');
 				$data['user_locator'] = $this->input->post('user_locator');
 				$data['user_timezone'] = $this->input->post('user_timezone');
-				$data['user_measurement_base'] = $this->input->post('user_measurement_base');
+				$data['user_measurement_base'] = $this->input->post('user_measurement_base') ?? 'Y';
```
phl0, line 203: This needs correction. Measurement base can be "K", "M" or "N" (or none). But "Y" does not make sense here. Mistake?

image

phl0 (changes requested):

application/controllers/User.php
phl0: I guess we should rather have "K" or "M" here instead of "Y".

PR #1673, opened 2025-02-18
[Cat API] Added CWU and CWL

AndreasK79:
As my radio returns CWU or CWL, I've added that to the API to return CW.


int2001 (approved):
works

Issue #1679, opened 2025-02-19
SN100IARU generates wrong QRZ link


lukezaa:
**This is for reporting bugs with the code, if this issue is regarding setting up on your local server please use the discussions area**

**Describe the bug**
SN100IARU generates wrong link do QRZ.com

https://www.qrz.com/db/SN10%C3%98IARU

**Version Information**
We need some information about your versions. You can get all Information in the Admin > Debug Menu
- OS:  Ubuntu 24.04
- PHP:  8.3
- MySQL/MariaDB:  MySQL 8.0.41
- Wavelog:  2.0

**Steps to Reproduce**
1. Try to log QSO with SN100IARU
2. Click QRZ

**Expected Behavior**
Redirect to SN100IARU not SN10ØIARU

**Screenshots**
![Image](https://github.com/user-attachments/assets/75dd1b3f-db6b-42b9-8486-5e52468ec6be)

**Desktop Information:**
- OS:  Ubuntu 24.04
- Browser: Firefox
- Version:  135

phl0:
Good find. Fix in https://github.com/wavelog/wavelog/pull/1680

phl0:
Fixed in dev branch now. Will be part of the next release.

PR #1680, opened 2025-02-19
Replace all slahsed zeros in qrz/hamqth link

phl0:
Fixes https://github.com/wavelog/wavelog/issues/1679.

### DIFF
--- a/assets/js/sections/qso.js
+++ b/assets/js/sections/qso.js
@@ -703,11 +703,11 @@ $("#callsign").on("focusout", function () {
 					}
 					$('[data-bs-toggle="tooltip"]').tooltip();
 				}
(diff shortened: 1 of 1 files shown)


int2001 (dismissed):
better: `replaceAll` instead of `replace`.
otherwise Calls like 'DL500POST' (or sith. like that) are failing

phl0:
> better: `replaceAll` instead of `replace`. otherwise Calls like 'DL500POST' (or sith. like that) are failing

Huh? Thats what I did ... 🤔

int2001:
> > better: `replaceAll` instead of `replace`. otherwise Calls like 'DL500POST' (or sith. like that) are failing
> 
> Huh? Thats what I did ... 🤔

4 get it - i am a bit next to myself today.

PR #1685, opened 2025-02-20
Added fence. ReUpload only to QRZ if really modified

int2001:
Check if QSO is really(!) modified. Only set to "M" if this is the case.

@HB9HIL don't know if already complete. join/add your things, if there is something


HB9HIL:
gonna check tonight.. need to sit down and wrap my head around this

phl0 (approved):
Works for me. Should at lest reduce number of useless uploads.

HB9HIL (approved):

PR #1690, opened 2025-02-20
List SATs with only displayname and prevent QSO being made via those from being uploaded to LoTW

phl0:
If we manually add a satellite that is not known to LoTW (yet) we cannot select it during logging a QSO because the list only contains satellites with their LoTW names. To work around this we show the displayname for those satellites which do not have a LoTW name (yet).


phl0:
Also added code which marks QSOs with SAT displayname and lotw = N as invalid for LoTW. This prevents those QSOs which do not have a LoTW name and lotw=N from being uploaded to LoTW.

int2001 (approved):
this seems to be okay now.
tested with:
- SAT added which has not LoTW-Name
- Shortwave-QSO upload
- Invalid one (Internet-Radio // DMR)

int2001 (approved):
better

PR #1696, opened 2025-02-24
List satellites in edit dropdown alphabetically

phl0:
We should list the birds in LBA edit dropdown alphabetically as well to prevent this:

Screenshot from 2025-02-24 08-58-17


int2001 (approved):

AndreasK79:
Makes sense :)

AndreasK79 (approved):

PR #1698, opened 2025-02-24
Widget error message formatting

phl0:
I would rather like the error messages to be shown in the widget and in widget design rather than filling our log (we cannot solve the problems on code level).
Also fixed various whitespace stuff.
/cc @filipmelik

### DIFF
--- a/application/controllers/Widgets.php
+++ b/application/controllers/Widgets.php
@@ -1,9 +1,9 @@
 <?php if ( ! defined('BASEPATH')) exit('No direct script access allowed');
 
 /*
-	
+
 	Widgets are designed to be addons to use around the internet.
-		
+
 */
 
 class Widgets extends CI_Controller {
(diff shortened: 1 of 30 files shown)


application/controllers/Widgets.php
```diff
@@ -157,9 +169,12 @@ public function on_air($user_slug = "") {
 			$this->load->view('widgets/on_air', $data);
 
 		} else {
-			show_error(
-				__("No CAT interfaced radios found. You need to have at least one radio interface configured.")
-			);
+			$theme = $this->input->get('theme', true) ?? $this->config->item('option_theme');
```
int2001, line 172: uh oh. just seeing this.
not good. the theme itself is sanitized but directly printed at the view. so with a bit cleverness one can do a XSS-Attack here.

what about checking for a valid-theme here?

application/controllers/Widgets.php
```diff
@@ -80,29 +80,46 @@ public function oqrs($user_callsign = 'CALL MISSING') {
 	 * @return void
 	 */
 	public function on_air($user_slug = "") {
+		// determine theme
```
int2001, line 83: See other annotation as well (issue checking the validity of the theme)

int2001 (changes requested):
it works in general. but while reading the code i noticed another issue.
i can try to fix this as well within this PR?! as u like...

int2001 (approved):

PR #1702, opened 2025-02-25
Fix accidently removed line for updating log when QSL was sent

int2001:
Bug was introduced bymyself, while refactoring the qrz-logic.
the update was missing.

testing:
- add some QSO's to the QSL-Queue (direct / bureau)
- go to labels
- single mark them as sent (they disappear, but arent't be marked as sent)

this one fixes that. sri!


phl0 (approved):
👍

Issue #1526, opened 2025-01-20
Contest logger does not set distance column


jax-mckenzie:
**This is for reporting bugs with the code, if this issue is regarding setting up on your local server please use the discussions area**

**Describe the bug**
Similar to #1339 when entering QSOs through the contest logger the distance is not computed and put into the database. 

**Version Information**
We need some information about your versions. You can get all Information in the Admin > Debug Menu
- OS:               docker
- PHP:              8.3.15
- MySQL/MariaDB:    10.11.6
- Wavelog:          1.9.1

**Steps to Reproduce**
Log QSOs in the contest logger, return to overview with distance column visible, observe empty distance column. 

**Expected Behavior**
Expect that it works similar to adding normal contacts or contacts via waveloggate (since #1339)

PR #1706, opened 2025-02-25
Add distance calculation to contest logging

phl0:
Addresses https://github.com/wavelog/wavelog/issues/1526 by adding calculations for distance / moving calculation further down the code.


int2001 (approved):
works (if Grid is chosen as exchange)

PR #1711, opened 2025-02-27
QSO widget remix

filipmelik:
Attempt to make QSO widget more configurable and better looking. It adds option to configure number of QSOs show, text size, theme and option to display or hide QSO times.

Since I am modifying already existing widget, the design will differ a little from what is it now. I do not know if this is an issue. If it is, one option is to create something like "qso widget new" with its own widget logic and template.  let me know what you think. 

Default designs shown here:

BEFORE:
Screenshot 2025-02-27 at 22 02 33

AFTER:
Screenshot 2025-02-27 at 22 00 13

### DIFF
--- a/application/controllers/Widgets.php
+++ b/application/controllers/Widgets.php
@@ -34,11 +34,6 @@ class Widgets extends CI_Controller {
 			$data['theme'] = "default";
 		}
 
-		// should we show time?
-		// TODO
-		$data['show_time'] = (($this->config->item('use_auth') && ($this->session->userdata('user_type') >= 2)) || $this->config->item('use_auth') === FALSE || ($this->config->item('show_time')));
-		$data['show_time'] = true;
-
 		// determine text size
 		$text_size = $this->input->get('text_size', true) ?? 1;
 		$data['text_size_class'] = $this->prepare_text_size_css_class($text_size);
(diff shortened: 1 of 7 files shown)


application/controllers/Widgets.php
```diff
@@ -14,17 +14,44 @@ public function index()
...
+			$data['theme'] = "default";
+		}
+
+		// determine text size
+		$text_size = $this->input->get('text_size', true) ?? 1;
+		$data['text_size_class'] = $this->prepare_text_size_css_class($text_size);
+
+		// number of QSOs shown
+		$qso_count_param = $this->input->get('qso_count', TRUE);
+		if ($qso_count_param != null) {
```
int2001, line 43: pse check if qso_count which was passed via `/?qso_count=nn` is numeric.

application/models/Logbooks_model.php
```diff
@@ -188,6 +188,19 @@ function public_slug_exists_logbook_id($slug) {
 		}
 	}
 
+	function user_id_from_logbook_slug($slug) {
```
int2001, line 191: this function is already in code. see https://github.com/wavelog/wavelog/blob/2dedf9f81b31f72f03ca3d802c5bda9ce72e0a89/application/models/Stationsetup_model.php#L211

suggestion: Use the one from stationsetup_model instead of adding redundant code

int2001 (changes requested):
works in general.
pse adjust the things i marked at the code.

Furthermore:
i am not sure about the formatting. the old table was "plain", so people were able to use the data with their own stylesheets while loading the plain-table via Javascript. This won't work any longer now. i personally never used that the way described, but there may(!) be people out there wo use it.

opinion on that: @HB9HIL / @phl0 / @AndreasK79 ?

application/controllers/Widgets.php
filipmelik: thanks, good idea!

application/models/Logbooks_model.php
filipmelik: wasn't aware of that one, fixed

PR #1714, opened 2025-02-28
Fix Leftover - Wrong permission check

HB9HIL:
The config value `sc_hide_usermenu` is not used anymore but it seems I forgot to change it here.

Thanks to the report of @R1BLH in #1712


int2001 (approved):
tested:
- with/without `sc_hide_usermenu` at `config.php`
- with/without `special_callsign`
- as club-officer
- as club-member
- as normal user

works

PR #1716, opened 2025-03-01
Very very strange bug. the ON-Clause for joining was wrong

int2001:
this was for a long time in the code-Base.

Repro:
A lot of Users PLUS a lot of station_ids are causing a mem-exhaust, because the join NEVER took care about the binding-element - the user_id.

tnx for Co-investigating @phl0


phl0 (approved):
Tested in prod and dev. Fixes the bux.

PR #1720, opened 2025-03-02
[Accumulated Statistics] Fixed missing propagation part in the DXCC query

AndreasK79:
Numbers will in some cases not be the same in DXCC between monthly and yearly when selecting propagation.


int2001 (approved):
it works. tnx a lot

phl0 (approved):
Fixes the diff in my log :) Tnx

PR #1726, opened 2025-03-03
Use contest name in QSO view as well (as does LBA)

phl0:
LBA shows contest name from contest table whereas the QSO view shows the ADIF name:

image

image

### DIFF
--- a/application/models/Logbook_model.php
+++ b/application/models/Logbook_model.php
@@ -2084,7 +2084,7 @@ class Logbook_model extends CI_Model {
 
 	function get_qso($id, $trusted = false) {
 		if ($trusted || ($this->check_qso_is_accessible($id))) {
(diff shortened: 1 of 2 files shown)


int2001 (approved):
Good one. Works here. Tested with:
- Contest-QSO which is present aswell at contest-table
- Contest-QSO which has to longtext, because it isn't at contest-table

But: the Filter at LBA filters on the ADIF-Name. so we're showing everywhere the long-name (if exist) but at filtering we use the ADIF-One.

phl0:
> But: the Filter at LBA filters on the ADIF-Name. so we're showing everywhere the long-name (if exist) but at filtering we use the ADIF-One.

This has been fixed in https://github.com/wavelog/wavelog/pull/1725 ;-)

PR #1731, opened 2025-03-04
Fix footer quoting in JS

DJ3CE:
This PR fixes a malformed quoting (… until someone uses double-quotes in a translation…) in the JS part of the footer.

If a translation uses a single-quote, that single quote ended the JS-string, giving an error.

This fix simply uses double quotes to circumvent the issue here, but further steps should be taken (escaped `__(...)`-variant maybe?).

### DIFF
--- a/application/views/interface_assets/footer.php
+++ b/application/views/interface_assets/footer.php
@@ -1311,7 +1311,7 @@ $($('#callsign')).on('keypress',function(e) {
 						    if (data.error == 'not_logged_in') {
 							    $(".radio_cat_state" ).remove();
 							    if($('.radio_login_error').length == 0) {
-								    $('.qso_panel').prepend('<div class="alert alert-danger radio_login_error" role="alert"><i class="fas fa-broadcast-tower"></i> ' + '<?= sprintf(__("You're not logged in. Please %slogin%s"), '<a href="' . base_url() . '">', '</a>'); ?>' + '</div>');
+								    $('.qso_panel').prepend('<div class="alert alert-danger radio_login_error" role="alert"><i class="fas fa-broadcast-tower"></i> ' + "<?= sprintf(__("You're not logged in. Please %slogin%s"), '<a href=\"' . base_url() . '\">', '</a>'); ?>" + '</div>');
 							    }
 						    }
 						    // Put future Errorhandling here


HB9HIL (approved):
Good find
In our setup this is the way to go

phl0:
I need some more input on this. On my system the English translation "You're not logged in" is written to the JS code as follows:

```
$('.qso_panel').prepend('<div class="alert alert-danger radio_login_error" role="alert"><i class="fas fa-broadcast-tower"></i> ' + 'You&#039;re not logged in. Please <a href="https://BLABLABLA.de/wavelog/">login</a>' + '</div>');
```

So the single quote is printed as HTML entity &#039;. That means a single quote does would not break the JS code. Where is the issue?

DJ3CE:
I'm working with the english translation as well and got the fatal `'` 'printed' into the . Working with Firefox 135, Linux, Wavelog 2.0.1 (Commit da342c7b07e3edaf95de471de9ad9847e673129a).

DJ3CE:
That's right, obviously :). Server side is the oldest supported PHP on an Apache. Any specific questions?

phl0:
> That's right, obviously :). Server side is the oldest supported PHP on an Apache. Any specific questions?

Which version? :) Seems like escaping functions are not working (properly) here. Using PHP 8.1.2 here.

DJ3CE:
PHP 7.4.30 (oldest supported by Wavelog)

DJ3CE:
… some PHP digging later: Looks like `htmlentities()`, which gets called within `_encode()` from `_gettext()`/`__()` has had its default flags changed with PHP8.1 (see https://www.php.net/manual/de/function.htmlentities.php), where the default flags now include 'ENT_QUOTES', thus explaining the difference in our observations. So I'd guess PHP 8.0 is affected as well.

phl0:
If I switch off ENT_QUOTES on my PHP8.1 instance I can reproduce the error:

image

Could you try this one instead of changing the quotes (which would run into problems if one has double quotes in a translation):

```
diff --git a/application/hooks/gettext/vendor/php-gettext/gettext.php b/application/hooks/gettext/vendor/php-gettext/gettext.php
index 22fae16dd..7767d7028 100644
--- a/application/hooks/gettext/vendor/php-gettext/gettext.php
+++ b/application/hooks/gettext/vendor/php-gettext/gettext.php
@@ -211,7 +211,7 @@ function _encode($text)
                if ($source_encoding != $target_encoding) $text = mb_convert_encoding($text, $target_encoding, $source_encoding);
        }
 
-       return htmlentities($text);
+       return htmlentities($text, ENT_QUOTES);
 }
 
 // Custom implementation of the standard gettext related functions
```

HB9HIL:
> If I switch off ENT_QUOTES on my PHP8.1 instance I can reproduce the error:
> 
> image
> 
> Could you try this one instead of changing the quotes (which would run into problems if one has double quotes in a translation):
> 
> ```
> diff --git a/application/hooks/gettext/vendor/php-gettext/gettext.php b/application/hooks/gettext/vendor/php-gettext/gettext.php
> index 22fae16dd..7767d7028 100644
> --- a/application/hooks/gettext/vendor/php-gettext/gettext.php
> +++ b/application/hooks/gettext/vendor/php-gettext/gettext.php
> @@ -211,7 +211,7 @@ function _encode($text)
>                 if ($source_encoding != $target_encoding) $text = mb_convert_encoding($text, $target_encoding, $source_encoding);
>         }
>  
> -       return htmlentities($text);
> +       return htmlentities($text, ENT_QUOTES);
>  }
>  
>  // Custom implementation of the standard gettext related functions
> ```

Currently you can really control which types of quotes are used in the translation. We would nee to write a check for that. Therefore wqe discussed that a year ago or so, when I implemented gettext, that it's important to use double quotes around translations in JS. Exactly for that reason. It's more likely to happen that someone uses single quotes (for example in spanish or other languages where single quotes are part of the language itself) then double quotes as weblate is escaping double quotes pretty good so far. Therefore I still approve this PR and recommend to merge it.

phl0:
I am fine with that as well. Just wanted to point out the double quote issue (agn). So merge?

PR #1732, opened 2025-03-04
Fix SQL: Switch from '=' to 'in'

DJ3CE:
There was a left-over `station_id =` where in the following a list of station-ids was presented – of course leading to an SQL error. So this commit/"PR" simply swaps the `=` for an `in` and adds some brackets ;) (… as already used in the line above).


phl0 (approved):
Tested and fixes the bug. Thanks for your contribution.

PR #1733, opened 2025-03-04
fix: use correct data source for radio cat_timeout in on-air widget

filipmelik:
Fix based on report from [here](https://github.com/wavelog/wavelog/discussions/1574#discussioncomment-12372971)


phl0 (approved):
Works

PR #1734, opened 2025-03-04
Fix Iota map, bring IOTA up to default standards

DJ3CE:
Some additional work on the IOTA-Award page. Make considered qsl-method selectable (copied over from other award pages) and replace the hardwired qsl-method in the code. Fix the worked/confirmed part on the map.


AndreasK79 (approved):

AndreasK79:
@dj3ce thanks for updating the award :)

PR #1737, opened 2025-03-06
Fix LBA search for SATs without LoTW name

phl0:
With the SAT edit function we can have satellites in our db which are not accepted by LoTW yet and therefore do not have a (lotw)name in the satellite table. LBA search for those baed on orbit etc. fails due to empty argument for joining the tables.

### DIFF
--- a/application/models/Logbookadvanced_model.php
+++ b/application/models/Logbookadvanced_model.php
@@ -371,7 +371,7 @@ class Logbookadvanced_model extends CI_Model {
 			SELECT qsos.*, dxcc_entities.*, lotw_users.*, station_profile.*, satellite.*, dxcc_entities.name as dxccname, mydxcc.name AS station_country, exists(select 1 from qsl_images where qsoid = qsos.COL_PRIMARY_KEY) as qslcount, coalesce(contest.name, qsos.col_contest_id) as contestname
 			FROM " . $this->config->item('table_name') . " qsos
 			INNER JOIN station_profile ON qsos.station_id=station_profile.station_id
-			LEFT OUTER JOIN satellite ON qsos.COL_SAT_NAME = satellite.name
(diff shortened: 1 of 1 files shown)


phl0:
@int2001 Please asses DB timings for a comparison. Seems like it has some impact :)

int2001 (approved):
it feels faster. i donk't know why (the left-hand of the join is still the same), but it is...

works

AndreasK79:
It is not faster at all here when testing. LBA search is slowed noticeably with this query change :(

application/models/Logbookadvanced_model.php
```diff
@@ -371,7 +371,7 @@ public function searchDb($searchCriteria) {
 			SELECT qsos.*, dxcc_entities.*, lotw_users.*, station_profile.*, satellite.*, dxcc_entities.name as dxccname, mydxcc.name AS station_country, exists(select 1 from qsl_images where qsoid = qsos.COL_PRIMARY_KEY) as qslcount, coalesce(contest.name, qsos.col_contest_id) as contestname
 			FROM " . $this->config->item('table_name') . " qsos
 			INNER JOIN station_profile ON qsos.station_id=station_profile.station_id
-			LEFT OUTER JOIN satellite ON qsos.COL_SAT_NAME = satellite.name
+			LEFT OUTER JOIN satellite ON qsos.COL_SAT_NAME = COALESCE(NULLIF(satellite.name, ''), NULLIF(satellite.displayname, ''))
```
int2001, line 374:
```suggestion
			LEFT OUTER JOIN satellite ON qsos.col_prop_mode='SAT' and qsos.COL_SAT_NAME = COALESCE(NULLIF(satellite.name, ''), NULLIF(satellite.displayname, ''))
```

int2001: This speeds it up a little:

query before PR: 0.15 seconds
query with PR: up to 2s
query with suggestion: 0.55s

AndreasK79 (approved):

AndreasK79:
The last change helped😊

PR #1741, opened 2025-03-07
Ignore country-column since dxcc is joined

int2001:
i accidently logged a US Station with a wrong DXCC. At Country there was "Alaska" in it (But DXCC was "291" and not "6" - explanation at repro)
result was this at timeline:

image

Repro with example (Not easy to repro):
- log KL3R - DXCC will be 6 and Country Alaska
- correct the call (only the call) to K3LR (misheard during contest and in hurry)
- wait for LoTW-Confirm. LoTW changes DXCC to 291 but won't touch country-field

however: the patch fixes:
- the report (only adif-dxcc tells the DXCC)
- the LoTW-Importer (if dxcc is provided and updated, take the country as well)

### DIFF
--- a/application/models/Logbook_model.php
+++ b/application/models/Logbook_model.php
@@ -3727,7 +3727,7 @@ class Logbook_model extends CI_Model {
 		}
 	}
 
-	function lotw_update($datetime, $callsign, $band, $qsl_date, $qsl_status, $state, $qsl_gridsquare, $qsl_vucc_grids, $iota, $cnty, $cqz, $ituz, $station_callsign, $qsoid, $station_ids, $dxcc = null, $ant_path = null) {
+	function lotw_update($datetime, $callsign, $band, $qsl_date, $qsl_status, $state, $qsl_gridsquare, $qsl_vucc_grids, $iota, $cnty, $cqz, $ituz, $station_callsign, $qsoid, $station_ids, $country = null, $dxcc = null, $ant_path = null) {
 
 		$data = array(
 			'COL_LOTW_QSLRDATE' => $qsl_date,
@@ -3751,6 +3751,10 @@ class Logbook_model extends CI_Model {
 
 		if (($dxcc ?? '') != '') {
 			$data['COL_DXCC'] = $dxcc;
+		}
+
+		if (($country ?? '') != '') {
+			$data['COL_COUNTRY'] = $country;
 		}
 
 		if ($ituz != "") {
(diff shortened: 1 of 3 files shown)


application/models/Timeline_model.php
```diff
@@ -63,11 +63,11 @@ public function get_timeline_dxcc($band, $mode, $propmode, $location_list, $qsl,
 
 		$sql .= $this->addQslToQuery($qsl, $lotw, $eqsl, $clublog, $qrz);
 
-		$sql .= " group by col_dxcc, col_country
+		$sql .= " group by col_dxcc
 			order by date desc";
 
 		$query = $this->db->query($sql, $binding);
-
+		log_message("Error",$this->db->last_query());
```
AndreasK79, line 70: Remove this debug line?

int2001: of course. sorry!

application/models/Logbook_model.php
```diff
@@ -3727,7 +3727,7 @@ function qrz_update($primarykey, $qsl_date, $qsl_status) {
 		}
 	}
 
-	function lotw_update($datetime, $callsign, $band, $qsl_date, $qsl_status, $state, $qsl_gridsquare, $qsl_vucc_grids, $iota, $cnty, $cqz, $ituz, $station_callsign, $qsoid, $station_ids, $dxcc = null, $ant_path = null) {
+	function lotw_update($datetime, $callsign, $band, $qsl_date, $qsl_status, $state, $qsl_gridsquare, $qsl_vucc_grids, $iota, $cnty, $cqz, $ituz, $station_callsign, $qsoid, $station_ids, $country = null, $dxcc = null, $ant_path = null) {
```
phl0, line 3730:
```suggestion
	function lotw_update($datetime, $callsign, $band, $qsl_date, $qsl_status, $state, $qsl_gridsquare, $qsl_vucc_grids, $iota, $cnty, $cqz, $ituz, $station_callsign, $qsoid, $station_ids, $dxcc = null, $country = null, $ant_path = null) {
```

I think you reversed country and dxcc here ...

phl0 (changes requested):

application/models/Logbook_model.php
phl0: This even throws a mysql error:

image

phl0 (approved):
This works for me now :) Updates DXCC *and* country from LoTW.

AndreasK79 (approved):

PR #1742, opened 2025-03-07
Reset auto_increment to 1 before re-inserting user data

phl0:
Addresses https://github.com/wavelog/wavelog/discussions/1740 by resetting the auto increment counter upon clearing the table.


AndreasK79 (approved):

int2001 (approved):
works. resets to 1 after truncate

PR #1743, opened 2025-03-07
Make consequent use of NULL/1 and adjust defaults

int2001:
Bug:

- Call Awards --> DXCC and have a look at confirmed DXCC (remember the value)
- Now press "Show" (without changing any filter). The Number of cnfmd DXCC changes

Reason was: Different default-Values when calling page "fresh" compared to first "post".
Addon: since the Award-Model checks for `NULL`, we should take care that everything is `NULL` or `1`. Before this patch it was sometimes `0`, sometimes `NULL`, sometimes a string...


AndreasK79 (approved):

AndreasK79:
Consistency is key for correct numbers :)

PR #1757, opened 2025-03-12
Fix or-grouping conditions when (not) to update requested qsl

DJ3CE:
Without an or-group, you may end up not updating due to either the QSL being already requested (but with different method) or the method already being set (but status is not requested).

The other (Paper-)QSL-update related methods (ignore, update_sent, sent) don't suffer from this, as there's only one condition checked.

Maybe additionally it would be desirable to prevent setting QSL-sent-status to requested if the QSL status is already confirmed here in the SQL as well?

### DIFF
--- a/application/models/Logbook_model.php
+++ b/application/models/Logbook_model.php
@@ -1958,9 +1958,9 @@ class Logbook_model extends CI_Model {
 			);
 
 			$this->db->where('COL_PRIMARY_KEY', $qso_id);
-			$this->db->or_group_start();
+			$this->db->group_start();
 			$this->db->where('COL_QSL_SENT !=','R');
-			$this->db->where('COL_QSL_SENT_VIA !=', $method);
+			$this->db->or_where('COL_QSL_SENT_VIA !=', $method);
 			$this->db->group_end();
 
 			$this->db->update($this->config->item('table_name'), $data);
--- a/application/models/Logbook_model.php
+++ b/application/models/Logbook_model.php
@@ -1958,8 +1958,10 @@ class Logbook_model extends CI_Model {
 			);
 
 			$this->db->where('COL_PRIMARY_KEY', $qso_id);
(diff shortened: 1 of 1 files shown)


int2001 (changes requested):
This one is deadly, since it updates EVERYTHING.

the resulting SQL will be:

`UPDATE [table] set [col] WHERE COL_PRIMARY_KEY=$qso_id OR (COL_QSL_SENT != 'R' AND COL_QSL_SENT_VIA != $method)`

what you wanted to have (assumption) is:

`UPDATE [table] set [col] WHERE COL_PRIMARY_KEY=$qso_id AND (COL_QSL_SENT != 'R' OR COL_QSL_SENT_VIA != $method)`

application/models/Logbook_model.php
```diff
@@ -1958,8 +1958,10 @@ function paperqsl_requested($qso_id, $method) {
 			);
 
 			$this->db->where('COL_PRIMARY_KEY', $qso_id);
+			$this->db->or_group_start();
 			$this->db->where('COL_QSL_SENT !=','R');
 			$this->db->where('COL_QSL_SENT_VIA !=', $method);
+			$this->db->group_end();
```
int2001, line 1964: To achieve that, what i assumed in my comment, try this instead:

```suggestion
            $this->db->group_start();
            $this->db->where('COL_QSL_SENT !=','R');
            $this->db->or_where('COL_QSL_SENT_VIA !=', $method);
            $this->db->group_end();
```
```

DJ3CE: Tnx!

int2001 (approved):
good find.

Repro of bug:
- Mark QSO as "Requested bureau" via Context-Menu
- Try to Mark it as "Requested direct" via Contest-Menu
won't work.

with this patch it is working agn. tnx

phl0 (approved):
works :)

PR #1759, opened 2025-03-12
Match satellite's tles via NORAD catalogue id

phl0:


AndreasK79 (approved):

AndreasK79:
Tested updating. Norad_id is added and updated. TLE's are updated.

int2001 (approved):
tested this working.
NORAD-ID appears in database.

i think we're safe with int(9) for a time, or should we switch to character (number only makes sence, if we're going to calculate with it - or using it as a key...)

phl0:
> i think we're safe with int(9) for a time, or should we switch to character (number only makes sence, if we're going to calculate with it - or using it as a key...)

According to the official information it is a 9-digit number. That why I chose. No plans for keys or calculations at this time.

PR #1762, opened 2025-03-12
Order by col_time_on at eqsl-sync for upload (better debugging)

int2001:
Order the upload for eQSL by TIME_ON Ascending.
Background: If sth. fails, there's a much better chance to identify the "fishy" QSO


phl0 (approved):
Tested, works. Good fix.

PR #1763, opened 2025-03-12
Reduce Grids / Take oldest

int2001:
Example given: Following QSO-Constellation:

- Worked DF2ET at JO30 (Entered that at grid)
- First Time working JO30!
- DF2ET confirms, but with conjunction at VUCC_GRIDS via LoTW for JO30,JO31

Now you've JO30 new (for Grid itself), JO30 new (at VUCC_GRIDs) and JO31 new (at VUCC_GRIDs).

**Timeline**-View now doubles JO30, because the "reducer" for the Grid (out of grid and vucc_grid) was missing.

This one fixes it

### DIFF
--- a/application/models/Logbook_model.php
+++ b/application/models/Logbook_model.php
@@ -3793,6 +3793,7 @@ class Logbook_model extends CI_Model {
 				$data['COL_GRIDSQUARE'] = $qsl_gridsquare;
 				$data['COL_DISTANCE'] = $this->qra->distance($station_gridsquare, $qsl_gridsquare, 'K', $ant_path);
 			} elseif ($qsl_vucc_grids != "") {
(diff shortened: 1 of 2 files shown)


AndreasK79:
Bug confirmed, and fix works :)

phl0 (approved):
Code for grid reducting works. Now trying an implementation for deleting single grid on incoming LoTW cnfm with vucc_grids.

phl0:
Just pushed a small change that empties COL_GRIDSQUARE in case COL_VUCC_GRIDS is set in an incoming LoTW confirmation. This should prevent the buggy QSO grids in timeline from even being created.

int2001:
did a bit research, and it seems like either GRIDSQUARE **XOR** VUCC_GRIDs should be populated: 
image

so i think this is fine

PR #1765, opened 2025-03-14
strip < from qslmessage, because eqsl can't handle it

int2001:
it's much more than only being sure the content it properly encoded.

Current situation: imagine an instance with 100 Users, everyone has 50.000 QSOs == 5Mio QSOs.
Now assume 10 Users have either invalid-QSOs (Mode: "HTTY" or "Band 40m / Freq 21.800") or a Nickname which doesn't exist at eQSL.
Since we're processing QSO by QSO, and those exceptions weren't handled in the past, the syncer has to do this over and over and again until the Database gives us a timeout (becuse eQSL is so slow).

At the next run of thee syncer the same will happen. causing that some users never will be synced.

This patch adresses proper cathing of those exceptions. in detail:
- Wrong Nickname? abort sync for user // remove eQSL-Nick from station and show user a message (if manual synced)
- Invalid QSO (0 of 1 record inserted) ? Mark it as invalid, so it won't be fetched for future syncs.
- QSO-Payload invalid (0 of 0 records inserted): Skip the whole user for now and create a error-log-entry to do further investigations (exceptions we're not aware of)
- HTTP-Error other than 400,404 or 500 from eQSL: Skip the user for now and produce logentry and warning (exceptions we're not aware of)

### DIFF
--- a/application/models/Eqslmethods_model.php
+++ b/application/models/Eqslmethods_model.php
@@ -261,6 +261,7 @@ class Eqslmethods_model extends CI_Model {
 		// basic curl options for all requests
 		curl_setopt($ch, CURLOPT_RETURNTRANSFER, 1);
 		curl_setopt($ch, CURLOPT_HEADER, 1);
+		curl_setopt($ch, CURLOPT_CONNECTTIMEOUT, 10);
 
 		// use the URL we built
 		curl_setopt($ch, CURLOPT_URL, $adif);
--- a/application/models/Eqslmethods_model.php
+++ b/application/models/Eqslmethods_model.php
@@ -209,7 +209,7 @@ class Eqslmethods_model extends CI_Model {
 
 		// adding qslmsg if it isn't blank
 		if ($qsl['COL_QSLMSG'] != '') {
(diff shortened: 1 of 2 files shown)


application/models/Eqslmethods_model.php
```diff
@@ -45,6 +45,7 @@ function uploadUser($userid, $username, $password) {
 		$clean_userid = $this->security->xss_clean($userid);
 
 		$qslsnotsent = $this->eqsl_not_yet_sent($clean_userid);
+		log_message("Error",$this->db->last_query());
```
phl0, line 48: Forgotten debug?

phl0 (changes requested):

application/models/Eqslmethods_model.php
int2001: yes. still draft - but anyway thanks (better heads-up than forgotten)

and this PR seems like it doesn't catch all.

phl0 (approved):

phl0:
Tested with various test cases. Works for me (after modification).

PR #1779, opened 2025-03-21
Add display of hams of note

phl0:


application/models/Update_model.php
```diff
@@ -484,4 +484,37 @@ function update_norad_ids() {
 		return;
 	}
 
+	function update_hams_of_note() {
+		$this->db->empty_table("hams_of_note");
+		$this->db->query("ALTER TABLE hams_of_note AUTO_INCREMENT 1");
+		$file = 'https://www.ham2k.com/data/hams-of-note.txt';
+		$handle = fopen($file, "r");
```
int2001, line 491: consider using curl with timeout and so on instead of deprecated (and sometimes disabled) fopen

phl0: Implmeneted.

phl0:
The ham-2k team approved usage of the hamsofnote file. So undrafting here to merge after review.

application/migrations/239_add_hams_of_note.php
```diff
@@ -0,0 +1,76 @@
...
+					'null' => TRUE,
+					'default' => '',
+				),
+			));
+			$this->dbforge->add_key('id', TRUE);
+			$this->dbforge->create_table('hams_of_note');
+			$this->db->query("ALTER TABLE hams_of_note ADD INDEX `callsign` (`callsign`)");
+		}
+
+		if ($this->chk4cron('update_update_hon') == 0) {
```
int2001, line 46:
```suggestion
if ($this->chk4cron('update_update_hamsofnote') == 0) {
```

tbh: my fault

int2001 (changes requested):
see comment (was my fault).

rest works

PR #1780, opened 2025-03-21
Use last worked sat for preset of flightpath selected satellite

phl0:
Use last worked SAT for preset of selected satellite in flightpath view.

Aka. What was selsat used for? Did not have any results here so removed it.


application/models/Satellite_model.php
```diff
@@ -22,6 +22,22 @@ function get_all_satellites_with_tle() {
 		return $this->db->query($sql)->result();
 	}
 
+	function get_last_worked_sat() {
+		$this->load->model('logbooks_model');
+		$logbooks_locations_array = $this->logbooks_model->list_logbook_relationships($this->session->userdata('active_station_logbook'));
+		$location_list = "'" . implode("','", $logbooks_locations_array) . "'";
+		$sql = "select COL_SAT_NAME as sat from " . $this->config->item('table_name') .
+			" where station_id in (" . $location_list . ")" .
```
int2001, line 30: this wont work if you do a Non-SAT QSO in between

int2001 (changes requested):
just for documentation - as already discussed on another way:
the PR - in its current state - will break the link from Sat-Passes to the Sat-Map.

Reason: you removed the logic for `$selsat` at the controller.

phl0:
Link from passes page restored. Flightpass takes the selected sat from passes page or last worked satellite if that is empty.

AndreasK79:
Works :)

int2001 (approved):
works. nice feature

PR #1785, opened 2025-03-22
[Flightpath] View passes when clicking on sat

AndreasK79:


phl0 (approved):
This is a great idea. Love it.

PR #1787, opened 2025-03-23
Check key before doing huge SQLs

int2001:
While testing a third-party software, I stumbled across our statistics endpoint.
Nothing really dangerous, but it fires SQLs to the backend even without a valid key.

This patch prevents the endpoint from being used for a possible DoS attack


phl0 (approved):
works

PR #1788, opened 2025-03-23
[Quick Lookup] Added POTA

AndreasK79:
image


phl0 (approved):
Tested and works fine :) Useful addon.

PR #1790, opened 2025-03-24
[QSO Entry] Add summary for more awards

AndreasK79:
I added more award summary tables in the QSO entry.
image

- [x] POTA can have multiple entries, just use the first.
- [x] Gridsquare shows for the first grid only.
- [x] I also need to add some text about which entry is looked up, like dxcc name, cq zone etc.
- [x] If wwff, gridsquare, pota, sota or state is empty, nothing happens, which is probably confusing for the user.
A small text should probably be added that the input information is empty.
- [x] When you click between tabs, they are reloaded automatically. Perhaps this should be stopped, and a refresh button should be added instead, to minimize db calling.
- [x] IOTA support added
- [x] Translations needs to be fixed
- [x] Fix DXCC QSO list popup
- [x] Filter US State on DXCC in the frontend


phl0 (changes requested):
This needs an extra filter for DXCC on the "US State" tab. Seems it currently filters for state field only. In this example I entered DJ7NT and got "RP" for state. This is a valid value for Germany DXCC 230 but not for US :)

image

AndreasK79:
Seems we need to handle the state issue in the front-end, as it is filtered in the backend.

phl0 (approved):
Works fine for me now. Didn't spot any more issues so far.

PR #1793, opened 2025-03-24
[Flightpath] Draws sat path +- 60 minutes

AndreasK79:
image
Added AOS/LOS and status too:
image


int2001 (approved):
works.

did you try it with a bright-theme? the green background is hard to read there...

phl0 (changes requested):
Code works fine but we need to adjust colors for non-dark themes I am afraid.
image

AndreasK79:
@phl0 Classes are fixed.

phl0 (approved):
Looks good now after fix:

image

PR #1797, opened 2025-03-25
Be aware of NULL for hon-lookup

int2001:


phl0 (approved):
Cannot repro. But catching this does not harm.

PR #1804, opened 2025-03-25
Autofill az ele qso

int2001:
Autofill Azimuth & Elevation while working live-SATs.

todo-list:

- [x] Calculate Az/Ele for SAT
- [x] Catch Errors, where no TLE is available
- [x] Stop calculating on reset/removing SAT
- [X] Start Calculations when SAT is entered
- [x] Start Calculations when FAV-Band with SAT was chosen
- [x] put __current__ Home-QTH as lat/long into volatile JS-Vars and change them upon location-change
- [x] implement HomeQTH instead of 0/0 for calculations
- [x] Allow az/el autofill for POST QSO logging
- [ ] make it bulletproof
- [ ] ...


phl0 (approved):
Works nicely for me here :)

phl0:
Tnx for allowing me to log elevations below 0° :-)

Issue #1805, opened 2025-03-25
Obfuscate API Key


DA1EE:
**What Feature is Missing?**
Hi, I would like to ask you to obfuscate the API keys, e.g. the QRZ.com API Key in the Station Location.

**Why is This Feature Important?**
When I do showcase Wavelog to other Hams, I might go into the details of my station location. But I do not want to share the API key.

**Is This Feature Personal or Beneficial to Others as Well?**
Yes, it is beneficial for everyone ;)

PR #1806, opened 2025-03-26
Use stars to hide QRZ-Key and add "eye"-button

int2001:
should fix https://github.com/wavelog/wavelog/issues/1805

Default-View with this patch:
image

With a click on the eye the pw will be revealed
image


AndreasK79 (approved):

phl0 (approved):
Works. We should apply the same to API keys for HRDlog and QO-100 DX Club?

int2001:
> Works. We should apply the same to API keys for HRDlog and QO-100 DX Club?

sure. y not. i try to add

phl0 (approved):
Works fine now for all API-key/password fields.

PR #1812, opened 2025-03-27
Skip showing arrow for geostationary orbits

phl0:


int2001 (approved):
works.
tested switching between LEO/GEO. No problems.

Should we remove the path as well? not important, just idea

phl0:
> Should we remove the path as well? not important, just idea

Yeah. Why not. Will check.

phl0:
Added path printing to GEO-condition. Pls re-review.

int2001 (approved):
works. path is away for QO100

PR #1813, opened 2025-03-27
Added AOS and LOS to MapView/Info

int2001:
Show Next AOS or LOS Azimuth at info-Modal:

LOS:
image

AOS:
image


phl0 (approved):
Works fine here

Issue #1810, opened 2025-03-26
Option to add Custom Hams of Note


DA1EE:
**What Feature is Missing?**
Hi, I really like the implemented Ham of Note feature. I would really enjoy, if I could generate my own "Hams of Note" file (one or even a couple), and add it as a reference. Similar to the function of PoLo. This would allow me, to update my own files with hams of note.

**Why is This Feature Important?**
I would no longer be dependent on the HAM2K Hams of Note file, but use other files as well. E.g. from my local club, contest clubs or someone else.

**Is This Feature Personal or Beneficial to Others as Well?**
It would be beneficial for a lot of people.

int2001:
You can adjust the URL where the list is fetched only by altering the code here: https://github.com/wavelog/wavelog/blob/ebeaec4e3a4c2ffc6303098e9639b46e00aca9bb/application/models/Update_model.php#L490

Hope this helps at least for the first time, until we possibly(!) find another solution

DA1EE:
Hi, thanks for the hint.  I might add a copy of that file, and test that for myself, but on the other hand, I might just wait, until it is available, if it will come anytime.

I do understand, that it does not have highest priority.

int2001:
Feature is available in dev.

PR #1814, opened 2025-03-27
Make Hams Of Note configurable at Adminmenu

int2001:
Should resolve https://github.com/wavelog/wavelog/issues/1810

### DIFF
--- a/application/controllers/Options.php
+++ b/application/controllers/Options.php
@@ -6,26 +6,18 @@
 
 class Options extends CI_Controller {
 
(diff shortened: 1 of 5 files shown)


phl0 (approved):
Works. We have some work to do re the file format. I guess we should add a format description here. And the update code is prone to different line endings. Not a problem of this PR itself but it will cause issues. Guess I need to work on that (separately).

phl0 (approved):
Works fine.

PR #1817, opened 2025-03-28
Support multiple grids or pota

AndreasK79:
The award summary in QSO entry now supports multiple grids or potas with tabs:
image
image


int2001 (approved):
works.

as well for very short grids like JO,KO

PR #1818, opened 2025-03-28
Handle cases of no AOS/LOS

phl0:
Fix display of "null°"

image

Happens with QO-100 (no LOS) and IO-86 (no AOS)


int2001 (approved):
interesting. tested it while developing yesterday. worked there.
perhaps some different calculations today, or some patch which reverted it.

however: this works.

PR #1821, opened 2025-03-29
(Re-)calculate distance upon updating QSOs from callbook within LBA

phl0:


int2001 (approved):
tested:
- Call with callbook-informations: Works
- Call without anything at Callbook: Works as well (Stays empty / no dist / no error)

PR #1822, opened 2025-03-30
Fix an issue where qsl_sent wasn't set

int2001:
This one fixes two issues:

- [x] Issue 1:
Fresh QSO:
1. mark qsl as received. qsl arrow turns green. qrz arrow turns orange in UI (only)
2. reload, qrz is red
Logically correct because QRZ was never sent, but UI-Issue

- [x] Issue 2:
Fresh QSO:
1. mark qsl as sent. qsl arrow turns green.
2. mark qsl as received, qsl sent arrow turns green
3. reload, qsl received is red
Logically incorrect. Patch should solve that

### DIFF
--- a/application/models/Logbook_model.php
+++ b/application/models/Logbook_model.php
@@ -1880,7 +1880,7 @@ class Logbook_model extends CI_Model {
 			);
 
 			$this->db->where('COL_PRIMARY_KEY', $qso_id);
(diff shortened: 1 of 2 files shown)


phl0 (approved):
Works as expected now. Both test cases with positive result here.

PR #1828, opened 2025-04-01
Ignore empty QSL-Fields on empty/import

int2001:
Some logs emit weird values for (e)QSL/LoTW-Date-Fields (not NULL, but zero-byte-length, so date is defined but empty).
Wavelog tries to parse that (because it's defined) and fails with an error-message.
But: Errormessage isn't shown by most clients (e.g. FLDigi 4.2.0.6)

This one ignores those "empty" fields:

More details see:
https://github.com/wavelog/wavelog/discussions/1820#discussioncomment-12683871

### DIFF
--- a/application/models/Logbook_model.php
+++ b/application/models/Logbook_model.php
@@ -4205,7 +4205,7 @@ class Logbook_model extends CI_Model {
 	 qslrdate, qslsdate
 	 */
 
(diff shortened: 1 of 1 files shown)


HB9HIL (approved):
makes sense to me

PR #1838, opened 2025-04-03
[Sat TLE] Manual upload

AndreasK79:
You can now add/delete TLE for each satellite in the satellite view.
Validation is done by JS in the frontend, and you can use 2 or 3 line TLE.


phl0 (approved):
Works for me. Could not break it :D

int2001 (approved):
Works here. I like

PR #1839, opened 2025-04-03
Add markers to +/- 90 min position

phl0:
Adds markers and labels to the +/- 90 min position.
Tbd: hide for GEO sats.


AndreasK79:
Love it :)

PR #1841, opened 2025-04-04
Add foreign key constraint also to satmode table

phl0:
This makes sure that no satmodes exists without corresponding satellite. satmode gets deleted upon deletion of its satellite.

### DIFF
--- a/application/config/migration.php
+++ b/application/config/migration.php
@@ -22,7 +22,7 @@ $config['migration_enabled'] = TRUE;
 |
 */
 
-$config['migration_version'] = 240;
+$config['migration_version'] = 241;
 
 /*
 |--------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


int2001 (approved):
tested:
- manually deleting a SAT --> Mode is dropped as well
- updating TLE's --> no impact (good)
- updating SATs from LoTW --> no impact (good)

can't spot a problem here. i am not 100% sure if we delete/truncate (FK's aren't working on truncate of the parent-table) tables at cronjobs.

phl0:
> can't spot a problem here. i am not 100% sure if we delete/truncate (FK's aren't working on truncate of the parent-table) tables at cronjobs.

Reviewed the code. No truncate in the code for updating satellites or TLEs. So good to go imho.

PR #1845, opened 2025-04-05
[ITU Zones] Fixed missing zones in master table

AndreasK79:
Created migration script to update dxcc_master with missing zones.


int2001 (approved):
technically this works (tested).
re the content: good enrichment. didn't check if everything fits.

PR #1846, opened 2025-04-06
Reset distance-Field on new QSO/reset

int2001:
Distance was taken from previous QSO, if no grid was filled.

repro:
log DJ7NT and directly afterwards CT1AAA (doesn't exist).
CT1AAA-QSO has same dist like DJ7NT from your QTH.

this one resets the dist-field (after logging, on reset, etc.)


AndreasK79 (approved):

phl0 (approved):
Fix confirmed.

PR #1852, opened 2025-04-07
[Advanced Logbook] Added distance to batch edit

AndreasK79:
You can now set or blank distance on marked QSOs.


int2001 (approved):
great. works (after i understood xactly what it does ;)

PR #1854, opened 2025-04-08
Make counting of counties case independant

### LINKED ISSUE #1850: Same county in all caps vs mixed upper and lower case counted as different counties
**This is for reporting bugs with the code, if this issue is regarding setting up on your local server please use the discussions area**

**Describe the bug**
The US Counties display (Awards > USA > US Counties) counts the same county stored in all caps as a different county than the same county name stored in mixed case. 

**Version Information**
- Wavelog:         1.9.1

**Steps to Reproduce**
1. Go to 'Awards > USA > US Counties'
2. Click on any 'Counties Worked'
3. See that counties with the same name but different case are counted as separate counties.

Image

**Expected Behavior**
Same county should be counted as the same, regardless of county name case.

phl0:
fixes #1850

### DIFF
--- a/application/models/Counties.php
+++ b/application/models/Counties.php
@@ -99,7 +99,7 @@ class Counties extends CI_Model
 
 		$bandslots_list = "'".implode("','",$bandslots)."'";
 
-        $sql = "select distinct COL_CNTY, COL_STATE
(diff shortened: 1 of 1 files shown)


phl0:
Tnx @AndreasK79 this is not an issue of case sensitivity here but a case of spelling (whitespace). This cannot be fixed. So closing here.

PR #1857, opened 2025-04-09
Removed CDN for not needed files

AndreasK79:
Seems two of the JS files that were loaded, is not needed.

Added message if no satellites with TLE was found:
image

### DIFF
--- a/application/views/satellite/flightpath.php
+++ b/application/views/satellite/flightpath.php
@@ -48,7 +48,7 @@
 		</form>
 
 		<?php } else { ?>
-			<?= "No satellites with TLE found. Please update via CRON or satellite page. If you have no access to do this, ask your admin!"; ?>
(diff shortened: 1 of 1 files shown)


application/views/satellite/flightpath.php
```diff
@@ -46,12 +45,17 @@
 					echo '>' . $sat->satname . '</option>'."\n";
 				} ?>
 			</select>
+		</form>
 
+		<?php } else { ?>
+			<?= "No satellites with TLE found. Please update via CRON or satellite page. If you have no access to do this, ask your admin!"; ?>
```
HB9HIL, line 51: You forgot the gettext function

```suggestion
			<?= __("No satellites with TLE found. Please update via CRON or satellite page. If you have no access to do this, ask your admin!"); ?>
```

HB9HIL (changes requested):
just a minor thingy :)

HB9HIL (approved):

PR #1858, opened 2025-04-09
fix[installer] make sure the correct config dir is created if not existent

HB9HIL:
in the case the correct config directory is not existent yet (which is the case by non default environments), we can create it to prevent the installer from failing. This does not affect normal user installations but makes installer development with docker more stable.


int2001 (approved):
about the directory: good find.
about the ignore of docker-compose.yaml? i don't know if thats the best idea. However: i'll approve and we'll see

PR #1877, opened 2025-04-14
Add option to switch on callsings for all LBA map markers

phl0:
This addresses https://github.com/wavelog/wavelog/discussions/1871 by adding a switchbox for callsign labels for all map markers. Disabled by default as it may spoil the LBA map :)

image


int2001 (approved):
nice one.
tested and works (even with 5k Calls)

PR #1885, opened 2025-04-17
Truncate HoN input

phl0:


int2001 (approved):
works. now we have fences for everything here.
tested it as well on mobile.

PR #1893, opened 2025-04-20
Added number of results to simplesearch-footer

int2001:
Adds the number of found matches to the result-view.

e.g.:

image


phl0 (approved):
w0rx

Issue #1832, opened 2025-04-01
When the length of the label in adif is incorrect, the system will always handle it instead of reporting an error


vastsea-wuji:
I received a log with errors today, and when I imported it to wavelog, I was never successful. After examination, error lines were found in the log, and the line line was like this

`
<CALL:6>B0AB<GRIDSQUARE:4>OM14<MODE:3>FT8 <RST_SENT:3>-17 <RST_RCVD:-16> <QSO_DATE:8:D>20240303 <TIME_ON:4>1323 <qso_date_off:8>20240303 <time_off:6>132344<BAND:3>15M <freq:9>21.075887 <station_callsign:5>BA0AA<my_gridsquare:6>OM16uc <EOR>
`
This may be a manually edited log with a signal report filled in for the length of the RST_RCVD field, resulting in a negative length, I manually corrected this error and successfully imported. It has other errors, but does not affect the import process

This causes PHP to keep running the script until the timeout is reached. If possible, I request that the error line number be displayed for troubleshooting when the error log line is encountered

English is not my native language, I used the translation software, the tone may be very stiff, sorry!
:D

PR #1899, opened 2025-04-23
Fix for #1832 - when length of ADIF tag is wrong

AndreasK79:


phl0 (approved):
Checked the code with the mentioned example and it works without premature exit or timeout.

int2001 (approved):
tested this several days on my prod - no issues

PR #1904, opened 2025-04-26
[Advanced Logbook] Added !empty search parameter

AndreasK79:
In the text input fields, I added support for writing !empty to find where those columns are filled with something.

Added a help button:
image

And some information:
image

### DIFF
--- a/assets/js/sections/logbookadvanced.js
+++ b/assets/js/sections/logbookadvanced.js
@@ -595,7 +595,7 @@ $(document).ready(function () {
 			type: 'post',
 			success: function (html) {
 				BootstrapDialog.show({
-					title: 'Advanced Logbook Help',
+					title: lang_gen_advanced_logbook_help,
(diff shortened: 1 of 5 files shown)


assets/js/sections/logbookadvanced.js
```diff
@@ -589,6 +589,36 @@ $(document).ready(function () {
 		processNextCallbookItem();
 	});
 
+	$('#helpButton').click(function (event) {
+		$.ajax({
+			url: base_url + 'index.php/logbookadvanced/helpDialog',
+			type: 'post',
+			success: function (html) {
+				BootstrapDialog.show({
+					title: 'Advanced Logbook Help',
```
HB9HIL, line 598: Can we add a JS var in `views/logbookadvanced/index.php` so we can translate this string aswell?

AndreasK79: Sure thing. Added now!

HB9HIL (approved):
like :)

PR #1913, opened 2025-04-28
[Dashboard] User option for banner

AndreasK79:
User option to remove the dashboard banner. This must be the most requested feature I've seen so far. Not that I get why people are bothered by a simple banner...

As this option was added before multi-user was a thing, it was originally placed in the global options.


int2001 (approved):
works.
tested:
- Existing User who has never set en/disable (admin) --> Works
- Existing User en-/Disabling (admin/user) --> Works
- New User (user) --> Works

PR #1921, opened 2025-05-01
Buffering qsos when (temporary) offline

int2001:
Situation:
- If you killed your Internetconnection (e.g. by HF) or have weak internet, the QSO could not be saved.
- You've to wait until it's back and save or postlog the QSO

With this patch:
- The QSO will be buffered (at localstorage) // Timeout for saving is set to 10seconds - be patient ;)
- The buffer will be processed along with a (successful) logged new QSO
- The buffer will be processed when leaving the page. (not 100% reliable)
- The buffer will be processed when re-entering the QSO-Log-page.
- The buffer can hold multiple QSOs. Once the page was able to store them at the backend, the buffer is reduced by the successful ones.

### DIFF
--- a/assets/js/sections/qso.js
+++ b/assets/js/sections/qso.js
@@ -124,6 +124,7 @@ $("#qso_input").off('submit').on('submit', function (e) {
 					$("#noticer").show();
 					prepare_next_qso(saveQsoButtonText);
 					$("#noticer").fadeOut(2000);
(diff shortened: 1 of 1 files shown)


AndreasK79 (approved):

AndreasK79:
From what I was able to test, I got the backlog triggered. It worked by re-entering the page, and the second attempt to log a qso. The backlog was then logged.

PR #1938, opened 2025-05-06
[Club Log] Check if directory exists

AndreasK79:
Added an extra check for the directory where the temp file is written to.


phl0 (changes requested):
This will always override $filepath. Fix -> https://github.com/AndreasK79/wavelog/pull/44

phl0 (approved):
This works now

AndreasK79:
@phl0 thanks for noticing my mistake. I did it to provoke an error, and forgot to fix it.

PR #1946, opened 2025-05-11
add support for ldap

derFogel:
Hi,

we use ldap in to manage our members of our clubstation. The "recent" change with the special event callsings broke out integration so we kept the old version.
I finally had some time to hack some ldap implementation together.

The current implementation has a "bug" in development mode, where after logging in and creating the user account you get the following errors:
image
after pressing f5 you are logged in.

Personally I would like to add a new column to the usermodel called something like `ldap_user`. This would make the authentication much better. The current implementation tries to authenticate the user against the wavelog db and if that fails, it retries via ldap. But I don't know how to do that.

This is more a proof of concept that and no production ready code so I await your feedback.
Cheers
Lieven, DF1CN (ex DF1ASH)

### DIFF
--- a/application/config/config.sample.php
+++ b/application/config/config.sample.php
@@ -782,9 +782,32 @@ $config['max_login_attempts'] = 3;
 |--------------------------------------------------------------------------
 | Disable User QSO Count in User List (Admin Menu)
 | Reason for this setting is to prevent performance issues on large installations
-| where the QSO count is not needed. Set to true to disable the QSO count. 
+| where the QSO count is not needed. Set to true to disable the QSO count.
 | This also hides the last Operator for CLubstations. Default is false.
 |--------------------------------------------------------------------------
  */
 
- $config['disable_user_stats'] = false;
+$config['disable_user_stats'] = false;
+
+/*
+* LDAP Configuration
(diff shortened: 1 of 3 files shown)


int2001:
This won't work.

Please:
1. PR against dev (i am going to close this one. You'll need to PR a new one regardless if we accept it or not - but master doesn't work)
2. Describe a test-setup, how we can test it with minimal effort (i think no one of us has a running ldap-env)

Structural/Logical:
- pse have a look at default user-adding (User-Form for adding). There are many default-vals and a few other things to populate. We know - it's not really sexy to rely on the form - but that's the situation currently. Fill at least the date_format, measurement-base, user_cols, language, default-bands, etc.
- don't know if it's really a good idea to create a user, if it doesn't exists but comes via LDAP. i'd suggest an extra controller instead which handles new-users and so on. this would make it independend.
You could add a new controller which would be used as "default-login-page", if the instance uses ldap. That controller would handle:
  - New users / Registration (Accept GDPR-Terms, whatever things...) 
  - Session-creation for valid users
  - etc.

HB9HIL:
@derFogel SSO Support is soon implemented #2978

PR #1948, opened 2025-05-12
Fix URL for SOTA API

vk3arr:
(Undocumented) SOTA API has changed its URL and requests to api2 are going via a translation layer.   New URL removes translation layer in between requests and goes to new URL base.

Cheers,
Andrew
VK3ARR
SOTA IT Group leader


HB9HIL (approved):
Thx :)

phl0 (approved):
Works for me.

PR #1949, opened 2025-05-12
Replace 0 by slahed version in LoTW report

phl0:


int2001 (approved):
tested working:
- Show slashed Zero if cfm.
- Display details (not affected, since it works by id)

PR #1950, opened 2025-05-12
Add some hyperlinks to WWFF award page

phl0:


int2001 (approved):
Both links tested:
- QSO-Details: Works
- Link to WWFF: Works

PR #1952, opened 2025-05-12
Show LoTW confirmed gridsquare even if not updated in DB

phl0:
With https://github.com/wavelog/wavelog/pull/324 we implemented a check whether the LotW confirmed square is more precise than the one we already have in the log. If that it not the case we do not update the gridsquare. As a side effect the (not-updated) gridsquare is not shown in the results of a (manual) LoTW sync:

image

In this case the grids which were confirmed on LoTW were already in the log. But the results pretend that there was no grid (confirmed). This fixes the issue so that the already confirmed grid is shown in case it is already in the log:

image


int2001 (approved):
works

Issue #1935, opened 2025-05-06
LotW Upload failing in DB_query_builder


xyz667:
**Describe the bug**
My previous LotW certificate expired, so I applied for an extension and got a new cert.
The new cert was exported to p12 file without password using TQSL 2.8.1 and uploaded to Wavelog. Wavelog shows a valid certificate.
My QSOs don't upload to LotW anymore, however.

**Version Information**
We need some information about your versions. You can get all Information in the Admin > Debug Menu
- OS:               DietPi v9.12.1
- PHP:              8.2.28
- MySQL/MariaDB:    10.11.11-MariaDB-0+deb12u1
- Wavelog:          2.0.3

**Steps to Reproduce**
1. Wait for cron triggered LotW upload -> no upload.
2. Go to 3rd Party Services - LotW and start manual sync -> no upload

There is no error message displayed in the UI!
However, this is in the log:
`ERROR - 2025-05-06 06:34:25 --> Severity: error --> Exception: where_in() expects $values to be a non-empty array /var/www/system/database/DB_query_builder.php 884`

The certificate in question is valid from 2025-05-01 to 2028-04-30 for QSOs from 2024-05-01 to 2028-04-30.
The last successful QSO upload was on May 1st, still with the previous certificate (which expired on that day).

How could I help troubleshooting this issue?

EDIT: another user on my instance has the same problem with his recently renewed LotW cert.

phl0:
Can you please add some more details? How many station locations do you have? Which QSOs are supposed to be uploaded?
It seems that this is not LoTW specific but fails to select correct station locations.

xyz667:
> Can you please add some more details? How many station locations do you have? Which QSOs are supposed to be uploaded? It seems that this is not LoTW specific but fails to select correct station locations.

@phl0 thanks for the replay.
This is a setup with two users, both with failing LotW syncs after recent cert update.

My user has one station logbook with three station locations. Each of those locations has a valid LotW cert. Only one of those certs expired, was renewed, and uploaded to Wavelog. The other two still have valid certs, but no new QSOs to sync.
Nothing changed in the station setups in recent months, only a renewed cert uploaded to the 3rd party services LotW screen.

As mentioned, Wavelog shows the renewed callsign cert as valid with correct dates (see initial issue message) and "not synced". (Not sure if the "not synced" status only showed after I deleted the cert und re-uploaded it while troubleshooting.)
[...]

PR #1953, opened 2025-05-13
Check for empty sats for LoTW operations

phl0:
Preparing to fix https://github.com/wavelog/wavelog/issues/1935.


phl0:
This is ready for review now. Both unsupported modes and satellites are now checked for being empty before adding them to the SQL statements. This prevents bugs as described in https://github.com/wavelog/wavelog/issues/1935 where no unsupported SATs caused SQL queries to fail.

int2001:
looks good so far.

i'd like to see a additional `AND` at the updater, that it doesn't update records to invalid, which are already updated...

int2001 (approved):
nice. works

AndreasK79 (approved):

phl0:
Closes https://github.com/wavelog/wavelog/issues/1935.

PR #1959, opened 2025-05-15
Get rid of config directory for clublog uploads

phl0:
The config dir is not needed for clublog uploads so remove it.


AndreasK79 (approved):

AndreasK79:
Upload works :)

PR #1963, opened 2025-05-16
Rename reset button in public search result table

phl0:
Just a test trying to refactor search functions in the DT results table.
Need to find out how to rename "Search" next to the DT search box. Layout does not work for some reason.


int2001 (approved):
hopefully the confusion will stop with this.
Great...

Issue #1980, opened 2025-05-21
Post QSO - Band field cleared when frequency is changed only on 4m band


redgeler:
Hello, When trying to post some QSOs on the 4m Band. I've hit an issue where upon entering the frequency it cleared the band field, preventing the QSO from being saved.

As a work-around, I accepted the default frequency of "70.2" and then edited the QSO to update the frequency.

Also found the same behaviour for the 1.25cm Band.

I have tested this with two separate servers/instances of Wavelog. 
Tested using two desktop computers with multiple browsers. Edge, Firefox and Chrome.

**Version Information**
Instance1
- OS: Ubuntu 24.04.2
- PHP: 8.3.6
- MariaDB: 10.11.11-MariaDB-0ubuntu0.24.04.2
- Wavelog: 2.0.3

Instance2
- OS: Ubuntu 24.04.2
- PHP: 8.3.6
- MariaDB: 10.11.11-MariaDB-0ubuntu0.24.04.2
- Wavelog: 2.0.4

**Steps to Reproduce**
1. Go to "Post QSO"
2. Select the "4m" Band (The Frequency is automatically set to "70.2")
3. Change the Frequency  e.g. "70.3"
4. Do anything to change the focus away from the the Frequency field
5. The Band field is cleared.

**Expected Behaviour**
The Band to remain populated in the Band field.

**Screenshots**
Select the 4m Band
![Image](https://github.com/user-attachments/assets/dcee4155-f6f1-4707-aca5-cc319a2cfb24)
Change the Frequency
![Image](https://github.com/user-attachments/assets/fedce947-7f83-469b-b5e7-e004d5c8e814)
When focus is lost the band is cleared.
![Image](https://github.com/user-attachments/assets/a0575bbb-9f0e-417b-a47b-140059f3224d)

**Desktop Information:**
- OS: Windows 10
- Browser:  Edge, Chrome, Firefox
- Version:   136.0.3240.76, 136.0.7103.114, 138.0.4

Is there any further info I can provide?

Rob.

phl0:
Fix proposed in https://github.com/wavelog/wavelog/pull/1982.

redgeler:
Apologies for the delayed reply. I applied the changes manually and tested. That fixed my issue.

I did see the comment on the PR by HB9HIL. I did a quick search and can see that some countries have an allocation starting at 68 MHz. Denmark and Norway may have an allocation ending as high as 74.8 MHz

Thank you very much.
[...]

PR #1982, opened 2025-05-21
Add 4m to radiohelpers JS

phl0:
Closes https://github.com/wavelog/wavelog/issues/1980.

### DIFF
--- a/assets/js/radiohelpers.js
+++ b/assets/js/radiohelpers.js
@@ -34,6 +34,9 @@ function frequencyToBand(frequency) {
 	else if(result >= 50000000 && result <= 56000000) {
 		return '6m';
 	}
+	else if(result >= 70000000 && result <= 70500000) {
+		return '4m';
+	}
 	else if(result >= 144000000 && result <= 148000000) {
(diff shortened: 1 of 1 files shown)


int2001 (approved):

assets/js/radiohelpers.js
```diff
@@ -34,6 +34,9 @@ function frequencyToBand(frequency) {
 	else if(result >= 50000000 && result <= 56000000) {
 		return '6m';
 	}
+	else if(result >= 70000000 && result <= 70500000) {
```
HB9HIL, line 37: I just checked the DARC Country list for the possible range of the 4m band. Seems like we have to make this range a bit wider. For reference you can check Bosnia & Herzogovina

```suggestion
	else if(result >= 68000000 && result <= 75000000) {
```

phl0: As I missed this comment I pushed this change upstream with https://github.com/wavelog/wavelog/commit/c970b07bf7e7c40beed0807848880e9c7e66c63b

PR #1983, opened 2025-05-21
Update wapmap.js

pa8s:
Correcting typo in FR province marker coordinate (7.7 -> 5.7) , now the FR marker is in the right location (was Ost-Friesland Germany)

Screenshot from 2025-05-22 00-17-03


phl0 (approved):
Good point :-)

PR #1986, opened 2025-05-23
Make sure empty options are printed with any value

int2001:
This one fixes a bug in LBA, where LBA stops working.

Mitigations:

- [X] Make sure only setted values are printed
- [x] Make sure only valid values are stored


phl0 (approved):
Cannot reproduce the bug but code runs ok :)

PR #1987, opened 2025-05-23
Fix saving of LBA-Cols

int2001:
if browser posts neither TRUE nor FALSE when trying to save User-options, set them to default-Values.

@AndreasK79 would be nice if you could be my guest to do the same while retrieving/processing the values which are not true/false


AndreasK79 (approved):

AndreasK79:
Tested working here!

PR #2004, opened 2025-05-31
Modified Dockerfile for Non-root user.

John-WE1DER:
I have modified the Dockerfile to use a non-root user for running the container in keeping with docker security best practices. I have also included a modified docker-compose.yml file from my environment that has been modified to use the Dockerfile. I have redacted information from the docker-compose.yml file that is unique to my environment, but have left settings for using [traefik](https://traefik.io/traefik) reverse proxy, [DX Cluster API](https://github.com/int2001/DXClusterAPI), and locally managed docker volumes. Also, I did write a crude readme to help guide others through using the non-root Dockerfile (could definitely use some polishing). I have tested this locally and all seems to work. I have yet to run it on my production server, but I have no reason to believe that the result will be any different.

**Sorry for any potential headaches. I did not realize that I did not properly ignore a file containing security compromising information in my override file. As I had to force push to my repo to remove this file, I am opening a new pull request.


int2001:
This reads very good. Nice improvement.

But i have some concerns:
1. original Dockersetup was created "lean" that unexperienced users are also able to launch it via docker. Your's include the (optional) DXClusterAPI, (commented/disabled) functions for traefik and a lot of things which need to be configured by the User.
2. the docker-compose.yaml wasn't part of the repo for a reason. Most people (we know of) use their consumer-NAS with an UI to copypaste their yaml into it (See [wiki](https://github.com/wavelog/wavelog/wiki/Installation-via-Docker))
3. Regarding the dockerfile and the fact you refactored it to non-root (great!!), i am not sure if this could break existing installations.

HB9HIL:
Hello @John-WE1DER 

Thanks for putting effort in this. I'm on the same page as @int2001 

1. There is no need to merge a Markdown file with your extended PR description -> `non-root-docker.md`
2. As @int2001 already said. We didn't add the docker-compose file here for a reason and we would like to stick with that

About the non-root thing

You are right. In a lot of situations running a container as non-root user makes absolutely sense. This sense is usually running services which require a dedicated user. In this case though apache is already running as `www-data`. Besides that you decided to give the non-root user full permission to the webroot (and let apache run with this user). This doesn't make sense. If you want a full non-root user you would strictly seperate that. 

Unfortunately this PR is not mergable for us as it disagrees with some decisions we made about our docker service. We may come back to this topic but at the moment our current setup works.

PR #2006, opened 2025-06-01
[Advanced Logbook] Instead of blanking distance with batch edit, recalculate

AndreasK79:
As blanking didn't really make sense, I changed it to recalculate.
image


phl0 (approved):
This works for me. Tbh the old code failed on my db anyway:

image

PR #2009, opened 2025-06-02
[Satellite] Enforce numeric input on satellite qrg

AndreasK79:


int2001 (approved):
works.

1st fence: frontend/js
2nd fence: backend / only ints allowed

phl0 (approved):
This works fine for me now :)

Issue #2005, opened 2025-06-01
Subject: Feature Request: Grid Square Input via Map Click in QSO Logging Interface


XiangJiang-Wang:
Dear Wavelog Development Team,

First, thank you for your ongoing work on Wavelog – it’s an invaluable tool for amateur radio operators worldwide. I’d like to suggest a small but impactful enhancement to the QSO logging interface.

Current Workflow:
When logging a new QSO, users manually type grid squares (e.g., EM48) into the GRIDSQUARE field. While functional, this relies on prior knowledge or external tools to determine the grid.

Proposed Feature:
Add interactive grid square selection to the map widget (right panel of the QSO logging window). Specifically:

Enable a "Click-to-Set Grid" mode (e.g., via a new map toolbar button).

When activated, clicking any point on the map would:

Automatically convert the clicked coordinates (lat/long) into a 6-character Maidenhead grid square (e.g., EM48qx).

Populate this value into the GRIDSQUARE (or MY_GRIDSQUARE) field.

Optionally, display the clicked grid as a temporary marker/popup for confirmation.

Benefits:

Faster logging: Especially for portable/mobile operators who know their location visually but not their grid.

Reduced errors: Eliminates manual entry mistakes.

Intuitive UX: Aligns with mapping conventions in other geo-aware software.

Technical Considerations:

Existing maps (Leaflet/OpenStreetMap) likely support click-event handling.

Libraries like maidenhead (JavaScript) can handle coordinate-to-grid conversion.

Could extend to both MY_GRIDSQUARE (user’s grid) and GRIDSQUARE (contact’s grid) via toggle.

This feature would streamline logging for field operators and newcomers alike. Thank you for considering this suggestion!

phl0:
Implemented some basics for that. Needs fixing the calculation however.

phl0:
Integrated in dev branch with https://github.com/wavelog/wavelog/pull/2010. Please test and report.

AndreasK79:
I added some more stuff with #2012 .
1. fullscreen
2. grid overlay
3. show grid on mouse hover

PR #2010, opened 2025-06-02
Determine DX gridsquare by clicking map

phl0:
Adresses https://github.com/wavelog/wavelog/issues/2005. Needs fixing calculating 2nd set of characters (i.e. 5th and 6th place of grid).


phl0:
Reworked the code to use existing JS code instead of AJAX call to PHP lib.

AndreasK79 (approved):

AndreasK79:
Tested working here.

PR #2014, opened 2025-06-04
allow propagation in logbook overview as column

HB9HIL:
Allow choosing "Propagation" as column in the logbook overview. Direct feature request by DK5EW


phl0 (approved):
Works fine for me :)

image

PR #2015, opened 2025-06-05
Add option to push logged QSOs to another WL instance

phl0:
Add options to push logged QSOs to another WL instance. Currently PoC state :)

### DIFF
--- a/application/config/migration.php
+++ b/application/config/migration.php
@@ -22,7 +22,7 @@ $config['migration_enabled'] = TRUE;
 |
 */
 
-$config['migration_version'] = 245;
+$config['migration_version'] = 246;
 
 /*
 |--------------------------------------------------------------------------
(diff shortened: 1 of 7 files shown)


iu2frl:
Can't wait to have this!

Maybe, assuming it's possible, it would be nice to be able to select just some specific locations or logbooks to send to the remote server

int2001:
> Can't wait to have this!
> 
> Maybe, assuming it's possible, it would be nice to be able to select just some specific locations or logbooks to send to the remote server

tell us more about the reasons you need it.
this PR is currently stale, because there's no real usecase for it.

either you've a master-log or not. why have multiple instances?

iu2frl:
I have a private instance 

> > Can't wait to have this!
> > Maybe, assuming it's possible, it would be nice to be able to select just some specific locations or logbooks to send to the remote server
> 
> tell us more about the reasons you need it. this PR is currently stale, because there's no real usecase for it.
> 
> either you've a master-log or not. why have multiple instances?

I have an instance on a private server which is not publicly exposed and that's where I work on, that one is configured to my radios, external services and so on. Then I have a public instance on a public cloud provider which I'd like to use to create widgets and dashboards to be published on my website or on QRZ. That way if the public server ever gets attacked, everything is still safe in my private one

HB9HIL:
> That way if the public server ever gets attacked, everything is still safe in my private one

Isn't that the reason why we do backups and firewalls? :-D

phl0:
Closing here as its not of broader use. Keeping patch in my own branch.

PR #2020, opened 2025-06-07
[Advanced Logbook] Added county column

AndreasK79:
Started adding county to LBA:
- [x] County column
- [x] Option to show/hide column
- [x] Search option


int2001 (approved):
Tested:
- Dis-/Enabling County: Works
- Filtering: Works
- Map/GlobeMap: Works

PR #2021, opened 2025-06-07
Persist after edit at SAT-Editor

int2001:
Problem:
When using the Admin-Sat-Editor and filtering for e.g. "SO-", the filter is lost after editing TLE or Modes.
So if you copypaste things from a similar SAT you've to type in the filter again (or chose manually). Minor annoying.

Solution (in this PR)
- Save search short before reloading the (updated) page and fetch it again from localstorage shortly after rendering to fill the search-item again. 
- Remove saved search (it's only stored for "surviving" the reload which is done onClose of TLE or Mode-Edit)


AndreasK79 (commented):
Works :)

PR #2022, opened 2025-06-07
Multi sat

int2001:
Allow multiple SATs to be selected (and saved as fav in localstorage).

- [X] Refactor PassCalc for multiple SATs
- [X] Refactor SkedCalc for multiple SATs
- [x] Make multiselect nice


phl0 (approved):
Tested and works for me. My already stored presets didn't work. But those also failed on current dev. Deleted and re-added and working fine now in both branches. Go for merge.

PR #2023, opened 2025-06-07
[Advanced Search] Added button to open qso results in the advanced logbook

AndreasK79:
image
Transfers the qsoids via localstorage, and open them in the search.


AndreasK79:
@int2001 thanks for spotting that. Fixed now!

PR #2033, opened 2025-06-14
[Advanced Logbook] Added qth column

AndreasK79:


int2001 (changes requested):
Saving of the setting doesn't work.

See: See https://github.com/AndreasK79/wavelog/pull/47/files

PR #2036, opened 2025-06-16
[Advanced Logbook] Can search for empty cq zone or itu zone

AndreasK79:
Added empty search for cq and itu zone dropdown.
Added functions to update cq or itu zone from the dxcc table. Note that this means multi cq or itu zones will have the default zone. This is not a 100% fix, but at least it will help out some what.
image


int2001 (approved):
works.

Checks i did:
- Search empty CQ // Fix them
- Search empty ITU // Fix them
- Map
- Lowering/Raising limit

PR #2037, opened 2025-06-17
Amsat only accept 6 char-grids, so trunc them

int2001:
Repro:

try to enter a 8char-grid manually on the amsat-page. result is: 

image


phl0 (approved):
Ooops. Never tested with longer grids ... :)

PR #2038, opened 2025-06-17
[Search] Added button for opening QSOs in LBA

AndreasK79:
image


phl0 (approved):
works for me. But in map with >5k QSOs to be handled separately.

PR #2040, opened 2025-06-18
[Advanced Logbook] Added shift+click to checkboxes

AndreasK79:
Based on #2036, so that one needs to be merged first.


int2001 (approved):
works now.

tested:
- Shift-Click --> Map
- Shift-Click --> Globe
- some filters
- delesecting via SHIFT-Click (when evverything was checked)

PR #2041, opened 2025-06-19
This one converts the IDs posted for map/glibe to JSON

int2001:
This one converts the marked QSOs from "url-form" to a JSON-Object before posting it to the geoJSON-Handler.

prevents problems, where one runs into a PHP-Limit exceeding max amounts of params


AndreasK79 (approved):
Fixes the parameter issue in PHP!

PR #2044, opened 2025-06-20
Fixes an issue with shift-click and a js-error on 1st load of LBA

int2001:
Under rare circumstances, the LBA isn't working properly when it is called (SHIFT-Click-Feature introduced in #2040 )
it throws this one:

image

with this patch the error is catched, and a fallback to a simple draw (which enables the shift-click and prevents the error) is done.


AndreasK79 (approved):
I can't reproduce, and notice no difference with or without, so I guess it does not hurt.

PR #2049, opened 2025-06-22
Add DXCC flags to LBA view

phl0:
As per request (which is not documented to github yet) ...


int2001 (approved):
Works.

Tested:
- DXCCs without Flag
- DXCCs with Flag
- Map
- Search for DXCC (Filter)
- Sorting by DXCC (Asc / Desc)

PR #2055, opened 2025-06-23
fix wrong table field to COL_QSLSDATE

DB4SCW:
When one would use the function "Mark selected QSOs as sent" button in the qslprint view, it would correctly mark the QSOs as sent, but it would modify the timestamp of QSL received, not QSL sent.

This fixes it to use the right field in the database.


HB9HIL (approved):
Good find! Thx!

PR #2056, opened 2025-06-23
DXCC ADIF-Export and Flag-Fix

int2001:
Fixes the following issue:

~~1. Export to ADIF (from) contained always DXCC-ID from MY_DXCC instead of QSO-Partner~~
2. Flag wasn't properly updated/re-rendered when sending QSL's, fixing Zones, etc.


phl0 (commented):
Could not reproduce an/the error regarding wrong MY_DXCC in ADIF export from LBA.
Re-rendering flags upon edt / QSL stuff works. Tnx fix.

PR #2061, opened 2025-06-25
Slash some more zeros and adapt search functions

phl0:
Apply slashed zeros and data table search function in following pages:

- hams.at
- WWFF Award
- POTA Award
- SIG Award
- VUCC Award


phl0:
Causes more issues than benefit. I.e. searching for non-slashed zero stuff like grids or park ids does not work anymore.

PR #2089, opened 2025-07-01
Separate table ids

phl0:
Calling the quick lookup function from the Logbook overview and then showing some details popup for listing filtered QSOs results in the Logbook overview table being applied the data table stuff as well as to the table popping up. This is due to both tables being loaded from the same PHP and thus having the same ID.

The bug can be reproduced as follows:

- Go to Logbook -> Overview
- Call the quick lookup function with ALT-L
- Lookup a zone e.g. 14
- Press "Show"
- Click on the details for a mode/band combination e.g. 20m FT8

Result:

image

Selecting some other mode / band even makes it worse as the DT stuff is applied again:

image

And on the third try even resulting in a user-visible error message:

image

This fixes the bug by adding a variable to the view which sets a different id for the table if it is loaded as a popup so that the DT code only applies to the table shown in the popup.

### DIFF
--- a/application/controllers/Awards.php
+++ b/application/controllers/Awards.php
@@ -474,6 +474,7 @@ class Awards extends CI_Controller {
 		if (!empty($qsltype)) {
 			$data['filter'] .= __(" and ").implode('/', $qsltype);
 		}
+		$data['ispopup'] = true;
 		$this->load->view('awards/details', $data);
 	}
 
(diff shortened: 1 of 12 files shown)


AndreasK79:
Fix works :)

AndreasK79:
Seems the other parts, that were affected by the fix, are now fixed too :)

PR #2091, opened 2025-07-01
Refactor LoTW users update and prevent table from being emptied if file cannot be downloaded or is corrupt

phl0:
This refactors the LoTW users update code to prevent the table from being emptied if LoTW is down or the file empty. Also switched to cURL here to have a slightly better error handling.


DK9JC:
tested and works

AndreasK79:
This does prevent the table from being blanked, so a good fix if LoTW should be unavailable or doing maintenance.

AndreasK79 (approved):

PR #2095, opened 2025-07-02
[Locations] Add last qso date

AndreasK79:
Now showing last QSO date in the locations table in station setup.


int2001 (approved):
tested:
- empty station_loc: works
- delete station_loc: works
- copy station_loc: works
- sorting: works
- linking: works
- last_qso functionality: works

phl0 (changes requested):
Code itself works. But sorting table by last QSO needs a fix:

image

PR #2098, opened 2025-07-02
[Station setup] Filter for active locations

AndreasK79:
image


phl0 (approved):
Tested and works fine for me.

PR #2099, opened 2025-07-02
Multi orbit

phl0:
Based on an idea by @dl3jop we should allow for multiple orbit types to be selected:

image


int2001 (approved):
tested --> works

PR #2103, opened 2025-07-03
Dxcluster mode

AndreasK79:
image

- [x] Bandedges table migration added.
- [x] Code uses default table to determine mode.
- [x] Added view for showing bandedges.
- [x] Editing works
- [x] Add fences for editing
- [x] Create new band edge works

Might need to add more bandedges in the migration script.


phl0 (approved):
Tested and running fine on my dev instance with some checks on various bands and modes.

phl0 (approved):
Works on my end. Tested various (edge) cases in bandedge editor.

PR #2104, opened 2025-07-03
Add SAT summary to table

phl0:
This adds another tab to the summary table on QSO logging to show which satellites you have worked the DX station of. Base list ist all worked satellites in the active logbook.

image


AndreasK79 (approved):

AndreasK79:
Nice additions for sat ops!

PR #2105, opened 2025-07-03
[Contesting] Display a dialog to confirm new session

AndreasK79:
Added a confirmation dialog so that you don't reset session by accident:
image

Just a nice to have thing, since this button was mentioned too many times during the weekend :p


phl0 (approved):
Good idea. Works.

HB9HIL:
I like!

PR #2111, opened 2025-07-05
[Antennastats] Fixed always loading js for multiselect

AndreasK79:
A bit unfortunate adding of a js file meant that multiselect is always loaded on each page because of a document ready.
image

Removed loading of js-file and the file itself.
Moved the functions to antennastats.js, which is already loaded on the page.

@phl0 give it a test when you can.


int2001 (approved):
removes error

phl0 (approved):
Works purrfect

phl0 (approved):

PR #2122, opened 2025-07-08
Use displayname incase LoTW name is not set (yet)

phl0:
We added TEVEL2-3 which has no LoTW name (yet).
Current:

Screenshot 2025-07-08 110224

Patch:

Screenshot 2025-07-08 113126


AndreasK79:
Tested working :)

PR #2123, opened 2025-07-08
[Station location] Autofill ITU zone

AndreasK79:
This is a small one.
Since we autofill CQ zone, I added so that ITU zone is also autofilled.


phl0 (approved):
Works. Small but very nice addition :D

PR #2128, opened 2025-07-09
Preselect conf type based on user_default_confirmation

phl0:
Would make sense to pre-select confirmation type based on user setting. Also auto-executes the search on page loading.


phl0:
Drafted until https://github.com/wavelog/wavelog/pull/2127 is done.

PR #2134, opened 2025-07-10
Fixes the "select-all" bug

int2001:
When filtering things within LBA or simply pressing the loupe (without filter set), the select-all checkbox wasn't working.
this one should(!) fix it


phl0 (approved):
Yeah. Brings back the mark all function to my LBA :)

PR #2051, opened 2025-06-23
Use user setting's `user_date_format` for `qso_start_date` field

Cryolitia:

PR #2150, opened 2025-07-14
[QSO entry] Custom date format

AndreasK79:
Based on the work and the idea by @Cryolitia in #2051.

I have decided to split this into two parts.
The first part handles the QSO entry.
The second part (and second PR) will handle contesting.

QSO entry now uses the date format set by the user

I think I have managed to fix the problems in the original PR.
The frontend was working, but the date format was not handled in the backend, and also made a lookup function fail.

As far as I can tell, it is only QSO entry that sends in the date to logbook/json.

- [x] Live QSO works
- [x] Post QSO needs fixing
- [x] Reset time in post QSO sets date to incorrect format


HB9HIL:
There is some issue.. When entering a manual date in POST QSO this date will be replaced by 01.01.1970
Didn't dig into the code yet. But testing fails here

AndreasK79:
> There is some issue.. When entering a manual date in POST QSO this date will be replaced by 01.01.1970 Didn't dig into the code yet. But testing fails here

Yeah, I was so focused on getting formats like Jul 20, 2025 and Jul 20, 25 to work, that I forgot to go back and re-test when I got those working. Seems it's all because of the lookup that is being done by url and passing parameters like this:
`const url = ${base_url}index.php/logbook/json/${find_callsign}/${json_band}/${json_mode}/${stationProfile}/${startDate}/${last_qsos_count};`

That does not work well with / in the date and const startDate = encodeURIComponent($('#start_date').val());

This is another rabbit hole...

PR #2154, opened 2025-07-15
Make satellite flightpath webpage header translated

ludoux:
Before:
图片

After:
图片

Although `Satellites` and `Satellite pass` in this code file didn't use gettext before, these page use gettext of a hardcode text seperately in their code files (see example below), which make these translated at the end.

https://github.com/wavelog/wavelog/blob/76779b95b69726dd4b5bb9e70a3aaf9d0b71e3ff/application/views/satellite/flightpath.php#L36
https://github.com/wavelog/wavelog/blob/76779b95b69726dd4b5bb9e70a3aaf9d0b71e3ff/application/views/satellite/pass.php#L6

Obviously there are multiple solutions to this simple problem, and this pr is just one of them :)

### DIFF
--- a/application/controllers/Satellite.php
+++ b/application/controllers/Satellite.php
@@ -52,7 +52,7 @@ class Satellite extends CI_Controller {
 		];
 
 		// Render Page
-		$pageData['page_title'] = "Satellites";
+		$pageData['page_title'] = __("Satellites");
 		$this->load->view('interface_assets/header', $pageData);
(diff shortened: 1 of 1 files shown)


HB9HIL:
@ludoux since this is merged now the ci/cd pipeline will make the strings available at https://translate.wavelog.org in a few minutes. Go and translate them if something is missing :)

PR #2169, opened 2025-07-18
use regex... just in case

HB9HIL:
for reasons

### DIFF
--- a/application/controllers/Lotw.php
+++ b/application/controllers/Lotw.php
@@ -320,9 +320,7 @@ class Lotw extends CI_Controller {
 					}
 				}
 
-				$pos = strpos($result, "<!-- .UPL. accepted -->");
-
-				if ($pos === false) {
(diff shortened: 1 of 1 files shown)


application/controllers/Lotw.php
```diff
@@ -320,9 +320,7 @@ public function lotw_upload() {
 					}
 				}
 
-				$pos = strpos($result, "<!-- .UPL. accepted -->");
-
-				if ($pos === false) {
+				if (!preg_match('/<!-- \.UPL\.\s+accepted -->/', $result)) {
```
int2001:
```suggestion
				if (!preg_match('/<!-- \.UPL\.\s*accepted -->/', $result)) {
```

tested on my prod - works

int2001 (changes requested):
see suggestion.

more flexible

int2001:
Let's wait for the extra change Lotw is planning to rollout tomorrow

PR #2176, opened 2025-07-20
[QRZ callbook] Add logging if http code not 200

AndreasK79:
Added a logging message if the http code returned is not 200.


AndreasK79:
@int2001 tried debugging, and as far as I can tell, calling curl_getinfo twice does work.

PR #2177, opened 2025-07-20
Exclude no satname

int2001:
Collector-PR for:

SAT-QSOs without a SAT-Name are invalid (SAT)-QSOs, they don't count.
this one is a collector for all stats/awards/views, which show them


int2001:
No progress and no interest here. closing it.

PR #2183, opened 2025-07-22
better use site_url() instead base_url() to include index page (if set)

HB9HIL:
We should use `site_url()` instead `base_url()` in some places (especially when the URL is displayed in the UI) since `site_url()` includes the index_page (index.php) if still set in config.php. Some users were facing issues when in their config.php the index page is still set to "index.php" (instead empty string) and they use this url without index.php for API calls. Depending on the webserver / reverse proxy configuration this can cause issues


int2001 (approved):
tested with:
- `$config['index_page'] = '';` - works
- `$config['index_page'] = 'index.php';` - works

PR #2188, opened 2025-07-23
New 404 Error Page

HB9HIL:
Thought we can give this some color and eyecandy.. even it's an error page

Before

image

After

image

This only affects the function `show_404('message')` which was broken anyway (we added a message but it was never display :-)

It behaves the same like the login page or also the header
Theme is chosen based on global options (if not logged in) and Logo is chosen based on theme

### DIFF
--- a/application/views/errors/html/error_404.php
+++ b/application/views/errors/html/error_404.php
@@ -3,8 +3,6 @@
 <title>404 Page Not Found</title>
 	<!-- Bootstrap CSS -->
 	<?php
-	$CI =& get_instance();
-	$theme = $CI->optionslib->get_theme();
 	if ($theme) { ?>
 		<link rel="stylesheet" href="<?php echo base_url(); ?>assets/css/bootstrap-multiselect.css">
 		<link rel="stylesheet" href="<?php echo base_url(); ?>assets/css/<?php echo $theme; ?>/bootstrap.min.css">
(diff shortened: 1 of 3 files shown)


AndreasK79:
@HB9HIL seems translation it not available here:

> An uncaught Exception was encountered
> Type: Error
> 
> Message: Call to undefined function __()
> 
> Filename: C:\xampp\htdocs\wavelog\system\core\Exceptions.php
> 
> Line Number: 135

HB9HIL:
> @HB9HIL seems translation it not available here:
> 
> > An uncaught Exception was encountered
> > Type: Error
> > Message: Call to undefined function __()
> > Filename: C:\xampp\htdocs\wavelog\system\core\Exceptions.php
> > Line Number: 135

Strange. Here it worked. Will check  agn

phl0:
> @HB9HIL seems translation it not available here:
> 
> > An uncaught Exception was encountered
> > Type: Error
> > Message: Call to undefined function __()
> > Filename: C:\xampp\htdocs\wavelog\system\core\Exceptions.php
> > Line Number: 135

Same error on my dev system...

HB9HIL:
This is really tricky to solve. Reason is that the Exception Class within Codeigniter is loaded pretty early. Just before everything else is loaded. I'll need to find another solution. Closing here for now.

PR #2199, opened 2025-07-28
Fixes Az/Ele Calculation with respect to the user_date_format

int2001:
The individual date_format at QSO form which was introduced in #2150 breaks the Az/Ele calculation for the SAT Tab.
this one should fix it


AndreasK79 (approved):

AndreasK79:
Seems to work for me!

PR #2200, opened 2025-07-28
Fix _getLocator

tallcode:
image

There is a bug in the /gridmap page. The grid starting with A cannot display the worked/confirmed background color.

image

Because _getLocator returns the wrong grid name


int2001 (approved):
tested. good find.
to make sure, wait for @AndreasK79 approve, since the map is his domain ;)

tnx agn for contributing

AndreasK79:
@tallcode nice find! This has gone unnoticed for several years :)

AndreasK79 (approved):

Issue #2208, opened 2025-08-02
All cron jobs are failing


blhoward2:
**Describe the bug**
Last week, wavelog properly executed cron jobs uploading and downloading from LOTW, etc. Last night I realized they are now all failing when run from cron manager; however, they all work properly when manually synced through the Third Party Services menu.

I tried updating to the latest docker image, with the same result. 

With logging set to 4, no errors are logged:
```
root@cde81feccc8d:/var/www/html/application/logs# tail -f log*
INFO - 2025-08-02 16:18:01 --> Model "User_options_model" initialized
INFO - 2025-08-02 16:18:01 --> Model "Logbooks_model" initialized
INFO - 2025-08-02 16:18:01 --> Model "Stations" initialized
INFO - 2025-08-02 16:18:01 --> Model "Oqrs_model" initialized
INFO - 2025-08-02 16:18:01 --> Controller Class Initialized
INFO - 2025-08-02 16:18:01 --> Model "Cron_model" initialized
DEBUG - 2025-08-02 16:18:01 --> CRON: clublog_download is due and will be executed.
DEBUG - 2025-08-02 16:18:01 --> CRON: qrz_download is due and will be executed.
INFO - 2025-08-02 16:18:01 --> Final output sent to browser
INFO - 2025-08-02 16:18:01 --> Total execution time: 0.1548
```

**Version Information**
We need some information about your versions. You can get all Information in the Admin > Debug Menu
- OS:               Latest docker image
- PHP:             8.3.23
- MySQL/MariaDB:    11.3.2-MariaDB-1:11.3.2+maria~ubu2204
- Wavelog:          2.0.7 

**Steps to Reproduce**
1. Schedule a cron job to run
2. Job fails and is marked as failed in UI
3. No error messages logged
4. Manually running a sync for each does work and new QSOs are downloaded, etc.

**Expected Behavior**
Cron jobs should run as scheduled.

**Screenshots**

<img width="2597" height="621" alt="Image" src="https://github.com/user-attachments/assets/a6e2cbd6-055f-42a2-8fc8-41bd4384daed" />

**Desktop Information:**
- OS:               NixOS
- Browser:          Firefox
- Version:          140.0.2

**Additional Context**
Everything worked last week before I went on vacation. No changes, and it didn't work when I got home.

blhoward2:
Also, to be clear it is literally ALL cron jobs failing. I had every single one enabled and they all failed, from the version update check to QRZ, LOTW, etc.

[...]

PR #2214, opened 2025-08-03
Default localhost for cronmanager in docker environments

HB9HIL:
There is a logical issue since the beginning of the cronmanager implementation in https://github.com/wavelog/wavelog/pull/306

It only affects installations which are using docker and keep the base_url at localhost and a different port then 80. It's quite simple (e.g. `$config['base_url'] = 'http://localhost:8086/';`)

`localhost` inside the container !== `localhost` in the host machine. 

Reason is that we use port `8086` per default in our [docker-compose.yml](https://github.com/wavelog/wavelog/wiki/Installation-via-Docker#docker-composeyaml). This leads into the problem that the mastercron is correctly called inside the container with `http://localhost[:80]` ([see Dockerfile](https://github.com/wavelog/wavelog/blob/master/Dockerfile#L45)) but the cron/run method gets `http://localhost:8086` from the base_url config ([via local_url() method](https://github.com/wavelog/wavelog/blob/6704c35b1dea2b4539c1102af94a30fc668ae1cd/application/controllers/Cron.php#L95)). This obvisouly fails inside the container as in this little cute blue whale Wavelog is running on port `80`, and not `8086`.

There was a good solution around a year ago by @winnieXY in https://github.com/wavelog/wavelog/pull/795 but the people (and myself) forget about the config option. Therefore we use now `http://localhost[:80]` by default in docker environments which always works (because inside the container Wavelog always runs on localhost port 80). Thanks to @blhoward2 for poking me here. Even this is not a bug, it leads into a bad user experience when people running Wavelog in docker for the first time on their local machine and the cronmanager fails... 

Solves 
- https://github.com/wavelog/wavelog/issues/2208
- https://github.com/wavelog/wavelog/discussions/2180
- .. and probably some more

### DIFF
--- a/application/controllers/Cron.php
+++ b/application/controllers/Cron.php
@@ -92,7 +92,17 @@ class cron extends CI_Controller {
 						echo "CRON: " . $cron->id . " -> is due: " . $isdue_result . "\n";
 						echo "CRON: " . $cron->id . " -> RUNNING...\n";
 
-						$url = local_url() . $cron->function;
+						if (ENVIRONMENT == "docker") {
+							// In Docker, we use the localhost[:80] to call the cron directly inside the container
+							$url = 'http://localhost/' . $cron->function;
+							log_message('debug', 'Docker Environment detected. Using URL: ' . $url);
+						} else {
+							// In other environments, we use the local_url() function to get the default url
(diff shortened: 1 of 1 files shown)


int2001 (approved):
Good find

PR #2220, opened 2025-08-05
Added RST to LBA

int2001:
i personally see no real sense mass-editing RST at LBA, since those are QSO-individual values.
But as it was demanded and doesn't hurt - y not?


AndreasK79 (approved):

AndreasK79:
Tested working :)

PR #2223, opened 2025-08-07
Started on renaming logbook to container

AndreasK79:
Since the name "Station logbook" is confusing for many users, renaming it to container seems like a more intuitive thing.

image
image


EelcoD:
The term container is confusing for me as its refer to as something like container in a docker/podman environment

AndreasK79:
@EelcoD and people are confused by the station logbooks concept. Do you have any other suggestions?

iu2frl:
Upvote for `Station logbook`, because `Container` is too generic and might create even more confusion (it reminds me Docker containers 😄 )

PR #2229, opened 2025-08-12
Remove only qrgunits instead of killing all

int2001:
localStorage was killed completly, when calling QSO-Page (for QRGUnit-Reasons).
Instead of killing all, and saving single values, this one only kills the qrgunits.

Reason:
We're using localstorage for - e.g. - SAT-Favs. Everytime you call the qso-page the SAT-Favs were gone.
This fixes it.


int2001:
tested this one 1 week on my own prod without issues.
Any objection, @HB9HIL ?

Issue #2235, opened 2025-08-14
Missing "DOK" edit feature


int2001:
Add DOK to LBA-Massedit

### Discussed in https://github.com/wavelog/wavelog/discussions/2234

<div type='discussions-op-text'>

<sup>Originally posted by **dirkklossek-svg** August 14, 2025</sup>
Hi there, I wanted to edit some QSO´s to add DOK´s in it, but there is no "DOK" in the pulldown menu
<img width="320" height="930" alt="edit" src="https://github.com/user-attachments/assets/3c0c4dd2-898a-4650-b2e1-1ab57657a054" />
</div>

int2001:
done

PR #2236, opened 2025-08-14
Added DOK to LBA-Bulk-Editor

int2001:
Solves #2235


phl0 (changes requested):
We should enforce DOK to be upper-case:

https://github.com/int2001/wavelog/pull/19

phl0 (approved):
worx!

PR #2239, opened 2025-08-15
Change default RST on fav selection

phl0:
We should also set default RST according to mode on selection of a saved fav. Mode is set from fav but defautl RST not:

https://github.com/user-attachments/assets/61414f95-29e9-47d7-ace1-1e5267146510


int2001 (approved):
good finding. works

PR #2248, opened 2025-08-19
Show confirmation-Method at DXCC Award instead of simple "C"

int2001:
This one shows the confirmation-Method (if checked) instead of the "C".
so it's easy to see which DXCC was confirmed via QSL-Card but not via LoTW and so on.

image

open issues:
- [x] Get map back to work
- [x] Add Legend for QLEZC (**Q**SL-Card, **L**oTW, **E**QSL, QR**Z**, **C**lublog)


AndreasK79:
Like this feature. Much easier to see the confirmations if you are looking for a specific confirmation.

PR #2260, opened 2025-08-23
feat(simplefle): Add TIMEZONE support to FLE parser

xiaosuan2b:
### What does this PR do?
This pull request adds support for the `TIMEZONE` keyword to the Simple Fast Log Entry (FLE) parser.

### Why is it needed?
This feature allows users to input QSO times in their local timezone (e.g., `TIMEZONE +8` followed by `1400 BH6SKD`). The parser will then automatically convert the local time to the correct UTC time (`0600`) for logging. This greatly improves convenience for users who log in real-time but think in their local timezone.

### How to test it:
1.  Go to the SimpleFLE page.
2.  In the textarea, enter a timezone offset, for example: `TIMEZONE +8`.
3.  On a new line, enter a QSO with a time in that local timezone, for example: `1400 BH6SKD`.
4.  **Observe:** The QSO list on the right should display the time correctly converted to UTC (e.g., `0600`).
5.  Saving the QSO should also store the correct UTC time in the logbook.

---
*This feature was developed with the guidance of Gemini.* 😉

I'm a new contributor and this is my first pull request here. I'd appreciate it if you could review my code and point out any areas for improvement.

### DIFF
--- a/assets/js/sections/simplefle.js
+++ b/assets/js/sections/simplefle.js
@@ -486,8 +486,8 @@ function handleInput() {
 			}
 
 			const tableRow = $(`<tr>
-			<td>${extraQsoDate}</td>
-			<td>${qsotime}</td>
+			<td>${utcDate}</td>
+			<td>${utcTime}</td>
 			<td>${callsign}</td>
 			<td><span data-bs-toggle="tooltip" data-placement="left" title="${freq}">${band}</span></td>
 			<td>${mode}</td>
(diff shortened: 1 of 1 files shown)


HB9HIL (approved):
Can't spot any major issues. Would be good to see documentation for it in this wiki entry: https://github.com/wavelog/wavelog/wiki/SimpleFLE
everyone can edit it.

Thanks for contribution and welcome to Wavelog :)

HB9HIL:
Some of the team should review aswell please

int2001 (approved):
Works as designed.
tbh: it's an international hobby. everyone commited to log in UTC. however, if it helps others...

PR #2271, opened 2025-08-28
[Contesting] Add support for user defined date format

AndreasK79:
This PR adds support for user defined date format.


phl0 (approved):
This works for me. Tested several date formats and didn't encounter any issue.

PR #2282, opened 2025-09-03
Fixes DXCC in Search when no QSO is found and callbook returns nothing

DB4SCW:
This PR fixes something that has bugged me a long time.

When using QRZ without a subscription, a search for a callsign that existed, but was not yet worked, always resulted in 

``"DXCC: - NONE - (E.G. /MM, /AM)"``

which was almost always obviously wrong.

In the case of using QRZ free lookup, the callbook lookup returns empty string for the dxcc. 

This PR adds that in this case, we could try looking it up ourselves using our lookup function to display sensible information, while leaving the rest of the data intact.


int2001 (approved):
since both scenarios are tested: approve. good find

PR #2284, opened 2025-09-04
Add POTA ref import function

phl0:
This adds a functionality to enrich existing QSOs with POTA references from an ADIF file. There are tools like https://github.com/adamfast/pota-hunter-log-adif which allow generating ADIF files from a POTA hunter log.

### DIFF
--- a/application/controllers/Adif.php
+++ b/application/controllers/Adif.php
@@ -355,6 +355,74 @@ class adif extends CI_Controller {
 			$this->load->view('interface_assets/footer');
 		}
 	}
+
+	public function pota() {
+		$this->load->model('stations');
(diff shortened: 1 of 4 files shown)


int2001 (approved):
nice. tested it.

tbh: a bit complex to download the pota-hunter-file at pota (no WL issue), but once you got it, it's nice

phl0:
> tbh: a bit complex to download the pota-hunter-file at pota (no WL issue), but once you got it, it's nice

Totally agree. I asked POTA dev team several times to offer ADIF download directly but it was denied (argument: the logs are probabsly wrong and you should keep your own log). They didn't get the idea... or I was bad at explaining ... 🤷‍♂️

PR #2286, opened 2025-09-04
Prevent auto-usage from (probably) wrong qrz.com grids

phl0:
We should skip using a grid from qrz.com if it is obviously guessed and not set by the user but "calculated" from unknown information. Also we display a warning next to the grid in search results if this is the case.

Detection is based on XML tag called geoloc in qrz.com API. If this is set to "grid" it is wrong in almost every case. So we should not use it (automatically).

Screenshot 2025-09-04 162917

Screenshot 2025-09-04 162943


int2001 (approved):
tested and works.

tested with:
- calls i already wked (takes old grid)
- calls i never worked and callbook didn't provide grid (omits grid) 
- calls i never worked and callbook provides grid (takes grid)

PR #2297, opened 2025-09-09
Set proper last_run for DCL-Cron

int2001:
Copypaste of "set_last_run" from lotw wasn't working, since naming of the job was done "different". (My fault)
But since we already have a migration, this is less effort.

Tested it on my prod. Without patch, "last run" was never set (cosmetics)


phl0 (approved):
Works. Tested:
- cron call directly (i.e. curl)
- cron manager

PR #2306, opened 2025-09-12
Add option to display frequency and mark all QSOs for given QSL method in QSLPrint view

DB4SCW:
This PR adds 2 pieces of functionality to the qslprint view (+ a bit of cosmetics):

1) Allow the user to pick if he wants to see the band, the frequency or both values in the QSO table.
Reason: Some people use this view to hand-write QSL cards. If you have a frequency field on your QSL card instead of band, this saves you a click on each of the QSOs to view the frequency information. Default behaviour is still "band-only".

2) Allow the user to let wavelog mark all QSOs in the queue based on the QSL method. 
Reason: Event calls or club stations sometimes have a great amount of "QSL card debt". If - for example - you have 1000 QSOs to send, 400 Direct and 600 Bureau, now you can just let wavelog tag every QSO for the Bureau without clicking 400 times to remove all the "Direct" QSOs from the list. This way you can quickly export those QSOs to ADIF and set them as sent via the other buttons.

Finally, I made the labels for the select boxes a bit more pleasant to look at.

### DIFF
--- a/application/models/Qslprint_model.php
+++ b/application/models/Qslprint_model.php
@@ -70,7 +70,7 @@ class Qslprint_model extends CI_Model {
 	function get_qsos_for_print($station_id = 'All') {
 		$binding=[];
 		$binding[]=$this->session->userdata('user_id');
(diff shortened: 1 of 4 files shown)


AndreasK79 (approved):

AndreasK79:
@DB4SCW works. Nice addition for those bigger QSL jobs :)

PR #2307, opened 2025-09-13
Mark DCL-Confirmed QSOs as RCVD regardless of DOK

int2001:


phl0 (approved):
Works fine with me. Updated all confirmed QSOs in DCL as DCL_RCVD as expected. First run was rather slow because ~ 3.5k QSOs needed update. But quick now after initial import.

PR #2310, opened 2025-09-14
[QSL Queue] Fix the "Add QSO" feature

AndreasK79:
@DB4SCW FYI: I missed the problem with adding more QSOs to the list via the QSO list button. Since this inserts lines in the main table, we need to make sure the columns are correct.

I fixed this by always showing frequency in the dialog, and checking what is set in the dropdown when inserting into the main table.


DB4SCW:
somehow I overlooked that feature. good find.

Issue #2305, opened 2025-09-11
API to set callback for a configured radio


iu2frl:
**By submitting feature requests, you have the opportunity to contribute your ideas and inspiration to the Wavelog Dev-Team. Please be aware that not every feature request will be addressed or answered.**

**What Feature is Missing?**
The current API set does not provide a way to programmatically configure the endpoint of a radio

**Why is This Feature Important?**
I am creating a plugin for an SDR software which supports 4 independent receivers, and each receiver which is synched to Wavelog using the `api/radio` APIs should also be able to be tuned using the Wavelog bandmap. Altough manual configuration of the callbacks is possible, it is not the best user experience

**Is This Feature Personal or Beneficial to Others as Well?**
Probably yes as it might help supporting transceivers with multiple VFOs

int2001:
at Hardware Interfaces --> Edit you can adjust the callback-URL for each radio, which is used for QSYing.

perhaps i don't understand your issue. do you want to set this callback-URL via API?
The current feature with adjusting the callback is very rarely used, and tbh.: one has to set it once if it's needed (very special usecase).

iu2frl:
Yeah I know that one can edit the callbacks from such menu and I have no issues with that. 
I just think that radio interfaces are added to that list as soon as an API call is triggered, which is good, but the only thing that needs manual interaction is the callback URL. Maybe an additional parameter in the JSON API like "cat_url" (which I'm quite sure I saw that being output from WavelogGate when analyzing the protocol using Wireshark) might help saving the user a few clicks

int2001:
so its really the edgecase of the edgecase.
putting it on my list. but not sure if we're going to implement it.

Reason:
With your special-setup the Radio is created when it appears the first time. So it's a few more clicks which have to be done only ONE time. or do you own a radio-shop with new radios every day?

iu2frl:
Yes it's just an user commodity, not a ground breaking feature for sure. Low priority but a nice to have, it shouldn't be too hard to implement

int2001:
Great and happy to hear! Every contribution is welcome. Please add:
- sanitize checking of URL
- store it into user_options
- URL should be optional (please. take care of default)

Looking forward to your PR. Happy about everyone who helps.

iu2frl:
Completed in #2314, waiting for feedbacks

PR #2314, opened 2025-09-14
Adding support for optional `cat_url` argument in `api/radio` endpoint

iu2frl:
# Radio API Callback URL Enhancement

## Overview
The `/api/radio` endpoint has been enhanced to support an optional `cat_url` field that allows programmatic configuration of the CAT callback URL for each radio as proposed in #2305.

### 1. API Controller (Api.php)
- Added support for **optional** `cat_url` field in the JSON payload
- Implemented URL validation and sanitization
- Added `sanitize_cat_url()` method to ensure URLs are properly formatted

### 2. CAT Model (Cat.php)
- Enhanced the `update()` method to handle `cat_url` field
- Ensures callback URLs are stored in the database alongside other radio data

### JSON Payload Example

```json
{
  "key": "your_api_key_here",
  "radio": "FT-950",
  "frequency": 14075000,
  "mode": "SSB",
  "power": 100,
  "timestamp": "2025-09-14 16:47:00",
  "cat_url": "https://log.example.com:54321"
}
```

image

## Test script

This was the Python script I used to test the feature

```python
import requests
from datetime import datetime

WAVELOG_API_URL = "http://wavelog.localhost/index.php/api/radio"
WAVELOG_API_KEY = "aaBBccDDeeFFgg"
RADIO_NAME = "ICOM IC-7850"

# Define required headers
headers = {
    "Content-Type": "application/json",
    "Accept": "application/json"
}

payload = {
   "key": WAVELOG_API_KEY, 
   "radio": RADIO_NAME,
   "frequency": 14075000,
   "mode": "SSB",
   "power": '', # Optional field defined in watts
   "timestamp": datetime.now().strftime("%Y/%m/%d %H:%M"),
   "cat_url": "http://localhost:12345/meow", # Optional field
}

# Send the POST request
try:
    response = requests.post(WAVELOG_API_URL, json=payload, headers=headers)
    # Check if the request was successful
    if response.status_code == 200:
        print("Request successful!")
        print("Response:", response)
    else:
        print(f"Request failed with status code: {response.status_code}")
        print("Response:", response.text)
except requests.exceptions.RequestException as e:
    print("An error occurred:", e)
```

## Documentation

Wiki will be updated accordingly as soon as this PR gets approved

### DIFF
--- a/application/controllers/Api.php
+++ b/application/controllers/Api.php
@@ -655,7 +655,7 @@ class API extends CI_Controller {
 
 		// Handle optional cat_url
 		if (isset($obj['cat_url']) && !empty($obj['cat_url'])) {
-			$cat_url = $this->sanitize_callback_url($obj['cat_url']);
+			$cat_url = $this->sanitize_cat_url($obj['cat_url']);
 			if ($cat_url !== false) {
 				$obj['cat_url'] = $cat_url;
 			}
@@ -1138,7 +1138,7 @@ class API extends CI_Controller {
 	 * @param string $url The URL to sanitize
 	 * @return string|false Returns sanitized URL or false if invalid
 	 */
-	private function sanitize_callback_url($url) {
+	private function sanitize_cat_url($url) {
 		// Basic sanitization
 		$url = trim($url);
 		
--- a/application/controllers/Api.php
+++ b/application/controllers/Api.php
@@ -653,6 +653,13 @@ class API extends CI_Controller {
 				break;
 		}
 
+		// Handle optional cat_url
+		if (isset($obj['cat_url']) && !empty($obj['cat_url'])) {
+			$cat_url = $this->sanitize_callback_url($obj['cat_url']);
+			if ($cat_url !== false) {
+				$obj['cat_url'] = $cat_url;
+			}
+		}
 
 		// Store Result to Database
 		$this->cat->update($obj, $user_id, $operator);
@@ -1124,6 +1131,30 @@ class API extends CI_Controller {
 		// Return result
 		http_response_code(200);
 		echo json_encode(['status' => 'successful', 'message' => 'Export successful', 'statistics' => $data]);
+	}
+
+	/**
+	 * Sanitize and validate callback URL
+	 * @param string $url The URL to sanitize
+	 * @return string|false Returns sanitized URL or false if invalid
+	 */
+	private function sanitize_callback_url($url) {
+		// Basic sanitization
+		$url = trim($url);
+		
+		// Check if URL is valid and uses http or https
+		if (!filter_var($url, FILTER_VALIDATE_URL) || 
+			(!preg_match('/^https?:\/\//', $url))) {
+			return false;
+		}
+		
+		// Remove trailing slashes
+		$url = rtrim($url, '/');
+		
+		// Additional XSS cleaning
+		$url = $this->security->xss_clean($url);
+		
+		return $url;
 	}
 
 }
--- a/application/models/Cat.php
+++ b/application/models/Cat.php
@@ -27,6 +27,11 @@
 				'timestamp' => $timestamp,
 			);
 
+			// Handle callback URL if provided
+			if (isset($result['cat_url']) && !empty($result['cat_url'])) {
+				$data['cat_url'] = $result['cat_url'];
+			}
+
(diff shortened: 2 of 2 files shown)


int2001:
Nice. This looks really promising. Going to check it tomorrow.

int2001 (changes requested):
works. tested:
1. New Radio with URL-posted: url is sanitized and saved.
2. Existing Radio (default-URL) with URL in request: Radio-CB-URL is updated.
3. New radio without URL-posted: default url is taken
4. Existing Radio (default-URL) without URL in request: stored Radio-URL is taken.

thinking loud about "2". if someone uses an app which updates the CB-URL, it may be that other apps aren't working any longer. And the "normal"-user doesn't know where to look at.

would you mind to remove the update thing? or other better ideas here? perhaps adding a string to the radio-name, like "_specialURL", to make it visible for the user?

iu2frl:
Hmm yes, I see your point. I think this is more a UX/UI issue. Fabian once emailed me about this "callback" thing and it took me some good time to find it in the "hardware interfaces" tab under the "edit" button. I think it is more a "discovery" issue perhaps? Maybe just need to document it better? I reworked the Wiki last week to add some notes as there wasn't literally anything about that

int2001:
history shows:
- people are not reading.
- if there's a button, people click the button (regardless of captions / warnings)
- people always ask, before they read.

tbh: would it be a problem to add the appname or some hint to the radio-name, if it was altered by URL? the radio-name is visible at many places . the edit is - indeed - a bit hidden.

i'd appreciate it, because during the last-time we were flooded with tickets about obvious things. everything which helps to reduce-support-load is helpful.

iu2frl:
I'm not sure, it would look ugly to me, even because that radio name is displayed ok the widget... Maybe have a "decorator" if the callback URL is not the default? But not sure how that should look like

int2001:
i am also fine with a decorator. some UTF8-Char which is displayed // concatenated to the radio-name.

i fear the people who use your app which alters the URL. atm they need to alter it "manually". but with your PR this can happen in Background. No one will see it. and as already mentioned - people are installing / trying out things, without knowing what's happening.

99% of the user use the Gate which runs on localhost. they're not aware of "callback-URLs" and so on.

and tbh: even if your app/3rd-party apps show a warning: it's not under control of Wavelog. we're going to be flooded with support-tickets, not the 3rd-party app.

iu2frl:
Ok, what do you think about this? This would:

- Need no changes in the radio name
- Be a small non-impacting change in visuals or functionalities
- Offer a quick way to restore default behavior

This was just a mockup done on Paint with my smartphone, there's surely space for improvements. Better tooltip content strings are welcome.

image

int2001:
nah, this doesn't help. it needs to be part of the radio-description.

imagine a "normal" user who tried your app or an app which alters the default-URL.
The User would raise a ticket: "Cluster doesn't work" - in best-case with a screenshot of the cluster (where radio-name is displayed).

so he'd see on first sight (or one of us): "Ah, altered URL", instead of investigating over and over. and tbh: in a year no one remembers of this edge-edge-feature.

just want to reduce work for all of us.

what's so bad with adding a hint to the radio-name?!

iu2frl:
Because changing the radio name will make other people unhappy as they want to call it "radio" while it ends up like "radio_custom" (or whatever string we choose), it just look bad to me. The same "ugly" name will also appear in the radio public widgets. As soon as that user with non-working cluster goes to the radio list, they will see that icon and at least see why only that one has the info button

iu2frl:
In addition, the same issue would appear if the user manually tampers with the callback URL and then forgets about it. Today's there's no clue whatsoever about some kind of manipulation of that string

int2001:
ok. i tried it. hopefully not everyone uses the feature. and we could also add something like that what you call "ugly" later, if there are tickets about it.

tnx for contribution.

iu2frl:
@int2001 I don't want to convince you, I am just trying to find together the better solution with as few complains from end users as possible

int2001:
don't want to argue with you. all okay.

my point was: Cluster-functionality could now be "destroyed" by 3rd party-apps, without letting the user know it. not really happy about it. but lets see what future brings us.

iu2frl:
Sure, I totally see your point. I was just wondering if there's any way to make it easier to debug, even if the user manually tampers with that string for whatever reason

int2001:
if he manually alters it, he knows about it.
therefore the idea of the concatination. however, let's wait for at least one other WL-Team-Member and then we merge this beauty.

PR #2323, opened 2025-09-16
Allow LoTW confirmations including CNTY for some DXCCs

phl0:
LoTW confirmations only include STATE and CNTY for U.S., China and Canada. This aims to also add the other DXCCs which are handled in tqsl. Code has been taken from tqsl sources and adapted accordingly.

This does not include UI patches to set the CNTY part correctly. E.g. for Japan the user would have to enter the number according to https://adif.org.uk/316/ADIF_316.htm#Secondary_Administrative_Subdivision. This should be handled in a separate PR.

Also this fixes some code that compares country names as strings to decide which LoTW cert tot take. This was fixed to use the official ADIF ID instead.

### DIFF
--- a/application/controllers/Lotw.php
+++ b/application/controllers/Lotw.php
@@ -1122,6 +1122,23 @@ class Lotw extends CI_Controller {
 	}
 
 	/*
+	|	Function: lotw_ru_oblast_map
+	|	Requires: russian oblast map $ru_oblast
+	*/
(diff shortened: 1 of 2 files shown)


int2001 (approved):
sri 4 delay.
did a few blind-tests. seems to work

phl0:
Ok then lets see how this behaves in dev :)

PR #2327, opened 2025-09-17
Adds creationdate and modifydate to nearly every table

int2001:
This mig adds two columns to the most-volatile tables of wavelog.
* CREATION_DATE
* LAST_MODIFIED

it helps debugging and future logics

### DIFF
--- a/application/migrations/256_crea_modidates.php
+++ b/application/migrations/256_crea_modidates.php
@@ -22,11 +22,25 @@ class Migration_crea_modidates extends CI_Migration {
 	}
 
 	function add_create_modi($table) {
-			$this->dbtry("ALTER TABLE ".$table." ADD COLUMN `CREATION_DATE` TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,  ADD COLUMN `LAST_MODIFIED` TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP;");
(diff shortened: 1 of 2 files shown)


phl0 (changes requested):
Basically works but needs a few changes imho:
- tables label_types and paper_types already have a column named 'last_modified' thus throwing an error because the existence is not checked before
- downgrade does not work on MySQL servers because "IF EXISTS" is not supported. So no columns are removed
- 'last_modified' columns of the two tables mentioned before would get removed on downgrade although they should be kept because they existed before mig 256

HB9HIL:
@int2001 Just a question: And why the column descriptions in capital letters? Why not `last_modified` instead `LAST_MODIFIED`

int2001:
> @int2001 Just a question: And why the column descriptions in capital letters? Why not `last_modified` instead `LAST_MODIFIED`

just like to make clear of internal cols here. no special reason.

application/migrations/256_crea_modidates.php
```diff
@@ -0,0 +1,54 @@
...
+        	if (($col_check) && !(($table == 'paper_types') || ($table == 'label_types'))) {
+			$this->dbtry("ALTER TABLE ".$table." DROP COLUMN `last_modified`;");
+		}
+	}
+
+	function dbtry($what) {
+		try {
+			$this->db->query($what);
+		} catch (Exception $e) {
+			log_message("error", "Something gone wrong while altering the OQRS table: ".$e." // Executing: ".$this->db->last_query());
```
AndreasK79: Forgot hardcoded table name?

HB9HIL (approved):

PR #2334, opened 2025-09-22
Allow LBA edit function to set QSL sent/rcvd

phl0:
Personal feature request by DF5DH


phl0:
Oh wait, we have buttons for those ....

PR #2339, opened 2025-09-22
HAMqsl integration & Solar/Propagation information can be displayed at the dashboard

szporwolik:
This is initial integration of the HAMqsl Solar and propagation data.

- cronjob is added to pull the XML from HAMqsl every hour (this is done by migration)
- new model for HAMsql was added and handles basic functions
- user can set if the propagation box shall be visible at the dashboard via user profile edition (@int2001 / @phl0 per comment in the code, this will require your review?)
- the basic propagation is displayed at the dashboard, below statistics
- core idea was to have propagation data in the app, this can be reused in various places

Error handling was implemented as well as it was tested at my staging server.

### DIFF
--- a/application/views/dashboard/index.php
+++ b/application/views/dashboard/index.php
@@ -355,13 +355,13 @@ function getDistance($distance) {
 			<tr>
 				<td width="50%"><?= __("Sent"); ?></td>
 				<td width="25%"><?php echo $total_qsl_sent; ?></td>
-				<td width="25%"><a href="javascript:displayContacts('','All','All','All','All','QSLSDATE','');"><?php echo $qsl_sent_today; ?></a></td>
+				<td width="25%"><?php echo $qsl_sent_today != 0 ? "<a href=\"javascript:displayContacts('','All','All','All','All','QSLSDATE','');\">".$qsl_sent_today."</a>" : "0"; ?></td>
 			</tr>
 
 			<tr>
 				<td width="50%"><?= __("Received"); ?></td>
 				<td width="25%"><?php echo $total_qsl_rcvd; ?></td>
-				<td width="25%"><a href="javascript:displayContacts('','All','All','All','All','QSLRDATE','');"><?php echo $qsl_rcvd_today; ?></a></td>
+				<td width="25%"><?php echo $qsl_rcvd_today != 0 ? "<a href=\"javascript:displayContacts('','All','All','All','All','QSLRDATE','');\">".$qsl_rcvd_today."</a>" : "0"; ?></td>
 			</tr>
 
 			<tr>
@@ -382,13 +382,13 @@ function getDistance($distance) {
 			<tr>
 				<td width="50%"><?= __("Sent"); ?></td>
 				<td width="25%"><?php echo $total_lotw_sent; ?></td>
-				<td width="25%"><a href="javascript:displayContacts('','all','all','All','All','LOTWSDATE','');"><?php echo $lotw_sent_today; ?></a></td>
+				<td width="25%"><?php echo $lotw_sent_today != 0 ? "<a href=\"javascript:displayContacts('','all','all','All','All','LOTWSDATE','');\">".$lotw_sent_today."</a>" : "0"; ?></td>
 			</tr>
 
 			<tr>
 				<td width="50%"><?= __("Received"); ?></td>
 				<td width="25%"><?php echo $total_lotw_rcvd; ?></td>
-				<td width="25%"><a href="javascript:displayContacts('','all','all','All','All','LOTWRDATE','');"><?php echo $lotw_rcvd_today; ?></a></td>
+				<td width="25%"><?php echo $lotw_rcvd_today != 0 ? "<a href=\"javascript:displayContacts('','all','all','All','All','LOTWRDATE','');\">".$lotw_rcvd_today."</a>" : "0"; ?></td>
 			</tr>
 		</table>
 		<?php } ?>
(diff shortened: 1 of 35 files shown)


application/migrations/257_add_cron_hamqsl.php
```diff
@@ -0,0 +1,31 @@
+<?php
+
+defined('BASEPATH') or exit('No direct script access allowed');
+class Migration_add_cron_hamqsl extends CI_Migration {
+
+	public function up() {
+		if ($this->db->table_exists('cron')) {
```
int2001, line 8: Please add a check - or at least a `try/error`-Block - if cron-entry already exists.
without check it fails at the migration (it's just a safety net)

application/views/dashboard/index.php
```diff
@@ -456,6 +456,76 @@ function getDistance($distance) {
 
     </table>
     <?php } ?>
+
+	<?php if($solar_bandconditions && $solar_solardata){ ?>
```
int2001: Please add a safety-net as well here. if `$solar_bandconditions` isn't populated (because cron didn't run), PHP throws an error here.

```suggestion
<?php if ((($solar_bandconditions ?? '') != '') && (($solar_solardata ?? '') != '')){ ?>
```

image

application/models/Update_model.php
```diff
@@ -177,6 +177,29 @@ function wwff() {
         }
     }
 
+	function hamqsl(){
+		// This downloads and store hamqsl propagation data XML file
+
+		$contents = file_get_contents('https://www.hamqsl.com/solarxml.php', true);
```
int2001: Please use `curl` with timeout instead of `file_get_contents`.

i know - we've several places with `file_get_contents`, but it's deprecated, so at least new functions should do it with curl

application/views/user/edit.php
```diff
@@ -636,6 +636,16 @@
 											</select>
 											<small id="user_dashboard_banner_Help" class="form-text text-muted"><?= __("This allows to disable the global notification banner on the dashboard."); ?></small>
 										</div>
+
+										<div class="mb-3">
+											<label for="user_dashboard_solar"><?= __("Dashboard solar and propagation data"); ?></label>
+											<?php if(!isset($user_dashboard_solar)) { $user_dashboard_solar='Y'; }?>
+											<select class="form-select" id="user_dashboard_solar" name="user_dashboard_solar" aria-describedby="user_dashboard_solar_Help" required>
+												<option value='true' <?php if($user_dashboard_solar == "true") { echo "selected=\"selected\""; } ?>><?= __("Enabled"); ?></option>
+												<option value='false' <?php if($user_dashboard_solar == "false") { echo "selected=\"selected\""; } ?>><?= __("Disabled"); ?></option>
```
int2001: i'd appreciate if default could be `false` here.

One of the Waevlog-Mantras is: keep it simple and don't overwhelm the user with informations. if someone needs the prop-data, he could enable it.

application/views/dashboard/index.php
```diff
@@ -456,6 +456,76 @@ function getDistance($distance) {
...
+		<!-- Solar Data -->
+		<table class="table table-striped border-top">
+			<tr class="titles">
```
HB9HIL, line 463: Good morning

Credits are needed and important. As Opensource project we absolutely know that and N0NBH absolutely deserves it. 

..But this source description is too large and too obvious in my/our eyes. Please place the source description below the header as small italic textline. You can use my suggestion if you like.


```suggestion
			<tr class="titles">
				<td colspan="10" style="display: flex; justify-content: space-between; align-items: center;">
					<div>
						<div><i class="fas fa-sun"></i> <?= __("Solar Data & Propagation"); ?></div>
						<div class="small fst-italic text-muted">
							<?= sprintf(__("Last update at %s."), $solar_solardata['updated']); ?>
						</div>
					</div>
					<a class="ms-2 text-body fas fa-info-circle float-end" 
						data-bs-toggle="tooltip" 
						data-bs-placement="top" 
						data-bs-html="true"
						href="https://www.hamqsl.com/"
						target="_blank"
						title="<?= __("Data provided by HAMqsl."); ?>" 
						style="cursor: pointer;">
					</a>
				</td>
			</tr>
```

application/views/dashboard/index.php
```diff
@@ -456,6 +456,76 @@ function getDistance($distance) {
...
+							<td><?= $solar_solardata['aurora'] ?></td>
+						</tr>
+					</table>
+				</td>
+			</tr>
+			<tr>
+				<td colspan="10" class="small">
+					<?= sprintf(__("Data provided by %s."), '<a target="_blank" class="text-primary fw-bold text-decoration-underline" href="https://www.hamqsl.com/">'.'HAMqsl'.'</a>'); ?>
+					<?= sprintf(__("Last update at %s."), $solar_solardata['updated']); ?>
+				</td>
```
HB9HIL: and this needs to be removed. see comment above

application/views/dashboard/index.php
HB9HIL: grafik

szporwolik:
Thank you for the review, I have modified the PR per your comments.

Please excuse the bugs - I've to admit that I was not sure how to the migrations work, adding error handling was a definetly a good idea ;-) . There was in fact state cousing an variable not being defined when user switched off the widget in the settings, this shall be now fixed.

int2001:
Great. Could recheck earliest tonight

application/migrations/257_add_cron_hamqsl.php
```diff
@@ -0,0 +1,39 @@
...
+			if ($query->num_rows() == 0) {
+				// Insert the cron job only if it does not exist
+				$this->db->insert_batch('cron', $data);
+			}
+		}
+	}
+
+	public function down() {
+
+		$this->dbtry("delete from cron where id = 'update_update_hamqsl';");
```
phl0:
```suggestion
		$this->db->query("delete from cron where id = 'update_update_hamqsl';");
```

The dbtry function is not included and this the code fails here:

image


Imho it is sufficient to just use db->query here as the code will not fail even if 0 rows are affected.

application/models/Update_model.php
```diff
@@ -177,6 +177,37 @@ function wwff() {
...
+	function hamqsl(){
+		// This downloads and stores hamqsl propagation data XML file
+
+		$url = 'https://www.hamqsl.com/solarxml.php';
+		$ch = curl_init();
+		curl_setopt($ch, CURLOPT_URL, $url);
+		curl_setopt($ch, CURLOPT_HEADER, false);
+		curl_setopt($ch, CURLOPT_RETURNTRANSFER, true);
+		curl_setopt($ch, CURLOPT_USERAGENT, 'Wavelog Updater');
+		$contents = curl_exec($ch);
```
int2001, line 190: good one. but still missing a proper timeout. please add a timeout of 5sec. here

szporwolik:
Hope this shall be now fixed - per default the solar data widget shall be hidden. If there is no option, the default value shall be now set it to be hidden.

phl0:
> Hope this shall be now fixed - per default the solar data widget shall be hidden. If there is no option, the default value shall be now set it to be hidden.

Not yet. I did propose a change to the migration (down) because it will fail as dbtry does not exist. Please see my review with suggestion.

szporwolik:
> > Hope this shall be now fixed - per default the solar data widget shall be hidden. If there is no option, the default value shall be now set it to be hidden.
> 
> Not yet. I did propose a change to the migration (down) because it will fail as dbtry does not exist. Please see my review with suggestion.

Fixed, and also per your earlier comment, I have checked user creation - this option gets automatically set to disabled (hidden).

szporwolik:
I'm new here, what's the usuall workflow - after having this approved shall I Update or Update with rebase ?
PR initiators are doing that or the code approvers? 

Thanks! :)

phl0:
No worries. I handled it :) PR is merged :)

PR #2340, opened 2025-09-23
Allow LBA edit function to set QSL sent/rcvd

phl0:
Reviving https://github.com/wavelog/wavelog/pull/2334 as the actions under actions do not allow for QSL to be set to (I)nvalid and all other QSL methods are there anyway. So would not harm to have QSL in the edit function as well.


AndreasK79:
I agree. It doesn't hurt!

PR #2342, opened 2025-09-23
Show VUCC_GRIDS in case GRIDSQUARE is empty

phl0:
The LoTW analysis page should show VUCC_GRIDS in report if GRIDSQUARE is empty. There is a bug that if GRIDSQUARE is null VUCC_GRIDS are not shown.

Before:
image

After:
image

This fix shows the GRID/GRIDS which are incoming from LoTW and the log is updated with.


int2001 (approved):
good find

PR #2344, opened 2025-09-24
Make Wizard work on very very old computers with very very small screen

int2001:
Wizard should now even work on ancient devices down to 769 x 461 px

image


phl0 (approved):
Makes handling on 1024x768 screen in Internet Exploder a lot easier.

PR #2345, opened 2025-09-24
Suppress hyperlinks for zero cards

phl0:
Just don't display hyperlinks in case numbers are 0 (would cause an empty popup anyway).


int2001 (approved):
nice

PR #2346, opened 2025-09-24
derive wked/cnfm badge out of DXCC-ID instead of string

int2001:
when working a DXCC with special-characters (like `Trinidad & Tobago` - spot the `&`), the badge-update fails because of special-chars in URL.

This one changes the lookup to the more reliable DXCC-ID at two places:
- upon entering call in QSO-Form
- upon changing DXCC in QSO-Form

### DIFF
--- a/assets/js/sections/qso.js
+++ b/assets/js/sections/qso.js
@@ -776,7 +776,7 @@ function changebadge(entityval) {
 			}
 		})
 	} else {
-		$.getJSON(base_url + 'index.php/logbook/jsonlookupdxcc/' + convert_case(entityval) + '/0/' + $("#band").val() + '/' + $("#mode").val(), function (result) {
+		$.getJSON(base_url + 'index.php/logbook/jsonlookupdxcc/' + entityval + '/0/' + $("#band").val() + '/' + $("#mode").val(), function (result) {
 			// Reset CSS values before updating
 			$('#callsign_info').removeClass("lotw_info_orange");
 			$('#callsign_info').removeClass("text-bg-secondary");


assets/js/sections/qso.js
```diff
@@ -753,9 +753,9 @@ $(document).on('change', 'input', function () {
 	}
 });
 
-function changebadge(entityname) {
+function changebadge(entityval) {
 	if ($("#sat_name").val() != "") {
-		$.getJSON(base_url + 'index.php/logbook/jsonlookupdxcc/' + convert_case(entityname) + '/SAT/0/0', function (result) {
+		$.getJSON(base_url + 'index.php/logbook/jsonlookupdxcc/' + convert_case(entityval) + '/SAT/0/0', function (result) {
```
phl0:
```suggestion
		$.getJSON(base_url + 'index.php/logbook/jsonlookupdxcc/' + entityval + '/SAT/0/0', function (result) {
```

Should we remove the convert_case here? Its just a number now ...

assets/js/sections/qso.js
```diff
@@ -776,7 +776,7 @@ function changebadge(entityname) {
 			}
 		})
 	} else {
-		$.getJSON(base_url + 'index.php/logbook/jsonlookupdxcc/' + convert_case(entityname) + '/0/' + $("#band").val() + '/' + $("#mode").val(), function (result) {
+		$.getJSON(base_url + 'index.php/logbook/jsonlookupdxcc/' + convert_case(entityval) + '/0/' + $("#band").val() + '/' + $("#mode").val(), function (result) {
```
phl0:
```suggestion
		$.getJSON(base_url + 'index.php/logbook/jsonlookupdxcc/' + entityval + '/0/' + $("#band").val() + '/' + $("#mode").val(), function (result) {
```
same here ...

int2001: should be okay now

phl0 (approved):

PR #2354, opened 2025-09-26
Fix wrong logic on solar data session value

phl0:
There a bug in solar data dashboard panel. The session value does not get populated properly because the logic was inverted in https://github.com/wavelog/wavelog/pull/2339/commits/487d52d1055bbb3af47ee5b9360c73d1011b1957 by @szporwolik without re-arranging the options to the ternary operator. This bug can happens if you logout and just re-login. The session parameter will be empty and solar data not shown (regardless of user's setting). Editing and saving user options makes the solar data appear because session data is updated properly in user edit functions.


int2001 (approved):
!!

PR #2355, opened 2025-09-27
Set last_run properly

int2001:
The solardata-cron introduced by @szporwolik, updated the solar-data, but forgot to set the last_run properly.
This one fixes it.

Background:
last_run was only set, when triggering it manual at debug-menue, but never when triggered by cron


HB9HIL (approved):

phl0 (approved):
Confirmed.

PR #2357, opened 2025-09-27
Set custom header (which can be processed by reverse proxy)

int2001:
Set custom Header (`X-Login-Status: failed`) on failed login.
This one can be processed by your Reverseproxy to block/ban/tarpit a possible attacker.

Possible haproxy-Config in backend could be (This example blocks the requester IP for 5mins if there were more than 10 failed requests either on API or loginpage)
```
backend wavelog
        mode http
	stick-table type ipv6 size 100k expire 5m store gpc0
	http-request track-sc0 src
	acl ip_blocked sc0_get_gpc0 gt 10
	http-request deny deny_status 429 if ip_blocked
	acl login_failed status 401
	acl header_failed res.hdr(X-Login-Status) -i "failed"
	http-response sc-inc-gpc0(0) if login_failed or header_failed
        server local 127.0.0.1:81 check maxconn 250
```


phl0 (approved):
Worx

PR #2358, opened 2025-09-28
[ADIF] Added support for version 3.1.6

AndreasK79:


int2001 (approved):
works:
tried:
- upgrade
- downgrade

application/migrations/258_adif_3_1_6.php
```diff
@@ -0,0 +1,122 @@
...
+			'BANGGAI-DX',
+			'BEKASI-MERDEKA-CONTEST',
+			'ORARI-DX'
+		);
+
+		$this->db->where_in('adifname', $contest_names);
+		$this->db->delete('contest');
+	}
+
+	function dbtry($what) {
```
int2001: never used?!

application/migrations/258_adif_3_1_6.php
```diff
@@ -0,0 +1,122 @@
...
+				$this->db->insert('contest', $contest);
+			}
+		}
+
+		$table_name = $this->config->item('table_name');
+		$qso_fields = [];
+
+		$col_check = $this->db->query("SHOW COLUMNS FROM `$table_name` LIKE 'COL_EQSL_AG';")->num_rows() > 0;
+		if (!$col_check) {
+			$qso_fields[] = "ALTER TABLE `$table_name` ADD COLUMN COL_EQSL_AG VARCHAR(10) DEFAULT NULL AFTER COL_EQSL_STATUS;";
```
int2001, line 60: remove on downgrade?

int2001 (changes requested):

application/migrations/258_adif_3_1_6.php
AndreasK79: Since 3.1.5 didn't do that, I didn't either. And @phl0 remembered we discussed about not doing it.

phl0 (approved):
Works. We leave the EQSL_AG column in for the sake of not deleting user data upon downgrade.

PR #2359, opened 2025-09-28
[DXCC Award] Added new filter style

AndreasK79:
image


phl0 (approved):
Nice. Works as expected now.

PR #2362, opened 2025-09-29
Notes overhaul: updated view of the notes, added "Contacts" category with required logic, first step for callsign sticky notes

szporwolik:
## Summary
This PR introduces a full overhaul of the Notes section in Wavelog:
- Added an advanced view with search, pagination, and quick actions
- Improved layouts for note view, edit, and add pages
- Introduced a new note category: Contacts. This is the first step toward the much-requested feature of attaching notes to specific callsigns ([discussion #1653](https://github.com/wavelog/wavelog/discussions/1653))

With per-callsign notes in place, we’ll later be able to show an icon/link on QSO pages, making it easy for users to view and update their private notes.

## Developer Notes
This is a larger PR. After my previous Solar Data widget PR, I tried to be extra careful with testing. Still, since I’m fairly new to the Wavelog codebase, there may be some module interactions I didn’t catch.
I’d appreciate feedback and support from the Wavelog core dev team. Thanks in advance for your patience and all the support!

notes


application/controllers/Notes.php
```diff
@@ -1,119 +1,324 @@
...
+            $data['page_title'] = __("Add Notes");
+            $this->load->view('interface_assets/header', $data);
+            $this->load->view('notes/add', $data);
+            $this->load->view('interface_assets/footer');
+        } else {
+            // Ensure title is uppercase for Contacts before saving
+            $cat = $this->input->post('category', TRUE);
+            $title = $this->input->post('title', TRUE);
+            if ($cat === 'Contacts') {
+                $_POST['title'] = strtoupper($title);
```
HB9HIL: Good evening
just scrolled through and noticed this little security issue. Will go through this PR tomorrow in more detail, even it's a draft yet. 

The issue here is that you request the POST data again - and you call it with the PHP native method which does not include any XSS cleaning. Therefore this could be a vulnerability.

Simply call the variable instead, you already defined it in Ln 64



More tomorrow 👍😄
```suggestion
                $title = strtoupper($title);
```

application/controllers/Notes.php
```diff
@@ -1,119 +1,324 @@
...
-	{
-		parent::__construct();
+	// Main notes page: lists notes and categories
+    public function index() {
+        $this->load->model('note');
+        $data = [];
+        // Get all notes for logged-in user
+        $data['notes'] = $this->note->list_all();
+        // Get possible categories
+        $data['categories'] = Note::$possible_categories;
```
HB9HIL: This introduces OOP into the codebase. While it's okay for me and correct at this point, the team should agree here since we basically don't use OOP in this codebase.

application/controllers/Notes.php
```diff
@@ -1,119 +1,324 @@
...
+        foreach ($data['categories'] as $cat) {
+            $category_counts[$cat] = $this->note->count_by_category($cat);
+        }
+        $data['category_counts'] = $category_counts;
+        // Get total notes count
+        $data['all_notes_count'] = $this->note->count_by_category();
+        $data['page_title'] = __("Notes");
+        // Render views
+        $this->load->view('interface_assets/header', $data);
+        $this->load->view('notes/main', $data);
```
HB9HIL: No need to add $data to the main view aswell. It's already in the header and should therefore also available in the main view. Please remove and confirm that it still works


```suggestion
        $this->load->view('notes/main');
```

application/controllers/Notes.php
```diff
@@ -1,119 +1,324 @@
...
+            $cat = $this->input->post('category', TRUE);
+            if ($cat === 'Contacts') {
+                $this->load->helper('callsign');
+                $parsed = parse_callsign($this->input->post('title', TRUE));
+                $suggested_title = strtoupper($parsed['core']);
+            }
+            $data['suggested_title'] = $suggested_title;
+            $data['page_title'] = __("Add Notes");
+            $this->load->view('interface_assets/header', $data);
+            $this->load->view('notes/add', $data);
```
HB9HIL:
```suggestion
            $this->load->view('notes/add');
```

application/controllers/Notes.php
```diff
@@ -1,119 +1,324 @@
...
+            $cat = $this->input->post('category', TRUE);
+            if ($cat === 'Contacts') {
+                $this->load->helper('callsign');
+                $parsed = parse_callsign($this->input->post('title', TRUE));
+                $suggested_title = strtoupper($parsed['core']);
+            }
+            $data['suggested_title'] = $suggested_title;
+            $data['page_title'] = __("Edit Note");
+            $this->load->view('interface_assets/header', $data);
+            $this->load->view('notes/edit', $data);
```
HB9HIL:
```suggestion
            $this->load->view('notes/edit');
```

application/controllers/Notes.php
```diff
@@ -1,119 +1,324 @@
...
+            $this->load->view('interface_assets/footer');
+        } else {
+            // Ensure title is uppercase for Contacts before saving
+            $cat = $this->input->post('category', TRUE);
+            $title = $this->input->post('title', TRUE);
+            if ($cat === 'Contacts') {
+                $_POST['title'] = strtoupper($title);
```
HB9HIL:
```suggestion
                $title = strtoupper($title);
```

application/controllers/Notes.php
```diff
@@ -1,119 +1,324 @@
...
-	}
-	
-	/* Delete Note */
-	function delete($id) {
+	// API endpoint to check for duplicate note title in category for user
+	public function check_duplicate() {
+        $user_id = $this->session->userdata('user_id');
+        $cat = $this->security->xss_clean($this->input->get('cat', TRUE));
+        $title = $this->security->xss_clean($this->input->get('title', TRUE));
+        $id = $this->security->xss_clean($this->input->get('id', TRUE)); // Optional, for edit
```
HB9HIL: The TRUE parameter in the get() method already does XSS cleaning.

```suggestion
        $cat = $this->input->get('cat', TRUE);
        $title = $this->input->get('title', TRUE);
        $id = $this->input->get('id', TRUE); // Optional, for edit
```

application/controllers/Notes.php
```diff
@@ -1,119 +1,324 @@
...
+            $parsed = parse_callsign($title);
+            $core = strtoupper($parsed['core']);
+            // Only fail if prefix or suffix is present
+            if ($parsed['prefix'] || $parsed['suffix']) {
+                $this->form_validation->set_message('contacts_title_unique', 'Contacts note title must be a callsign only, without prefix/suffix. Suggested: ' . $core);
+                $_POST['title'] = $core;
+                return FALSE;
+            }
+            // Accept and normalize casing
+            $_POST['title'] = $core;
```
HB9HIL: This method has `$title` as GET parameter (Ln 265). Why calling it here from POST? And if this is correct the it should be `$this->input->post('title', TRUE)`. But as mentioned. This needs to be checked why it's a GET parameter for the method, but is called from POST here.

application/controllers/Notes.php
```diff
@@ -1,119 +1,324 @@
...
+                'title' => $core
+            ]);
+            foreach ($query->result() as $note) {
+                if ($note->id != $note_id) {
+                    $this->form_validation->set_message('contacts_title_unique_edit', 'A note with this callsign already exists in your Contacts. Please enter a unique callsign.');
+                    return FALSE;
+                }
+            }
+            $_POST['title'] = $core;
+            return TRUE;
```
HB9HIL: Same issue here. '$title' is GET parameter for the method, but called (insecure) from POST.

application/controllers/Notes.php
szporwolik: Thanks  :smile: ! Good catch, addressed in [3c175c4](https://github.com/wavelog/wavelog/pull/2362/commits/3c175c44e8f707e461528176d31eccdeb842be1a).

Changed to pass all of those as function arguments to the  model.

application/helpers/callsign_helper.php
```diff
@@ -0,0 +1,39 @@
+<?php
```
HB9HIL: This callsign helper is redundant code. We already have returns the plain callsign. 

https://github.com/wavelog/wavelog/blob/dev/application/libraries/Callbook.php#L171-L187

We want to reduce redundant code

HB9HIL: I just noticed even this method was twice in the codebase.. 

https://github.com/wavelog/wavelog/pull/2363

the callbook lib would be the way to go for you here. as mentioned above

application/models/Note.php
```diff
@@ -12,66 +20,208 @@ function list_all($api_key = null) {
...
+		$this->db->where('user_id', $user_id);
+		$this->db->group_by('cat');
+		$query = $this->db->get('notes');
+		$result = [];
+		foreach ($query->result() as $row) {
+			$result[$row->cat] = (int)$row->count;
+		}
+		return $result;
+	}
+
```
HB9HIL, line 164: We had just a discussion in the Core Team wherever its better to use the Codeigniter SQL Syntax or plain SQL language. We came to the conclusion that plain SQL syntax would be better as it's portable for future development and we have more control in terms of performance and security. I think it would be better to go with plain SQL language here. It's not wrong, more about strategic decisions. :)

application/views/notes/add.php
```diff
@@ -1,52 +1,60 @@
...
+        </div>
+        <div class="mb-3">
+          <label for="catSelect">
+            <?= __("Category"); ?>
+            <span class="ms-1" data-bs-toggle="tooltip" title="<?= __("Contacts") . __(" is a special note category used in various places of Wavelog to store information about QSO partners. This notes are private and are not shared with other users nor exported to external services.") ?>">
+              <i class="fa fa-question-circle text-info"></i>
+            </span>
+          </label>
+          <select name="category" class="form-select" id="catSelect">
+            <?php foreach (Note::$possible_categories as $cat): ?>
```
HB9HIL: A general comment: You use a lot of `$cat` as variable. Keep in mind that this ccause issues since the same variable is used for CAT control. I'd rather go with `$category`. Better safe then sorry :-D

application/views/notes/edit.php
```diff
@@ -1,57 +1,71 @@
...
+      <form method="post" action="<?php echo site_url('notes/edit'); ?>/<?php echo $id; ?>" name="notes_add" id="notes_add">
+        <div class="mb-3">
+          <label for="inputTitle"><?= __("Title"); ?></label>
+          <input type="text" name="title" class="form-control" value="<?php echo isset($suggested_title) && $suggested_title ? $suggested_title : set_value('title', $row->title); ?>" id="inputTitle">
+        </div>
+        <div class="mb-3">
+          <label for="catSelect">
+            <?= __("Category"); ?>
+            <?php if ($row->cat == 'Contacts'): ?>
+              <span class="ms-1" data-bs-toggle="tooltip" title="<?= __("Contacts") . __(" is a special note category used in various places of Wavelog to store information about QSO partners. This notes are private and are not shared with other users nor exported to external services.") ?>">
```
HB9HIL: In translations it's not a good idea to rip sentences apart. In different languages the grammar of the rest of the sentence depends on the main context. Make one sentence


```suggestion
              <span class="ms-1" data-bs-toggle="tooltip" title="<?= __("Contacts is a special note category used in various places of Wavelog to store information about QSO partners. This notes are private and are not shared with other users nor exported to external services.") ?>">
```

application/views/notes/edit.php
```diff
@@ -1,57 +1,71 @@
...
+                <i class="fa fa-question-circle text-info"></i>
+              </span>
+            <?php endif; ?>
+          </label>
+          <select name="category" class="form-select" id="catSelect">
+              <?php foreach (Note::$possible_categories as $cat): ?>
+                <option value="<?= htmlspecialchars($cat) ?>"<?= (set_value('category', $row->cat) == $cat ? ' selected="selected"' : '') ?>><?= __($cat); ?></option>
```
HB9HIL: You can't translate variables ->  `<?= __($cat); ?>`

You need to find another solution.

application/views/notes/add.php
```diff
@@ -1,52 +1,60 @@
...
+      <!-- Note creation form -->
+      <form method="post" action="<?php echo site_url('notes/add'); ?>" name="notes_add" id="notes_add">
+        <div class="mb-3">
+          <label for="inputTitle"><?= __("Title"); ?></label>
+          <input type="text" name="title" class="form-control" id="inputTitle" value="<?php echo isset($suggested_title) && $suggested_title ? $suggested_title : set_value('title'); ?>">
+        </div>
+        <div class="mb-3">
+          <label for="catSelect">
+            <?= __("Category"); ?>
+            <span class="ms-1" data-bs-toggle="tooltip" title="<?= __("Contacts") . __(" is a special note category used in various places of Wavelog to store information about QSO partners. This notes are private and are not shared with other users nor exported to external services.") ?>">
```
HB9HIL: Same as in application/views/notes/edit.php Line 43. Don't rip sentences apart


```suggestion
            <span class="ms-1" data-bs-toggle="tooltip" title="<?= __("Contacts is a special note category used in various places of Wavelog to store information about QSO partners. This notes are private and are not shared with other users nor exported to external services.") ?>">
```

application/views/notes/add.php
```diff
@@ -1,52 +1,60 @@
...
+              <i class="fa fa-question-circle text-info"></i>
+            </span>
+          </label>
+          <select name="category" class="form-select" id="catSelect">
+            <?php foreach (Note::$possible_categories as $cat): ?>
+              <option value="<?= htmlspecialchars($cat) ?>"<?= (set_value('category', 'General') == $cat ? ' selected="selected"' : '') ?>><?= __($cat); ?></option>
```
HB9HIL: This is not going to work `<?= __($cat); ?>`

application/views/notes/main.php
```diff
@@ -1,38 +1,94 @@
...
+					<!-- Category filter buttons -->
+					<div id="categoryButtons" class="btn-group me-2 flex-wrap" role="group" aria-label="Category Filter">
+						<button type="button" class="btn btn-sm btn-outline-secondary btn-light category-btn active" data-category="__all__">
+							<?= __("All Categories"); ?> <span class="badge bg-secondary"><?= $all_notes_count ?></span>
+						</button>
+						<?php foreach ($categories as $cat): ?>
+							<button type="button" class="btn btn-sm btn-outline-secondary btn-light category-btn" data-category="<?= htmlspecialchars($cat) ?>">
+								<?= __(htmlspecialchars($cat)); ?>
+								<?php if ($cat === 'Contacts'): ?>
+									<span class="ms-1" data-bs-toggle="tooltip" title="<?= __("Contacts") . __(" is a special note category used in various parts of Wavelog to store information about QSO partners. These notes are private and are neither shared with other users nor exported to external services.") ?>">
```
HB9HIL:
```suggestion
									<span class="ms-1" data-bs-toggle="tooltip" title="<?= __("Contacts is a special note category used in various parts of Wavelog to store information about QSO partners. These notes are private and are neither shared with other users nor exported to external services.") ?>">
```

application/views/notes/main.php
```diff
@@ -1,38 +1,94 @@
...
+						<a href="<?php echo site_url('notes/add'); ?>" class="btn btn-sm btn-success">
+							<i class="fas fa-plus"></i> <?= __("Create Note"); ?>
+						</a>
+					</div>
+				</div>
+				<div class="col-md-5 col-12">
+					<!-- Search box and reset button -->
+					<div class="input-group">
+						<input type="text" id="notesSearchBox" class="form-control form-control-sm" maxlength="50" placeholder="<?= __('Search notes (min. 3 chars)') ?>">
+						<button class="btn btn-outline-secondary btn-sm btn-light" id="notesSearchReset" type="button" title="<?= __('Reset search') ?>">
```
HB9HIL: Please always use double quotes in translations

```suggestion
						<input type="text" id="notesSearchBox" class="form-control form-control-sm" maxlength="50" placeholder="<?= __("Search notes (min. 3 chars)") ?>">
						<button class="btn btn-outline-secondary btn-sm btn-light" id="notesSearchReset" type="button" title="<?= __("Reset search") ?>">
```

application/views/notes/main.php
```diff
@@ -1,38 +1,94 @@
...
+							});
+							// Initialize Bootstrap tooltips for note titles
+							var tooltipTriggerList = [].slice.call(document.querySelectorAll('#notesTable a[data-bs-toggle="tooltip"]'));
+							tooltipTriggerList.forEach(function(el) {
+								new bootstrap.Tooltip(el);
+							});
+						});
+					// Pass browser timezone to JS
+					window.browserTimezone = Intl.DateTimeFormat().resolvedOptions().timeZone;
+				</script>
```
HB9HIL, line 66: This JS should probably go into a dedicated file which is loaded on this view/page

application/views/notes/view.php
```diff
@@ -1,26 +1,54 @@
...
+        </ul>
+      </div>
+      <div class="card-body">
+        <div class="row mb-3">
+          <div class="col-md-8">
+            <div class="mb-2">
+              <span style="font-size:1em;">
+                <?= __("Category"); ?>: <?= __($row->cat); ?>
+                <?php if ($row->cat == 'Contacts'): ?>
+                  <span class="ms-1" data-bs-toggle="tooltip" title="<?= __("Contacts") . __(" is a special note category used in various places of Wavelog to store information about QSO partners. This notes are private and are not shared with other users nor exported to external services.") ?>">
```
HB9HIL:
```suggestion
                  <span class="ms-1" data-bs-toggle="tooltip" title="<?= __("Contacts is a special note category used in various places of Wavelog to store information about QSO partners. This notes are private and are not shared with other users nor exported to external services.") ?>">
```

assets/js/sections/notes.js
```diff
@@ -1,68 +1,487 @@
...
+            li.appendChild(a);
+            ul.appendChild(li);
+        }
+        paginationContainer.appendChild(ul);
+    }
+
+    function renderNotesTable(data) {
+        var tbody = '';
+        if (data.length === 0) {
+            tbody = '<tr><td colspan="5" class="text-center text-muted">No notes were found.</td></tr>';
```
HB9HIL: "No notes were found" can be also translated. We have a little go around here since our  Gettex Implementation does not support Javascript. We place generic translations in the head of the footer (here: https://github.com/wavelog/wavelog/blob/7261290151912b25bc385b874f73c0a0d580be35/application/views/interface_assets/footer.php#L29-L58) and the simply use the javascript variable here. 


```suggestion
            tbody = '<tr><td colspan="5" class="text-center text-muted">' + lang_no_notes_found + '</td></tr>';
```

Add this in the mentiond footer section:

```javascript
var lang_no_notes_found = "<?= __("No notes were found"); ?>";
```

HB9HIL: PS: There are a lot of strings in JS which can be translated this way

assets/js/logbookhelpers.js
HB9HIL, line 1: I'm not that happy with this file. 

"logbookhelper" is not really generic. Maybe you can place this function (if really needed) in assets/js/sections/common.js

application/controllers/Notes.php
szporwolik: Wanted to have this defined in one place, can be moved to other place if needed; do we have a good place to define this  kind of constants?

application/controllers/Notes.php
szporwolik: OK, will clean this up. Went a little crazy with xss_ckean after your comment yesterday. ;-)

application/helpers/callsign_helper.php
szporwolik: Perfect! Will use this one then.

assets/js/logbookhelpers.js
szporwolik: It seems we can live without it, by doing all the checks in the backend. Will remove it.

application/controllers/Notes.php
HB9HIL: Well for me this is fine 😄👍 I just know our team 

@int2001 @AndreasK79 @phl0

int2001: Not the usual way, we're doing it. but still understandable. ok for me.

assets/js/logbookhelpers.js
HB9HIL: If you need it you can place it in common.js (which is loaded by default on each page). But less code is also no bad idea :) Less is more :D

application/controllers/Notes.php
HB9HIL: Whats the purpose of `$_POST['title'] = $core;` in Line 320?

application/views/notes/edit.php
szporwolik: Actually it works! The reason is that those strings were know to the translation module, I need to think where to define those.

Visual Studio Code

application/views/notes/add.php
szporwolik: See https://github.com/wavelog/wavelog/pull/2362#discussion_r2391711566

application/views/notes/edit.php
HB9HIL: Where are they defined? That this can display the strings is true. But our workflow requires them to be defined somewhere. Gettext just displays the translations, but does not translate by itself.

HB9HIL: Anyway.. We don't want to use variables in gettext methods. This breaks the standards which are everywhere in Wavelog. You won't find any other place. You will need to find another way :)

szporwolik: > Where are they defined? That this can display the strings is true. But our workflow requires them to be defined somewhere. Gettext just displays the translations, but does not translate by itself.

I used the same category as the current revision in master - without updating mo files and translation. So the strings were defined in the previous version, this is why only "Contacts" is not translated.

But, case is clear - I will handle that differently. 👌

HB9HIL: It's not wroong in general to use PHP variables in gettext (even it's not great pratice). But in our case it's a bit different. The CI/CD will run after each merge to `dev` branch. The script parses through all *.php files and searches for `__()`, `_ngettext()` and `_pgettext()`. From all the found strings it generates the .po files and pushes these changes to dev aswell. Afterwards it triggers the webhook to https://translate.wavelog.org

If you now use $variables in these methods, the shell script will not add those values to the po file (in some cases it threw errors in the past). therefore we decided to not use variables in gettext since it was not used that way anyway. So in the end we have a pretty good working CI/CD for translations which we don't wanna break or change :D 

In this case I'd assume to use an defined array and referencing on the array index to choose the correct value. The array can defined with classic __("string") syntax. 

What do you think?

assets/js/logbookhelpers.js
szporwolik: File was removed as the logic is now fully in the backend.

application/views/notes/edit.php
szporwolik: Thanks! Yes, the an array is a way to go.

application/controllers/Notes.php
szporwolik: Reworked completetly, took care of this $_POST references.

szporwolik:
@HB9HIL thank you for an outstanding review - I was not expecting, so deep and quick review. 👍🏻 

I think all items are now addressed, but this is not a small PR... 

The one thing which bothers me is the layout of the main table. I might be wrong, but in my eyes the whole project lacks consistency with the layouts/frontend - can you list a few lately modified pages (with tables! 😉), so I can base on ?

HB9HIL:
Good morning @szporwolik 

Just a question before I start with the second review-round. Is there a reason you coded the javascript in Vanilla rather then jquery? Every JS in Wavelog is used to be jquery (with some exceptions).

Cheers and have a nice day

application/views/notes/add.php
```diff
@@ -1,52 +1,60 @@
...
+        </li>
+      </ul>
+    </div>
+    <div class="card-body">
+      <!-- Show validation errors if any -->
+      <?php if (!empty(validation_errors())): ?>
+      <div class="alert alert-danger">
         <a class="btn-close" data-bs-dismiss="alert" title="close">x</a>
         <ul><?php echo (validation_errors('<li>', '</li>')); ?></ul>
+      </div>
```
phl0:
```suggestion
      <div class="alert alert-danger alert-dismissible fade show" role="alert" style="margin-top: 1rem;">
      <span class="badge text-bg-info"><?= __("Warning"); ?></span>
        <?php echo (validation_errors('<span>', '</span>')); ?>
        <button type="button" class="btn-close" data-bs-dismiss="alert" aria-label="Close"></button>
      </div>
```

The error banner looks a bit awkward. Tried to align it with the layout on dasbord. Before:

image

After:

image

phl0 (approved):

szporwolik:
> Good morning @szporwolik
> 
> Just a question before I start with the second review-round. Is there a reason you coded the javascript in Vanilla rather then jquery? Every JS in Wavelog is used to be jquery (with some exceptions).
> 
> Cheers and have a nice day

To be honest I started with vanillia as I'm currently involved in other projects using it and didn't give it a lot of thinking. 😉

What I will do is collect all those information and put those in the Wiki for myself and other contributors.

HB9HIL:
> The one thing which bothers me is the layout of the main table. I might be wrong, but in my eyes the whole project lacks consistency with the layouts/frontend - can you list a few lately modified pages (with tables! 😉), so I can base on ?

You asked for layout and style. If you want to stick with the usual style of Wavelog I'd recommend the StationSetup page. Here you have Tables and Buttons as we usually style them in Wavelog. However I like your design it's just a bit different. 

I'm not a good frontened developer but I see what you mean with "lacks consistency with the layouts/frontend" haha

application/controllers/Notes.php
```diff
@@ -1,119 +1,317 @@
 <?php if ( ! defined('BASEPATH')) exit('No direct script access allowed');
 
+// Notes controller: handles all note actions, with security and input validation
 class Notes extends CI_Controller {
+    // API endpoint: check for duplicate note title in category for user
+    // Ensure only authorized users can access Notes controller
+    function __construct() {
+        parent::__construct();
+        $this->load->model('user_model');
+        $this->load->library('callbook'); // Used for callsign parsing
```
HB9HIL: I don't think this should be loaded before the security check and in the constructor anyway. 
This library is used in 4 of 12 methods in this controller. It's more efficient to load it in each of that 4 methods separatly

szporwolik:
Thanks! Let's go with it then "as it is", if my time will allow, I hope to craft PR or two to address the frontent topics.

Also I hope you don't mind, but I put some of the feedback received here into https://github.com/wavelog/wavelog/wiki/Development-Notes

application/controllers/Notes.php
```diff
@@ -1,119 +1,317 @@
...
+        }
+        $this->note->delete($clean_id);
+        redirect('notes');
+    }
+
+    // Duplicate a note by ID for the logged-in user, appending #[timestamp] to title
+    public function duplicate($id = null) {
+        $this->load->model('note');
+        $clean_id = $this->security->xss_clean($id);
+        if (!is_numeric($clean_id)) {
```
HB9HIL: We could make these check persistent just as the others aswell


```suggestion
        if (!is_numeric($clean_id) || !$this->note->belongs_to_user($clean_id, $this->session->userdata('user_id'))) {
```

application/controllers/Notes.php
```diff
@@ -1,119 +1,317 @@
...
+
+    // Form validation callback for add: unique Contacts note title for user, only core callsign
+    public function contacts_title_unique($title = null) {
+        $category = $this->input->post('category', TRUE);
+        if ($category === 'Contacts') {
+            $user_id = $this->session->userdata('user_id');
+            $core = strtoupper($this->callbook->get_plaincall($title));
+            // Only fail if prefix or suffix is present
+            if (strtoupper($title) <> $core) {
+                $this->form_validation->set_message('contacts_title_unique', 'Contacts note title must be a callsign only, without prefix/suffix. Suggested: ' . $core);
```
HB9HIL: This string should be translated


```suggestion
                $this->form_validation->set_message('contacts_title_unique', sprintf(__("Contacts note title must be a callsign only, without prefix/suffix. Suggested: %s"), $core));
```

application/controllers/Notes.php
```diff
@@ -1,119 +1,317 @@
...
+                return FALSE;
+            }
+            $existing = $this->db->get_where('notes', [
+                'cat' => 'Contacts',
+                'user_id' => $user_id,
+                'title' => $core
+            ])->num_rows();
+            if ($existing > 0) {
+                $this->form_validation->set_message('contacts_title_unique', __('A note with this callsign already exists in your Contacts. Please enter a unique callsign.'));
```
HB9HIL: Please use only double quotes in translations. (Documentation: https://github.com/wavelog/wavelog/wiki/Translations)


```suggestion
                $this->form_validation->set_message('contacts_title_unique', __("A note with this callsign already exists in your Contacts. Please enter a unique callsign."));
```

application/controllers/Notes.php
```diff
@@ -1,119 +1,317 @@
...
+    // Form validation callback for edit: unique Contacts note title for user (ignore current note), only core callsign
+    public function contacts_title_unique_edit($title = null) {
+        $category = $this->input->post('category', TRUE);
+        if ($category === 'Contacts') {
+            $user_id = $this->session->userdata('user_id');
+            $note_id = $this->input->post('id', TRUE);
+            $core = strtoupper($this->callbook->get_plaincall($title));
+			// Only fail if prefix or suffix is present
+            if (strtoupper($title) <> $core) {
+                $this->form_validation->set_message('contacts_title_unique_edit', __('Contacts note title must be a callsign only, without prefix/suffix. Suggested: ' . $core));
```
HB9HIL: Please use `sprintf()` and `%s` placeholders for dynamic strings. Putting strings together with `.` is problematic for good translation quality.


```suggestion
                $this->form_validation->set_message('contacts_title_unique_edit', sprintf(__("Contacts note title must be a callsign only, without prefix/suffix. Suggested: %s"),$core));
```

application/controllers/Notes.php
```diff
@@ -1,119 +1,317 @@
...
+                return FALSE;
+            }
+            $query = $this->db->get_where('notes', [
+                'cat' => 'Contacts',
+                'user_id' => $user_id,
+                'title' => $core
+            ]);
+            foreach ($query->result() as $note) {
+                if ($note->id != $note_id) {
+                    $this->form_validation->set_message('contacts_title_unique_edit', __('A note with this callsign already exists in your Contacts. Please enter a unique callsign.'));
```
HB9HIL:
```suggestion
                    $this->form_validation->set_message('contacts_title_unique_edit', __("A note with this callsign already exists in your Contacts. Please enter a unique callsign."));
```

application/models/Note.php
```diff
@@ -1,9 +1,25 @@
 <?php
 
 class Note extends CI_Model {
+	// Get list of possible note categories with translations
+	public static function get_possible_categories() {
+		return [
+			'Contacts' => __('Contacts'), // QSO partner notes
```
HB9HIL:
```suggestion
			'Contacts' => __("Contacts"), // QSO partner notes
```

application/models/Note.php
```diff
@@ -1,9 +1,25 @@
...
+			'General' => __('General'),   // General notes
```
HB9HIL:
```suggestion
			'General' => __("General"),   // General notes
```

application/models/Note.php
```diff
@@ -1,9 +1,25 @@
...
+			'Antennas' => __('Antennas'), // Antenna-related notes
```
HB9HIL:
```suggestion
			'Antennas' => __("Antennas"), // Antenna-related notes
```

application/models/Note.php
```diff
@@ -1,9 +1,25 @@
...
+			'Satellites' => __('Satellites') // Satellite-related notes
```
HB9HIL:
```suggestion
			'Satellites' => __("Satellites") // Satellite-related notes
```

application/models/Note.php
```diff
@@ -12,64 +28,205 @@ function list_all($api_key = null) {
...
+	function add($category, $title, $content, $local_time = null) {
+		$user_id = $this->session->userdata('user_id');
+		$check_title = $title;
+		if ($category === 'Contacts') {
+			$check_title = strtoupper($title);
+		}
+		$sql = "SELECT COUNT(*) as count FROM notes WHERE cat = ? AND user_id = ? AND title = ?";
+		$check_result = $this->db->query($sql, array($category, $user_id, $check_title));
+		if ($check_result->row()->count > 0 && $category === 'Contacts') {
+			show_error('In Contacts category, the titles of the notes need to be unique.');
```
HB9HIL: Translate if it's not some kind of API output. Not sure here, so I may be wrong

```suggestion
			show_error(__("In Contacts category, the titles of the notes need to be unique."));
```

application/models/Note.php
```diff
@@ -12,64 +28,205 @@ function list_all($api_key = null) {
...
+		$user_id = $this->session->userdata('user_id');
+		$check_title = $title;
+		if ($category === 'Contacts') {
+			$check_title = strtoupper($title);
+		}
+		$check_sql = "SELECT id FROM notes WHERE cat = ? AND user_id = ? AND title = ?";
+		$check_result = $this->db->query($check_sql, array($category, $user_id, $check_title));
+		foreach ($check_result->result() as $note) {
+			if ($note->id != $note_id && $category === 'Contacts') {
+				show_error('In Contacts category, the titles of the notes need to be unique.');
```
HB9HIL:
```suggestion
			show_error(__("In Contacts category, the titles of the notes need to be unique."));
```

application/models/Note.php
```diff
@@ -12,64 +28,205 @@ function list_all($api_key = null) {
...
+				'note' => $row->note,
+				'creation_date' => $row->creation_date,
+				'last_modified' => $row->last_modified
+			];
+		}
+		return [
+			'notes' => $notes,
+			'total' => $total
+		];
 	}
```
HB9HIL, line 230: @AndreasK79 Can you review this one for us? This is too complex SQL for my head :-D You are better at this

HB9HIL (changes requested):
Looks pretty good so far! These changes are just small ones.

application/views/notes/edit.php
HB9HIL: To be fair: I have to appologize

I've used this technique by myself in the installer for the languages

https://github.com/wavelog/wavelog/blob/56497034b3e99b046300157a35015639ee2cefe5/install/index.php#L1035-L1039

Your current solution seems to work very well, the other option would have been okay aswell. I'm sorry, I was wrong here

application/models/Note.php
szporwolik: This is unlikely to be seen by anybody as it's the path not available from the GUI. But... in some very rare cases, this error my be needed. I would translate it.

application/views/notes/edit.php
szporwolik: No worries! 😄  Per my understanding this will work as long as the string to be translated is "known" in the translation database, so there shall be other place where the translation script will able to extract it and add to the translation list. 

So, it will work as long as this other place exist - I completely agree that this is a bit of "risky" approach. 😃

application/controllers/Notes.php
szporwolik: Agreed and changed.

szporwolik:
Project related question, do you guys usually expect commits to be squashed or it's OK to commit as it is?

Asking as this PR review generated a lot of commits 😉 :
image

HB9HIL:
Yeah I'll squash merge this one.

phl0 (approved):
works fine for me.

application/models/Note.php
AndreasK79: @HB9HIL yep, looks good to me. The necessary binding is done where needed.

Thumbs up @szporwolik!

HB9HIL:
Thank you very much for your contribution @szporwolik! well done!

PR #2367, opened 2025-10-01
Fence // Don't import QSOs without a band

int2001:
Skip QSOs which do not contain a band at import


HB9HIL:
If the record has a frequency we could calculate the band from it

int2001:
> If the record has a frequency we could calculate the band from it

that's correct. we're doing that already. this one catches if there's even not a frequency.
see codelines above.

HB9HIL (approved):
yep sorry .. my bad :)

PR #2368, opened 2025-10-02
API-QSO  Expose additional import  parameters to the API

RhinoLance:
This PR is in response to the discussion at https://github.com/wavelog/wavelog/discussions/2364

Additional optional parameters are now exposed buy the QSO API calls, defaulting to false if not provided:
"skipDuplicate"
"markClublog"
"markLotw"
"dxccAdif"
"markQrz"
"markEqsl"
"markHrd"
"markDcl"

As they're optional, and the defaults match the previously hard coded values, there is no breaking change to any existing API user.

This has been tested from within Logbook_model->import_bulk to ensure the correct values flow through the API to the model correctly.

**Prior API data**
```
{
    "key": STRING,
    "station_profile_id": INT,
    "type": STRING,
    "string": STRING
}
```

**Proposed API data**
```
{
    "key": STRING,
    "station_profile_id": INT,
    "type": STRING,
    "string": STRING
    "skipDuplicate": BOOLEAN, Optional, default=false,
    "markClublog": BOOLEAN, Optional, default=false,
    "markLotw": BOOLEAN, Optional, default=false,
    "dxccAdif": BOOLEAN, Optional, default=false,
    "markQrz": BOOLEAN, Optional, default=false,
    "markEqsl": BOOLEAN, Optional, default=false,
    "markHrd": BOOLEAN, Optional, default=false,
    "markDcl": BOOLEAN, Optional, default=false,
}
```

e.g.
```
curl -X POST https://example.com/api \
     -H "Content-Type: application/json" \
     -d '{
              "key": "{{ApiKey}}",
              "station_profile_id": 1,
              "type": "adif",
              "string": "<CALL:6>VK6ABC<QSO_DATE:8>20250916<TIME_ON:6>044044<FREQ:7>14245.0<MODE:3>USB<RST_SENT:2>42<RST_RCVD:2>41<NAME:5>Grant<COMMENT:7>Zig Zag<BAND:4>14.0<STATION_CALLSIGN:7>VK7TEST<EOR><CALL:6>VK3XYZ<QSO_DATE:8>20250916<TIME_ON:6>114306<FREQ:7>3532.47<MODE:2>CW<RST_SENT:3>439<RST_RCVD:3>599<NAME:5>Ketut<COMMENT:6>Keilor<BAND:3>3.5<STATION_CALLSIGN:7>VK7TEST<EOR>"
              "skipDuplicate": true,
              "markClublog": false,
              "markLotw": false,
              "dxccAdif": false,
              "markQrz": false,
              "markEqsl": false,
              "markHrd": false,
              "markDcl": false
          }'
```


int2001:
do i understand that correct? Those flags are part of the ADIF?
asking because you fetch it out of the `obj`.

edit: 4get my comment. they're not part of the adif. perhaps you should add some description / an example for a curl to the PR (and to the wiki after it has been merged)

application/controllers/Api.php
```diff
@@ -237,6 +236,24 @@ function qso($dryrun = false) {
 		$mystation=$this->stations->profile_clean($obj['station_profile_id']);
 		$mygrid=($mystation->station_gridsquare ?? '');
 
+		//check for optional parameters
+		$boolval_or_false = function (array $arr, $key, $default = false) {
```
int2001, line 240: Question: Why not as a standard-(private-)function within the controller?

application/controllers/Api.php
```diff
@@ -237,6 +236,24 @@ function qso($dryrun = false) {
...
+			return isset($arr[$key]) && is_bool($arr[$key]) 
+				? $arr[$key] 
+				: $default;
+		};
+		
+		$importOptions = [
+			'skipDuplicate' => $boolval_or_false($obj, 'skipDuplicate'),
```
int2001, line 247: skipDuplicate has a bit "weird" logic (due to history) in code (see https://github.com/wavelog/wavelog/blob/cf4bc5900d8c673805d51a193bffafb057906b44/application/models/Logbook_model.php#L4141 )
EMPTY OR true means: don't skip them (import dupes)
FILLED (with anything but true) means: skip them

we know - weird.
but as it was only internal during the past, this one could be really missleading when exposed to the API.
Please change your logic to sth. more understandable.

RhinoLance: I'm happy to make the change, however before I do, I'd like to put an argument forward not to.

As it stands, the checks use loose truthy evaluation for what really should be a boolean input.  I guess this was done to simplify evaluating input-form checkboxes.  However, currently only internal code, of which the dev team has control is able to exploit this loose truthy evaluation.  If at some point the code is refactored to more robust checking, there is no problem as the calling page can also be updated.

However, as soon as this loose evaluation is exposed via the API, refactoring has the potential to become a breaking change for API users.  The proposed change does not alter the internal truthy evaluation, it simply forces the API user to provide an actual boolean value.

Thoughts?

application/controllers/Api.php
RhinoLance: I made the call to do it that way for a couple of reasons:
1.  It's a tiny function with a simple logic check,
2.  It's only used in that one section of code, so by being defined there, it's more likely to be removed if that piece of code is refactored, and not left to languish, unloved and abandoned.

If as a convention the dev team doesn't want inline functions, I'm happy to extract it to the controller.

application/controllers/Api.php
int2001: Up to you.
Either you implement a logic, where the API-Endpoint exposes a clear logic instead of using the actual (indeed weird) internal logic which has been built for the checkboxes, 
OR you refactor the whole logic (much more effort and much more pitfalls) including the checkbox-handling.

For the OR i am looping in @AndreasK79 here.

Regarding the API-Logic and compatbility: In both cases make sure, that the current API works as before when omitting the new parameters.

RhinoLance: Reading your response, perhaps I misunderstood your request:_" ... Please change your logic to sth. more understandable."_  What is "sth."?

I read this as "please mirror the existing logic".  Instead, did you mean to get rid of the ternary operators in favour of regular if/else statements?

int2001: ok, to clarify:

current (only internal) logic says: 
- `skipDuplicates = true` means: IMPORT Dupes

This is okay for internal use (missleading there as well - but it is like it is).
When exposing this to a potential API-User it is simply wrong.

option a)
- you change **your** logic, that skipDuplicates REALLY means skipDuplicates, when using it via API (Default should be "don't import dupes")

option b)
- you change the **whole** logic, that skipDuplicates REALLY means skipDuplicates, when using it via API (Default should be "don't import dupes"). you also need to adjust/change the current logic at manual ADIF-Import where import_bulk is used. Keep in mind, that - as seen from userside - the behaviour for the user MUST be the same

option c)
- you close your PR. We - as WL-Team - are going to fix the internal logic (can't tell u when this will happen), and after this has been done, you can open a new PR

RhinoLance: OK.  Having looked at it more, I'm going to close this PR.  Getting the logic into better place is probably better suited to the dev team with a better knowledge of how it all fits together, and me changing the API to work with the current implementation is only going to increase the effort required and add another layer of confusion for that task.

PR #2369, opened 2025-10-02
Make PHP8 the minimum required version

phl0:
A few controllers make use of str_contains function which was introduced in PHP8. PHP7.4 systems will run into errors.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ https://demo.wavelog.org
 ### Recommended: Classic-LAMP-Stack:
 * Linux based Operating System (Windows stack may also work)
 * Apache (Nginx should work)
-* PHP Version 8.0 up to PHP 8.2 (PHP 8.3. in friendly Usertest)
+* PHP Version 8.0 up to PHP 8.2 (PHP 8.4. in friendly Usertest)
 * MySQL or MariaDB (MySQL 5.7 or higher // MariaDB 10.1 or higher)
 
 ### Alternative - Easy start with prebuilt Docker-image:
--- a/application/views/dashboard/index.php
+++ b/application/views/dashboard/index.php
@@ -108,7 +108,7 @@ function getDistance($distance) {
 
 	<?php if (version_compare(PHP_VERSION, '8.0.0') <= 0) { ?>
 		<div class="alert alert-danger alert-dismissible fade show" role="alert">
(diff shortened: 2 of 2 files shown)


int2001 (approved):
WL is already shipped with PHP8.3 when it comes to docker-installations

application/views/dashboard/index.php
```diff
@@ -108,7 +108,7 @@ function getDistance($distance) {
 
 	<?php if (version_compare(PHP_VERSION, '7.4.0') <= 0) { ?>
 		<div class="alert alert-danger alert-dismissible fade show" role="alert">
-			<?= __("You need to upgrade your PHP version. Minimum version is 7.4. Your version is") . ' ' . PHP_VERSION . '.';?>
+			<?= __("You need to upgrade your PHP version. Minimum version is 8.0. Your version is") . ' ' . PHP_VERSION . '.';?>
```
HB9HIL: Shall we make this string dynamic so we don't have to retranslate it in the future over and over again?


```suggestion
			<?= sprintf(__("You need to upgrade your PHP version. Minimum version is %s. Your version is: %s."), "8.0", PHP_VERSION); ?>
```

README.md
```diff
@@ -41,7 +41,7 @@ https://demo.wavelog.org
 ### Recommended: Classic-LAMP-Stack:
 * Linux based Operating System (Windows stack may also work)
 * Apache (Nginx should work)
-* PHP Version 7.4 up to PHP 8.2 (PHP 8.3. in friendly Usertest)
+* PHP Version 8.0 up to PHP 8.2 (PHP 8.3. in friendly Usertest)
```
HB9HIL: @phl0 @int2001 @AndreasK79 

Shall we already bump the friendly user test to 8.4? I already tested with 8.4 and couldn't find any issues.


```suggestion
* PHP Version 8.0 up to PHP 8.2 (PHP 8.4. in friendly Usertest)
```

HB9HIL (commented):
What do you think about my suggestions?

README.md
int2001: ok for me

phl0:
> What do you think about my suggestions?

Fine with me. Pushed.

HB9HIL:
thx :)

PR #2371, opened 2025-10-02
Reentered birds

phl0:
The code did not cope with SATs which have invalid or outdated TLEs. The satellites were still selectable in flightpath and pass predicition.
Fix:
- Clear all TLEs before (re-)importing them so that reentered birds have empty TLEs in the table
- Prevent pass predictions for satellites which have NULL TLEs (safety measure)
- Use the content of TLE column for deciding whether TLEs are present rather than the updated column.


int2001 (approved):
works. and is 8.4 compatible now.

PR #2373, opened 2025-10-03
Update README.md

arguser:
Announce Umbrel availability.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,6 +36,14 @@ https://demo.wavelog.org
   Username: demo  
   Password: demo
 
+## Now available on Umbrel
+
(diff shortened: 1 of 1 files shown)


HB9HIL:
Hello @arguser 

While we really appreciate your work by providing Wavelog as easy deployable application for the Umbrel OS, we want to keep neutrality also for other third party players. In addition to that we are not responsible in any way that the Umbrel Variant of Wavelog and it's installation process work properly. Therefore we can't add this to the Readme. 

But I added Umbrel on our Thirdparty Wiki Page here: https://github.com/wavelog/wavelog/wiki/Third-Party-Tools#third-party-platforms

Thanks for understanding

73 de HB9HIL

PR #2379, opened 2025-10-04
[OQRS] Fix an error if you selected all for results in OQRS view

AndreasK79:


int2001 (approved):
bug fixed (tested). tnx

PR #2382, opened 2025-10-05
Update contests list from WA7BNM Contest Calender - Cabrillo Names

goodspeed34:
Update contests list from WA7BNM Contest Calender so that contests found in Contest Calender could also be found in Contest Logging.

The newly added entries were deduplicated with a script and manually inspected.

### DIFF
--- a/application/config/migration.php
+++ b/application/config/migration.php
@@ -22,7 +22,7 @@ $config['migration_enabled'] = TRUE;
 |
 */
 
-$config['migration_version'] = 261;
+$config['migration_version'] = 262;
 
 /*
 |--------------------------------------------------------------------------


phl0 (changes requested):
This does not work because the existing migration will never be executed on existing installations. You would need to create a new one.

goodspeed34:
P.S. This doesn't confirm to the possible contest IDs defined in ADIF specification. But the problem is that, contest comes and goes. Some contests wouldn't make their way to the ADIF specification while the Cabrillo export uses the ADIF name... How do you think of this? @phl0

phl0 (approved):

phl0 (approved):

phl0:
Fixed and merged now. :) Sri for delay

PR #2383, opened 2025-10-05
[Advanced Logbook] Rewrote filters and actions to use dropdowns instead

AndreasK79:
Since we have those "toolbars" taking up so much space, I am trying a different approach.
These buttons are dropdown filters/command instead:
image

image

image

image

image


int2001 (approved):
can't spot an issue here. everything working wonderful. even on mobile. great.

PR #2386, opened 2025-10-06
feat(oqrs): Oqrs page detect browser language

xiaosuan2b:
### Description

This pull request introduces an automatic language detection feature for the public-facing OQRS and Log Search pages.

When a first-time, non-authenticated visitor accesses these pages, the application now detects their browser's preferred language (`Accept-Language` header). If the detected language is supported by Wavelog, the page will automatically switch to that language, significantly improving the user experience for international visitors.

A new option has been added to the User Profile page, allowing each user to enable or disable this feature for their public logbook pages. It is enabled by default.

### Implementation Details

-   A new helper function, `parse_accept_language`, has been added to `MY_language_helper.php` to parse the `Accept-Language` header according to RFC 2616.
-   A new private method, `_initialize_visitor_language`, has been added to the `Oqrs.php` controller.
-   This method contains robust logic to handle various server environments and potential race conditions with default cookies. It compares the browser's preferred language with the currently set language cookie.
-   If a language switch is necessary, it sets a new cookie with the application-specific `folder` name (e.g., `chinese_simplified`) and performs a one-time redirect to apply the change immediately.

While implementing this feature, I encountered several complex issues across different server environments (like URL rewriting and HTTPS proxies) and tried to provide a robust solution. Although the current implementation has been tested on my setup, I'm sure there is still room for improvement. I would be very grateful for any feedback or suggestions from the maintainers to help me better align with the project's standards. Thank you!
DE BH6SKD, 73

### DIFF
--- a/application/controllers/Oqrs.php
+++ b/application/controllers/Oqrs.php
@@ -380,7 +380,8 @@ class Oqrs extends CI_Controller {
 		$this->load->model('user_options_model');
 
 		if ($public_slug ?? '' != '') {
-			$user_id = $this->user_model->get_user_id_from_slug($public_slug);
+			$user = $this->user_model->get_by_slug($public_slug)->row();
+			$user_id = $user->user_id ?? null;
 
 			if ($user_id) {
 				$lang_setting = $this->user_options_model->get_options('oqrs', array('option_name' => 'oqrs_use_visitor_browser_language', 'option_key' => 'boolean'), $user_id)->row();
--- a/application/controllers/Oqrs.php
+++ b/application/controllers/Oqrs.php
@@ -371,37 +371,62 @@ class Oqrs extends CI_Controller {
 		echo json_encode(array('status' => 'success', 'message' => __("QSO match added successfully.")));
 	}
 	/**
-	 * Initializes the visitor's language based on OQRS user settings and browser headers.
+	 * Initializes the visitor's language and applies it immediately.
+	 *
+	 * This function checks for a language cookie. If not present, it detects the visitor's
+	 * browser language, sets a corresponding language cookie, and then performs a
+	 * redirect to the current URL. The redirect forces the browser to make a new
+	 * request with the new cookie, allowing the Gettext hook to apply the language
+	 * change on the initial visit.
 	 *
 	 * @param string|null $public_slug The public slug from the URL.
 	 */
 	private function _initialize_visitor_language($public_slug = NULL) {
-		$this->load->model('user_model');
+		// 获取 gettext cookie 的配置名称
+		$cookie_name = $this->config->item('gettext_cookie', 'gettext');
+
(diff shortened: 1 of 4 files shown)


HB9HIL:
Hello @xiaosuan2b 

I'm not sure if I get this to 100%. This PR brings a feature to disable the auto negotiation of the language?

Because currently the language of a public page is automatically set by the browsers language. And as I get this right you want be able to disable that so your page would be always displayed in chinese? 

Am I correct?

HNO3Miracle:
> Hello @xiaosuan2b
> 
> I'm not sure if I get this to 100%. This PR brings a feature to disable the auto negotiation of the language?
> 
> Because currently the language of a public page is automatically set by the browsers language. And as I get this right you want be able to disable that so your page would be always displayed in chinese?
> 
> Am I correct?

Hello! Thank you for taking the time to review this pull request.

I'd like to clarify the core purpose of this feature, as I believe there may be a misunderstanding due to my expression.

The current version of Wavelog **does not** have an automatic visitor language detection feature for OQRS page. The public OQRS page is always displayed in the application's default language (usually English), regardless of the visitor's browser settings.

This PR introduces this **new functionality from scratch**.

- **Before this PR:** The public page is always in English for everyone for OQRS page.
- **After this PR:** The page will now automatically detect the visitor's browser language and, if the language is supported by Wavelog, display the content in their native language on the first visit.

The on/off switch that I've added is part of this **new** feature, allowing users to enable or disable this new automatic detection behavior if they prefer their public page to always be in one specific language.

I hope this clears things up! Please let me know if you have any other questions.

HB9HIL:
Exactly this was my point. 

> Before this PR: The public page is always in English for everyone for OQRS page.

If this is true we have a bug, because all public pages shouldn't be always in english..

HB9HIL:
Sorry misclicked here :-D

HB9HIL:
So I tested here. When I clear all cookies and I set my preferred language in firefox/chrome to english
-> OQRS page is displayed in english

When I clear all cookies and I set my preferred language in firefox/chrome to portuguese
-> OQRS page is displayed in portuguese

There is just an issue with chinese. This PR solves a problem which doesn't exists (OQRS page isn't always english; btw: this is the same for all Wavelog pages when you aren't logged in). Therefore there is a bug around which does only affect at least chinese. Didn't check every language yet.

This PR will probably get closed since I think the issue is somewhere else. I need to check but please expect delay..

HNO3Miracle:
You are absolutely right. I've just tested with French and Japanese on my end, and they both work correctly, just as you found with Portuguese.

I'm a Chinese speaker, I only tested Chinese, so I mistakenly believed the feature was missing entirely.

It confirms your theory that this is a specific bug affecting how certain languages like Chinese are handled, rather than a missing feature.

Thank you for taking the time to investigate this so thoroughly. I'll close this PR now and will follow the issue for the bug fix. Let me know if you need any more testing from my side.

PR #2388, opened 2025-10-06
Randomize Clublog Jobs

HB9HIL:
We randomize all default expressions for clublog jobs to reduce load to clublog servers since amount of Wavelog Instances increased 🎉🥳


int2001 (commented):
works.

PR #2393, opened 2025-10-07
updated some bot related files

HB9HIL:
The robots.txt wasn't touched for years. In addition we tell the bots, that there is no sitemap for Wavelog


int2001 (commented):
good idea

int2001 (approved):

PR #2394, opened 2025-10-07
Clubcalypse2

int2001:
Trying to mitigate whatever is possible regarding Clublog-API.

Problem:
Each single 403 seems to trigger some kind of firewall which blocks the instance after 6 to 10 403s.

Things done:

- [x] use realtime.php interface of clublog, if there's only one QSO. Otherwise Clublog returns a 403 with `There are too many uploads already queued for your account. Please allow existing uploads to complete. If you are trying to do realtime uploads, please use the appropriate API (realtime.php, not putlogs.php)`
- [x] Process feedback from realtime.php - different messages than putlog.php from clublog
- [x] More intense logging, if something fails
- [x] Catch the Firewall-Case at Download, and stop downloading immediatly
- [x] Some Users use passwords like `Wavelog & WavelogGate is super!` (really, spaces and ampersands in it. this is now properly processed with the use of buildquery

Testing hints for reviewers:
Indeed a bit tricky. But here we go:
- Try logging ONE QSO with realtime disabled. Cron should now use the single-QSO-API from clublog
- Try logging it with wrong credentials and so on // try to enforce IP-Block at clublog
- Set your Clublog-password to something really weird.
- Try to reupload an already uploaded QSO (e.g. from another Logger or from wavelog).

### DIFF
--- a/application/models/Clublog_model.php
+++ b/application/models/Clublog_model.php
@@ -51,7 +51,7 @@ class Clublog_model extends CI_Model
 								$this->mark_qsos_sent($station_row->station_id);
 								log_message('info', 'Clublog singlepush upload for ' . $station_row->station_callsign . ' successfully sent and marked.');
 							} else {
-								log_message("Error", "Singlepush for ".$station_row->station_id." / ".$station_row->station_callsign." faied: ".$singlepush['status']);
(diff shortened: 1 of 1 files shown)


phl0:
Patch applied to my prod env. Will watch the next few days and report...

PR #2395, opened 2025-10-07
[Notes] Basic notes support at the QSO logging and QSO view 

szporwolik:
This continues the implementation of call sign–related notes and initially addresses this feature request:  
https://github.com/wavelog/wavelog/discussions/1653  

The changes are minor and affect two areas on the frontend.

## QSO Add Page
A small icon has been added to the QSO page with basic JS logic:  
2025-10-07 22_59_28-Greenshot

- By default, the icon is grayed out.  
- Once lookup is active, the existence of a note is checked. If a note for the callsign is available, the icon becomes active.  
- Clicking the icon opens a new tab allowing the user to **Add/Edit** note pages.  
- Resetting the form grays out the icon again.  

## QSO View
2025-10-07 23_48_06-qso php - wavelog (Workspace) - Visual Studio Code

- If a note for the callsign exists, a small icon is displayed.  
- Clicking it opens a new tab allowing the user to view the note.  

## General Notes
Callsigns are normalized by the backend using an existing function, so only one note per callsign is allowed.  
A user may enter `SP9MOA`, `SP9MOA/P`, `DL/SP9MOA`, or even `DL/SP9MOA/M` — all of these will refer to the same `SP9MOA` note.  

The idea (as in the recent PR for the Notes overhaul https://github.com/wavelog/wavelog/pull/2362) is that notes describe the **operator or club (i.e., the callsign)** rather than being tied to a specific QSO.

### DIFF
--- a/application/controllers/Logbook.php
+++ b/application/controllers/Logbook.php
@@ -617,6 +617,12 @@ class Logbook extends CI_Controller {
 		$data['query'] = $this->logbook_model->get_qso($id);
 		$data['dxccFlag'] = $this->dxccflag->get($data['query']->result()[0]->COL_DXCC);
 
+		// Check for note for this callsign and current user
+		$callsign = $data['query']->result()[0]->COL_CALL;
+		$user_id = $this->session->userdata('user_id');
+		$this->load->model('note');
+		$data['contacts_note_id'] = $this->note->get_note_id_by_category($user_id, 'Contacts', $callsign);
+
 		if ($this->session->userdata('user_measurement_base') == NULL) {
 			$data['measurement_base'] = $this->config->item('measurement_base');
 		}
--- a/application/controllers/Notes.php
+++ b/application/controllers/Notes.php
@@ -2,7 +2,6 @@
 
 // Notes controller: handles all note actions, with security and input validation
 class Notes extends CI_Controller {
-    // API endpoint: check for duplicate note title in category for user
     // Ensure only authorized users can access Notes controller
     function __construct() {
         parent::__construct();
@@ -41,19 +40,34 @@ class Notes extends CI_Controller {
     function add() {
         $this->load->model('note');
         $this->load->library('form_validation');
-		$this->load->library('callbook'); // Used for callsign parsing
+        $this->load->library('callbook'); // Used for callsign parsing
+
+        // Support prefilled title/category from query string
+        $prefill_title = $this->input->get('title', TRUE);
+        $prefill_category = $this->input->get('category', TRUE);
 
         $suggested_title = null;
         // Validate form fields
         $this->form_validation->set_rules('title', 'Note Title', 'required|callback_contacts_title_unique'); // Custom callback for Contacts category
         $this->form_validation->set_rules('content', 'Content', 'required');
         if ($this->form_validation->run() == FALSE) {
+            // Use POST if available, otherwise use prefill from query string
             $category = $this->input->post('category', TRUE);
-            if ($category === 'Contacts') {
-
-                $suggested_title = strtoupper($this->callbook->get_plaincall($this->input->post('title', TRUE)));
+            if (empty($category) && !empty($prefill_category)) {
+                $category = $prefill_category;
             }
+            if ($category === 'Contacts') {
+                $title_input = $this->input->post('title', TRUE);
+                if (empty($title_input) && !empty($prefill_title)) {
+                    $title_input = $prefill_title;
+                }
+                $suggested_title = strtoupper($this->callbook->get_plaincall($title_input));
+            }
+            // Pass prefill values to view
             $data['suggested_title'] = $suggested_title;
+            $data['prefill_title'] = $prefill_title;
+            $data['prefill_category'] = $prefill_category;
+            $data['category'] = $category;
             $data['page_title'] = __("Add Notes");
             $this->load->view('interface_assets/header', $data);
             $this->load->view('notes/add');
@@ -241,26 +255,36 @@ class Notes extends CI_Controller {
         $id = $this->input->get('id', TRUE); // Optional, for edit
         $check_title = $title;
         if ($category === 'Contacts') {
-            $check_title = strtoupper($title);
+            $this->load->library('callbook');
+            $check_title = strtoupper($this->callbook->get_plaincall($title));
         }
         $where = [
-            'category' => $category,
+            'cat' => $category,
             'user_id' => $user_id,
             'title' => $check_title
         ];
         $query = $this->db->get_where('notes', $where);
-        $duplicate = false;
+        $exists = false;
+        $note_id = null;
         if ($id) {
             foreach ($query->result() as $note) {
                 if ($note->id != $id) {
-                    $duplicate = true;
+                    $exists = true;
+                    $note_id = $note->id;
                     break;
                 }
             }
         } else {
-            $duplicate = $query->num_rows() > 0;
+            if ($query->num_rows() > 0) {
+                $exists = true;
+                $note_id = $query->row()->id;
+            }
         }
-        $this->output->set_content_type('application/json')->set_output(json_encode(['duplicate' => $duplicate]));
+        $response = ['exists' => $exists];
+        if ($exists && $note_id) {
+            $response['id'] = $note_id;
+        }
+        $this->output->set_content_type('application/json')->set_output(json_encode($response));
     }
 
     // Form validation callback for add: unique Contacts note title for user, only core callsign
--- a/application/controllers/Notes.php
+++ b/application/controllers/Notes.php
@@ -253,31 +253,17 @@ class Notes extends CI_Controller {
         $category = $this->input->get('category', TRUE);
         $title = $this->input->get('title', TRUE);
         $id = $this->input->get('id', TRUE); // Optional, for edit
-        $check_title = $title;
-        if ($category === 'Contacts') {
-            $this->load->library('callbook');
-            $check_title = strtoupper($this->callbook->get_plaincall($title));
-        }
-        $where = [
-            'cat' => $category,
-            'user_id' => $user_id,
-            'title' => $check_title
-        ];
-        $query = $this->db->get_where('notes', $where);
         $exists = false;
         $note_id = null;
-        if ($id) {
-            foreach ($query->result() as $note) {
-                if ($note->id != $id) {
-                    $exists = true;
-                    $note_id = $note->id;
-                    break;
-                }
-            }
-        } else {
-            if ($query->num_rows() > 0) {
+        $this->load->model('note');
+        $note_id_found = $this->note->get_note_id_by_category($user_id, $category, $title);
+        if ($note_id_found) {
+            // If editing, ignore current note
+            if ($id && $note_id_found == $id) {
+                $exists = false;
+            } else {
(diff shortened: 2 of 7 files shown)


int2001 (changes requested):
Nice idea. some comments on it:
- the note-icon should only be visible if the user has enabled notes and the user has entered a call.
- Why not use a modal for it, instead of a new tab/page? this breaks userflow while logging

HB9HIL:
Hello @szporwolik 

I have to agree with @int2001 

- Would be better to find another solution for the little icon... Probably as mentioned something dynamic
- I'd also go with a modal or a dedicated section on the QSO page itself. Breaking workflows is usually not a good idea

szporwolik:
Hello,
- Sure, adding a modal to the QSO page won’t be a problem; I actually considered it initially, but held off to avoid being too intrusive during the logging process
- Toggling the icon visibility based on user preferences is straightforward
- Regarding the icons in QSO View and QSO Logging — they load dynamically while the callsign is being looked up, per @int2001 comment I will hide it completly untill we are sure that callsign was entered;
@HB9HIL, could you clarify what you meant by “another solution”? Just a general direction of your thinking would be helpful 😄

int2001:
> Hello,
> 
> * Sure, adding a modal to the QSO page won’t be a problem; I actually considered it initially, but held off to avoid being too intrusive during the logging process
> * Toggling the icon visibility based on user preferences is straightforward
> * Regarding the icons in QSO View and QSO Logging — they load dynamically while the callsign is being looked up, per @int2001 comment I will hide it completly untill we are sure that callsign was entered;
>   @HB9HIL, could you clarify what you meant by “another solution”? Just a general direction of your thinking would be helpful 😄

you could also add a new tab (when notes are enabled) to the logging form. so no modal is needed, and the user could simply switch to the tab to put in his personal notes. just as an idea

HB9HIL:
@szporwolik Here are several ways possible. Some are better, some are not. 

We/You have to think about the inital usecase of those notes. The callsign-notes are made to see special information about an specific callsign. So if you typed something in, you probably want to see it again when you hit the same callsign again. 

Therefore I personally wouldn't go with a new tab in the QSO form, as this would require the user to activly see if there are any notes. Not really user-friendly.
On the other hand we don't want a big "Enter your shit here" for every callsign, since the user probably don't want to add a note for every single callsign he made a QSO with. 

The truth lies, as so many times before, somewhere in between. So let's see what we got:

grafik


This is our QSO window in it's natural state. We don't need anything here, and since there is no Callsign entered yet we don't need any input forms or note displays aswell. So let's enter a callsign:

Not worked before:
grafik

So there we go. We got a lot of information about this callsign. But we now want to see our Callsign Notes for this Call. **I personally** would love to see those notes right away. This is why I added them, right? My solution, would be to display the callsign notes somewhere, where I don't have to click anything to see them, but the do not disturb my workflow and they fit well into the whole "QSO Form System". I think we should allow also displaying notes for callsigns we didn't work before. Maybe I make a note about a really special call and when I finally work it I see all the details I did note about it.
But this last fact also affects the way to display the note, since the result on the QSO form is slighly different. 

grafik

The notes also could displayed right below the call itself, but this feels kind of intrusive (note that the colors are chosen to make it clear which parts are added)

grafik
 
If no note is added yet I think this would be a okaish solution to just add a little icon. However - we know our users - this will cause questions as people won't understand what to do here and how to add notes. 

grafik

Adding notes should probably done in a modal or a editable field in the QSO form (this can be another tab).

Now you see a little but what and how I'm thinking :D hahahaha Is this what you expected?

szporwolik:
LOL, thanks for this — it's as detailed as it could possibly be! 🤣

To be honest, I initially started with a tab-based approach similar to what @int2001 suggested. I even drafted a prototype, but I did not like it. The main reason was that it somehow ties the note to the QSO logic, which we want to avoid. As mentioned, we want to be able to create notes for callsigns that haven’t been worked yet etc. Putting this into tab would mean that we would need to save on "Save QSO" etc., I really do not like this.

Then I considered a modal approach, but I wasn’t sure about blocking the entire QSO panel for note editing/creation — it felt too intrusive.

I really like this approach:
2025-10-08 11_46_35-Desktop - File Explorer

This would display note/possibility to create or edit one when edited.

What do you think @int2001  @HB9HIL - shall this be a direction ?

HB9HIL:
As I said. This space is probably way to big for this purpose.. But if you design it well and it fits into the whole layout. why not. If notes feature is disable this won't be shown anyway. 

But make sure to design ist according to the rest of the page and make it intuitive, so users know exactly whats happening here (sounds easy but well you probably know, it's not always like that)

int2001:
My 5 Ct:

seen the configurable fields like "DOK" or "WWFF"-References?
one could configure them into the QSO-Mask for not switching tabs.

Perhaps this logic could also be nice for the notes.

- Notes Enabled and not configured to "front / 1st tab": Notes are visible/editable at a new tab.
- Notes Enabled and configured to "front / 1st tab": Notes are visible/editable at the free space.
- Notes disabled: No Note-Tab, no notes at free space.

szporwolik:
It was NOT an easy battle :tired_face:

What I did:

- checked QSO tabs in the code, checking the configurable fields per @int2001 suggestion and for me personally adding it into this QSO-related section somehow does not fit; we have QSO there — all get saved with QSO and it's almost strictly ADIF-related
- renamed existing Notes to Comment as this is, in fact, the true meaning of that field (we refer to it as comment in logbooks)
- tried to find a spot on the right side, next in the row with the partner's image — which I like, but would require massive playing with the state machine (both image and notes have preferences, can be pulled from QRZ or not, etc.); I can do it, but from the Bootstrap perspective it would be wise then to always display the image card (even if there is no image, some placeholder or so)
- then decided to chase @HB9HIL idea with using the space below the QSO panel — besides the crazy long General panel, I like it

This is how it looks now:
prototype

**Note:** code is NOT ready for review - this "simple to be" PR again adds almost 500 lines of code. I need to do some refactoring and testing.

Additional question - as you can see, I have used Bootstrap Toast to display some of the information. I think this is a nice, dynamic way to inform the user about certain actions taken. What do you think - is Wavelog ready for those? 😄 
If yes, I would include a generic container and simple JS in the general section, so this will be available across the tool.

phl0:
Looks fine to me so far. I just added a note for K1JT. Then disabled notes in the user options. The icon is still shown and I can edit the note though notes are disabled. I guess we should hide the icon if notes are disabled.

Screenshot from 2025-10-09 09-02-25

szporwolik:
> Looks fine to me so far. I just added a note for K1JT. Then disabled notes in the user options. The icon is still shown and I can edit the note though notes are disabled. I guess we should hide the icon if notes are disabled.

Yes Sir! 😄 This will be gone, I just left it for now for debug/testing purposes, so I can see the state of the note while I'm playing with the visibility of dedicated card.

szporwolik:
Generally, I would say this is ready for the review. 😄

HB9HIL:
I know.
Pretty busy here atm but did not forgot you
sry

HB9HIL:
@szporwolik Can you please resolve the conflicts? Then I can test it today. Sorry I was really busy the last couple days but today I planned to take care about this one

szporwolik:
@HB9HIL resolved, shall be fine now. 😄

HB9HIL (approved):
From my side it looks good so far and works pretty good. I'm sure there will popup small things after more users will test this. 

But I think it's merge ready. Checked code, but did not check logic. Anyway this should fit atm.

@int2001 should check aswell once more since he had some requested changes at the beginning. 

Approval from me and more testing in dev branch I guess

HB9HIL:
One more comment about JS files. Users will need to clear caches in order to work properly. We know about this and there are several points we can improve this. I will start tidy up this a bit during winter. Nothing serious. We just have to keep this in mind when people come up with JS issues.

int2001:
I am ok / fine with the PR.
If you ask me, i'd honestly answer, nice feature, **but** :)

image

you added the small arrow to the waterfall, where one can hide the thing. Not every QSO will have Notes, and i'm personally not really happy with the placing (since we have tabs). but now it is there. as a compromise: would you mind to add this small arrow to hide the input-field? (Default hidden)

so one could click the arrow and take / see notes if needed.

szporwolik:
> One more comment about JS files. Users will need to clear caches in order to work properly. We know about this and there are several points we can improve this. I will start tidy up this a bit during winter. Nothing serious. We just have to keep this in mind when people come up with JS issues.

@HB9HIL there is simple fix for that which I usually use in the other projects; if we have or would have global php variable holding used master code revision, adding it into JS inclusion would solve that.

Browsers are aggressive with the js caching, but take below spinets as separate files - doing force cache reload:
```
<script src="app.js?v=1.2.3"></script>
```
```
<script src="app.js?v=1.2.4"></script>
```

HB9HIL:
> > One more comment about JS files. Users will need to clear caches in order to work properly. We know about this and there are several points we can improve this. I will start tidy up this a bit during winter. Nothing serious. We just have to keep this in mind when people come up with JS issues.
> 
> @HB9HIL there is simple fix for that which I usually use in the other projects; if we have or would have global php variable holding used master code revision, adding it into JS inclusion would solve that.
> 
> Browsers are aggressive with the js caching, but take below spinets as separate files - doing force cache reload:
> 
> ```
> <script src="app.js?v=1.2.3"></script>
> ```
> 
> ```
> <script src="app.js?v=1.2.4"></script>
> ```

we have something more automated already in place

https://github.com/wavelog/wavelog/blob/e4d17a2a07be52b12420cb351b35b72ab4c9e8fa/application/controllers/Stationsetup.php#L30-L35

This loads the JS files via controller and with filemtime it checks for last modified attributes. It's similar to your solution but automated version based on mtime attributes

HB9HIL:
We started once to rebuild all fileloads in the footer to this controller based loading. but well as time goes on this gets forgotten from time to time :-D

szporwolik:
@int2001 I will be stubborn on that - as stated previously, the tabs are QSO related, not call-sign related! 😄 

But... you proposed the middle ground, so here it comes 👍🏻 :
image

image

@HB9HIL @int2001 : Ok ?

phl0:
Are the stored notes supposed to be shown if I enter a callsign for logging?

szporwolik:
> Are the stored notes supposed to be shown if I enter a callsign for logging?

Yes, that's the idea behind. You can access the notes via Notes page from header or have the one related to callsign displayed after Callsign lookup is completed.

One note per base callsign - so DL/SP9MOA/P shall create/show note for SP9MOA.

phl0:
So (still) a bug here?

https://github.com/user-attachments/assets/f43bdd60-21bb-4fd5-ae6d-eb102cc0ecd6

phl0:
Ah wait. My fault. Probably tried the patch against the wrong branch. QRX

szporwolik:
> Ah wait. My fault. Probably tried the patch against the wrong branch. QRX

Please ensure js is up to date with CTRL+F5, not sure it was a few weeks from my major work on that, but exactly - looks like some old js file.

szporwolik:
... which would only proof @HB9HIL point with javascript files! 😆

phl0 (approved):
Checked out your branch and everything is working. Approved from my point of view.

PR #2410, opened 2025-10-15
[Winkey] Added 5 more macros

AndreasK79:
image

As I changed from using cwmacros to user_options, the macros need to be set again.


phl0 (approved):
Added a few lines. Removed debug code. Let Øs replace. Works for me now :)

PR #2429, opened 2025-10-22
[DXWaterfall] DX Waterfall for QSO page

szporwolik:
## Background

This one falls into the "love it or hate it" category... 😉 
 
As an *amateur-amateur* ham radio operator with a simple wire antenna setup and a single TRX at QTH, I decided to switch from desktop logging software to **Wavelog**. It works well for most of my needs, but I really miss having a convenient **DXCluster** view directly on the logging page to quickly check what's happening on the bands.  

That’s where the proposed **DX Waterfall** feature comes in. 🎉

## Overview
The DX Waterfall is a **JavaScript object** built on a `canvas` element.  
It appears above the QSO logging section on the QSO page when operating in **Live Logging** mode.

---

---

---

---

### Main Features

- Uses Wavelog’s cluster model to pull DX spots.  
  A minor change was proposed to the [DXClusterAPI](https://github.com/int2001/DXClusterAPI/issues/31) to improve SOTA integration (a fallback is included if the change is not accepted).
- Works **with or without CAT control**.  
  When CAT control is active, the radio frequency has priority — moving the dial updates the center marker (red line), and changing bands on the radio triggers a DXCluster update.  
  It can also work in reverse, allowing the waterfall to control the radio.
- **Mouse interaction**: users can click anywhere on the waterfall to jump to a frequency or manually enter a value.
- **Keyboard navigation** (similar to N1MM Logger+):  
  Users can control the waterfall using keyboard shortcuts to:
  - Move between spots (Next/Previous).
  - Jump to new Continents/DXCCs (DX Hunter).
  - Adjust zoom levels.
  This allows for fast and efficient operation without relying on the mouse.
- **Navigation buttons**:
  - **Next / Previous** – cycle through spots on the band.  
  - **DX Hunter** – cycles through new Continents/DXCCs (if available).
- Supports **5 zoom levels**, covering everything from HF CW to VHF/UHF operation.
- Quick switching of **spot origins** and **mode filters**.  
  The spot age and continent defaults are taken from existing Wavelog global options.
- Displays the appropriate **band plan** section based on the selected continent (TODO: improve Wavelog's Bandedges).
- When near a spot:
  - The top bar shows spot info.  
  - Two buttons appear:
    - Tune to the exact spot frequency.
  - Detected references (WWFF, IOTA, POTA, SOTA) are automatically filled, and spot details are fetched.

## General Notes

While this feature introduces over **7500 lines of code**, most of it resides in a single monolithic JS file managing the `DXWaterfall` object.  Only minor changes were made to the QSO page and `qso.js` for integration.
I had to touch user settings, so per comment in the code I'm tagging @int2001 @phl0 .

The feature can be **enabled per user** in their settings and is **disabled by default**.

---

**This is draft pull request, there is still lots to improve and especially to test - I will let you guys now when ready for the testing. If in the meantime anybody will be willing to check/test I'm open to the feedback.**

---
## TODO
- [x] CTRL-Space, CTRL-Shift-Space button presses (with Mac support?)
- [x] Mouse zoom-in / zoom-out
- [x] PHP error in User_mode.php
- [x] CAT puling rate into config.php, defaulted to 3000ms
- [x] Clicking near spot shall fill out the QSO form
- [x] Get rid of json file with IARU bandplan, use the build in Bandplan-Data
- [x] Rework/re-think display of the cluster data status
- [x] Turn On/Turn Off feature, even if feature is enabled in options
- [x] Make font size configurable / Implement size control buttons some place in the waterfall GUI
- [x] Tooltips for labels and "click to jump" feature
- [x] Websocket support, needs https://github.com/wavelog/wavelog/pull/2434
- [x] Extracting CSS styles to seperate file
- [x] Support of split operation
- [x] Prepare to be used at other subpages
- [x] Update logic of the QSO for prefilling

**Prior merge**
- [x] Remove js inclusion cache avoidance in dxwaterfall.php component
- [x] Remove redundant tune to workaround related to https://github.com/wavelog/WaveLogGate/pull/83
- [x] Remove debug logs
- [x] Final code refactor
- [x] Final testing

### DIFF
--- a/application/views/user/edit.php
+++ b/application/views/user/edit.php
@@ -430,6 +430,7 @@
 											} ?>
 										</select>
 									</div>
+									<hr />
 									<!--- DX Waterfall -->
 									<div class="mb-3">
 											<label for="user_dxwaterfall_enable"><?= __("DX Waterfall"); ?></label>


int2001:
WOW!
Simply WOW!

i just read your description and pulled it on my dev. stunning.
this is really really nice.

i need to dig in a bit and have already a few questions (just testing it out for about 5mins, not more):
1. Great, you even thought about dis- and enabling. I like!!! This feature could be so hot, that one wants to "toggle" the view from QSO to QSO. 1st thought was/is: what about a "show/hide" element at QSO?
2. You're using the DXCluster-Endpoints WL provides. this is neat. so information about QSL-Status is included, right? 
3. Clicking on a spot tunes the Radio to that spot, but call doesn't seem to be copied to QSO-Form. Perhaps i am using it wrong?
4. Bandplan: i saw you implemented a static JSON for the Edges. Qustion: Why not use those from the database instead?

i grabbed my 1st cup of coffee a few minutes ago and i am flashed about your PR. This is one of the best things i've seen during the last few months. Respect!

Let's try to make it production-ready together. however: really great work.

int2001:
Minor Issues i found so far:
Using a Mac here.
- [ ] HotKey CTRL-Space doesn't seem to work (Chrome). Consider preparing the log, if one clicks on cluster?!
- [ ] CTRL-Shift-Space doesn't work either here (QSY to Spot)
- [ ] The "Spots fetched message" prevents the user from clicking on the buttons
- [ ] Zooming/Moving in/out via Mouse? No real issue, just nice 2 have

This feature is really amazing. It even works together with the actual Bandlist at tools.  i am flashed.

AndreasK79:
@szporwolik looking at the pictures, I gotta say WOW myself. This looks really nice, and once we iron out the bugs, I guess it can be merged.

I stumbled onto an error here, that seems to kill most functionality:
image

application/views/interface_assets/footer.php
```diff
@@ -1465,13 +1563,19 @@ function toggleGridsquares(bool) {
...
 		    }
 	    };
 
-	    // Update frequency every three second
-	    setInterval(updateFromCAT, 3000);
+	    // Update frequency from CAT
+	    // Add a small delay (500ms) before first poll to allow radio interface to initialize
+	    setTimeout(function() {
+	        updateFromCAT();
+	        // Then poll every 500ms
```
int2001: This is far too much. We know of instances running a thousand users and more. 3seconds was set for a reason.
i know: websockets would make it much more easier here. but we're on PHP and websockets are not really possible without a daemon (which can't be started/run on "el-cheapo-hosters")

int2001: suggestion, as a compromise:
set it back to 3000 (as default), and make it configurable at config.php (for small instances).
please no configuration in UI. Reason: some admin without deeper knowledge could set it accidently to 500ms (or even lower) and crash the whole instance

szporwolik: Yep, the thing is that the lower you go, the more responsive it get's when you are dialing radio. 

What about new global setting for that - single, small instances hosted locally would probably go as low as 100ms, while admin's of the big ones could set it to 3 or event 5 secs. ?

szporwolik:
> @szporwolik looking at the pictures, I gotta say WOW myself. This looks really nice, and once we iron out the bugs, I guess it can be merged.
> 
> I stumbled onto an error here, that seems to kill most functionality: image

Will check, you are clearly using other PHP version than myself. Any chance to provide PHP version which you are on ?

int2001:
sure. i noticed the bug as well. but only once. my dev runs on 8.4.12. production on 8.3.24.
there are multiple versions used out there. it should work with every PHP version >=8.0.

Hint: @AndreasK79 found already out a possible root-cause: see here: https://github.com/wavelog/wavelog/pull/2429/files#diff-75a656aa0d26b2f4cb87a9a884fac77b4f8913708ad7abf984bdefe1d2b05324R227

the clubstation thing has to be the last parameter

AndreasK79:
@szporwolik PHP 8.2.4 here.

application/views/interface_assets/footer.php
int2001: 100ms? that even more than the 500ms atm.
i'd vote for a default of 3000ms.

and instance-owner can adjust the value within config.php. put a warning to config.php that this may cause a DoS when set too aggressive. 

i think with that compromise even "hold my beer while instance is burning"-owners could set it (on their own risk) to 100ms or less.
But default should be rock-solid and stable. 3000ms was and is tested even with huge instances in the wild.

[edit - thinking loud]
One could implement a Websocket-Server to WavelogGate and use that, instead of doing the roundtrip via old'n'busted PHP. but that's another huge refactor.

szporwolik:
> 1. Great, you even thought about dis- and enabling. I like!!! This feature could be so hot, that one wants to "toggle" the view from QSO to QSO. 1st thought was/is: what about a "show/hide" element at QSO?

I was thinking about it as well, so enabling the feature would give you an option to then later turn of/off the feature.
This will complicate things a little as the main challenge is to handle all of those race conditions with async calls to CAT and DX cluster. Will think about.

> 2. You're using the DXCluster-Endpoints WL provides. this is neat. so information about QSL-Status is included, right?

Yes, this is not the most intuitive thing in the world, but the spot labels have meaning: red (not worked), orange (worked not confirmed), green (confirmed). The border goes for continent, label's background for DXCC and the small dot next to the callsign - the callsign. I was space limited and did not wanted to clutter the view to much.

When spot is active, the gold, silver and bronze are displayed as well.

> 3. Clicking on a spot tunes the Radio to that spot, but call doesn't seem to be copied to QSO-Form. Perhaps i am using it wrong?

This is correct behavior, you can click/scroll through the band and the nearest spot information will be displayed at the top bar. Then you can use those two small icons at the top to copy info to QSO Form or tune to exact frequency.
I you want to tune and copy, CTRL+Arrow keys is a way to go.

> 4. Bandplan: i saw you implemented a static JSON for the Edges. Qustion: Why not use those from the database instead?

This is one of the features which we miss, the Edges does some of the job - but I wanted to really mark the available band depending on the IARU region you are work for. So, now there is json in the future I would probably go with the full region support in the Wavelog backend with bandplans stored in the MySQL.

int2001:
> > 1. Great, you even thought about dis- and enabling. I like!!! This feature could be so hot, that one wants to "toggle" the view from QSO to QSO. 1st thought was/is: what about a "show/hide" element at QSO?
> 
> I was thinking about it as well, so enabling the feature would give you an option to then later turn of/off the feature. This will complicate things a little as the main challenge is to handle all of those race conditions with async calls to CAT and DX cluster. Will think about.

Great. could make it much more useful

> > 2. You're using the DXCluster-Endpoints WL provides. this is neat. so information about QSL-Status is included, right?
> 
> Yes, this is not the most intuitive thing in the world, but the spot labels have meaning: red (not worked), orange (worked not confirmed), green (confirmed). The border goes for continent, label's background for DXCC and the small dot next to the callsign - the callsign. I was space limited and did not wanted to clutter the view to much.

once understood (you even thought about tooltips - wow!) it is a nice approach.

> When spot is active, the gold, silver and bronze are displayed as well.
> 
> > 3. Clicking on a spot tunes the Radio to that spot, but call doesn't seem to be copied to QSO-Form. Perhaps i am using it wrong?
> 
> This is correct behavior, you can click/scroll through the band and the nearest spot information will be displayed at the top bar. Then you can use those two small icons at the top to copy info to QSO Form or tune to exact frequency. I you want to tune and copy, CTRL+Arrow keys is a way to go.

Yes, sure. but it isn't intuitive. i had to search by myself how to copy the spot into the QSO-Form. i'd still vote for "copy it automatically to qso-form" when clicked. if you feel uncomfortable with it, what about a compromise: some switch at the waterfall, where one could say: "Take the clicked spot automatically to the form" when enabled (default). And when disabled he has to click on the icon or use the keyboard-shortcut? One main mantra of Wavelog is: keep it simple, reduce support for end-users.

> > 4. Bandplan: i saw you implemented a static JSON for the Edges. Qustion: Why not use those from the database instead?
> 
> This is one of the features which we miss, the Edges does some of the job - but I wanted to really mark the available band depending on the IARU region you are work for. So, now there is json in the future I would probably go with the full region support in the Wavelog backend with bandplans stored in the MySQL.

tbh: we're trying to get rid of redundant Bandplan-Data during the last months. i strongly recommend to use the data which is in database. even if Regions are missing there. Way forward could be: add regions later to the existing DB it and use them. but for now i'd rather see it without the Region-Edges than implementing an extra source for the data

application/views/interface_assets/footer.php
szporwolik: I will go with the config.php for now that is a good proposal and in fact it's easier than anything else. 😉 

There is a locking mechanism, so timer ticks every value given (like 500ms), but actual request is being send only if the previous was has completed.

szporwolik:
> * [ ]  HotKey CTRL-Space doesn't seem to work (Chrome). Consider preparing the log, if one clicks on cluster?!
> * [ ]  CTRL-Shift-Space doesn't work either here (QSY to Spot)

I'm not the Mac person, what would be the default key at Mac to modify role of the button ?

> * [ ]  The "Spots fetched message" prevents the user from clicking on the buttons

This will be displayed when there is no spot nearby, I wanted to limit the place the feature takes, so this is kind of workaround.

> * [ ]  Zooming/Moving in/out via Mouse? No real issue, just nice 2 have

Good idea! 😄

szporwolik:
Added TODO list to the PR description to better track the suggestion/bugs.

int2001:
> > * [ ]   HotKey CTRL-Space doesn't seem to work (Chrome). Consider preparing the log, if one clicks on cluster?!
> > * [ ]   CTRL-Shift-Space doesn't work either here (QSY to Spot)
> 
> I'm not the Mac person, what would be the default key at Mac to modify role of the button ?

i can only tell what's working: all those Hotkeys which are based on "ALT" as modifier key. but: sometimes Firefox overrides keybindings with its own hotkeys. E.g.: Have a look at ALT+L binding.

> > * [ ]   The "Spots fetched message" prevents the user from clicking on the buttons
> 
> This will be displayed when there is no spot nearby, I wanted to limit the place the feature takes, so this is kind of workaround.

Understand. but without the "click and log"-Feature i already mentioned, there's no copying of spot/call possible. i noticed the message as well, when there's a spot. not intutive, as told. it could all be solved with that "click and log"-feature. 

> > * [ ]   Zooming/Moving in/out via Mouse? No real issue, just nice 2 have
> 
> Good idea! 😄

application/views/interface_assets/footer.php
int2001: i started working on the websocket within WLGate-Feature, to reduce serverload and gain low-latency

szporwolik:
> > I'm not the Mac person, what would be the default key at Mac to modify role of the button ?
> 
> i can only tell what's working: all those Hotkeys which are based on "ALT" as modifier key. but: sometimes Firefox overrides keybindings with its own hotkeys. E.g.: Have a look at ALT+L binding.

Thanks, will think what to do with it. 
> > > * [ ]    The "Spots fetched message" prevents the user from clicking on the buttons
> > 
> > This will be displayed when there is no spot nearby, I wanted to limit the place the feature takes, so this is kind of workaround.
> 
> Understand. but without the "click and log"-Feature i already mentioned, there's no copying of spot/call possible. i noticed the message as well, when there's a spot. not intutive, as told. it could all be solved with that "click and log"-feature.

Added to todo, I think I have an idea how to handle this.

szporwolik:
Implemented initial changes after your feedback (thanks!):
@int2001 - please check the Mac support, I would need to visit my friends or family to test it - according to prof. GPT5 this shall work for you 😉; Cmd key is used as modifier now
@AndreasK79 - please check the PHP error, It was not smart of me to put non optional argument after optional - fixed, but I think I've also found the same bug already in the code - see https://github.com/wavelog/wavelog/issues/2435

Items changed:
- CTRL-Space, CTRL-Shift-Space button presses (with Mac support?)
- Mouse zoom-in / zoom-out
- PHP error in User_mode.php
- CAT puling rate into config.php, defaulted to 3000ms
- Clicking near spot shall fill out the QSO form
- Get rid of json file with IARU bandplan, use the build in Bandplan-Data
- Rework/re-think display of the cluster data status
- Turn On/Turn Off feature, even if feature is enabled in options

update2


More to come... Any feedback is more then welcome!

AndreasK79:
> * PHP error in User_mode.php

The error is gone now, so great work! As for the other issue you raised, I'll look into it at a later point (probably this weekend).

One thing I noticed though. The labels are very small, and I need to increase the zoom level in the browser to read them.
I have a couple of suggestions:

1. Make font size configurable
2. Implement size control buttons some place in the waterfall gui
3. Have a mouseover with information about the spot (time of spot, spotter, needed for what etc)

Keep it up!

application/views/interface_assets/footer.php
int2001: i just released WLGate V1.1.10 which supports websockets.
in combination with #2434 it reduces the load in a massive way. There are little delays when doing QSY, but that's because the Gate "debounces" every QSY/ModeChange to make sure frequency is settled.

szporwolik: Perfect! The websocket is the way to go, question about the back combability - having the current DX Waterfall to work with CAT and handling all async cases requires a lot of code. Implementing WebSocket in parallel will make everything even more complicated.

Proposal: I would drop support of the legacy code (CAT with MySQL in between) and make to code working only with the websockets.

What do you @int2001 think about ?

int2001: i built a ready 2 use codesnippet into footer.php and contesting.js. it supports both. see [here](https://github.com/int2001/wavelog/blob/1942e9da85709c206c5217a96ec3ca741b3bee93/assets/js/sections/bandmap_list.js#L310): 
on WebSocket: Realtime with pushes.
on Fallback/Roundtrip mySQL: 3000ms

once this is merged i could help you to refactor this beauty in a way it works with both.

int2001: Backwards-Compat is important. not everyone uses WLGate. Lots of people use own python/bash/$whatever scripts.

szporwolik: Got you, will check in the evening. Everything is possible, it's only just about additional lines of code. 😄

szporwolik:
Here you go:

> 1. Make font size configurable
> 2. Implement size control buttons some place in the waterfall gui

- [x] Turn On/Turn Off feature, even if feature is enabled in options
- [x]  Make font size configurable / Implement size control buttons some place in the waterfall GUI
 
> 3. Have a mouseover with information about the spot (time of spot, spotter, needed for what etc)

- [x]  Tooltips for labels and "click to jump" feature

Didn’t have much time to do thorough testing today. Also found another bug which took my like an hour to find and understand 🥲 https://github.com/wavelog/WaveLogGate/issues/82

Please check and let me know what do you think.

2025-10-25 00_02_39-Greenshot

AndreasK79:
@szporwolik the label fonts and sizing is working good.

One thing came to mind. I see you put the CSS in the view. My suggestions would be to move that to general.css. That way, it is active for all themes, and the waterfall can be styled in the overrides.css for each theme.

Should we have a dark/light mode as well for the bandmap? We have different styles of themes.

szporwolik:
> @szporwolik the label fonts and sizing is working good.

😃👍🏻 

> One thing came to mind. I see you put the CSS in the view. My suggestions would be to move that to general.css. That way, it is active for all themes, and the waterfall can be styled in the overrides.css for each theme.
> 
> Should we have a dark/light mode as well for the bandmap? We have different styles of themes.

Yes, my game plan is to move the CSS out of PHP; this is for sure to be done once the major part of development is done and just before PR merge.

Then gather user feedback. My expectation is that there will be tons of bugs coming once it reaches Dev and then hopefully Master branches. 😄 Once we are in the stable condition, without "dozens" of bugs coming in, work on dark/white mode of the DX Waterfall.

So, yes - I remember about those happy/optimistic guys not using dark modes, but let's keep them wait for the bright mode to not complicate the release with another path to debug...

AndreasK79:
Sounds like you have it planned out! Just continue the good work then :)

HB9HIL (changes requested):
I just gave this great feature a quick try. Really nice. There is just one thing which nags me..

When the DXwaterfall is enabled and I'm currently on the QSO Page it automatically choses a callsign and prefills the QSO data without any interaction. I guess this happens when new cluster data comes in. But IMHO there shouldn't get any data prefilled if the user don't interact in any way.

szporwolik:
> I'm currently on the QSO Page it automatically choses a callsign and prefills the QSO data without any interaction. I guess this happens when new cluster data comes in. But IMHO there shouldn't get any data prefilled if the user don't interact in any way.

Yes, after giving it a 2nd though I completely agree. 

I originally was not planning to pre-fill QSO form at all without user interaction, you can see the "Copy Callsign to QSO form" button in the early screenshots. Later after discussion here, but mainly due to own decisions I went definitely to aggressive here.

Looking wider - you are definetly right - user can be already filling/doing QSO with contact which was not spotted and the form and background DXCluster data update will fill your form.

My proposal:
- Keyboard arrows :right :arrow_right: QSO is prefilled ✅ 
- Clicking the prev/next spot buttons :arrow_right: QSO is prefilled ✅ 
- Mouse clicking at spot's label :arrow_right: QSO is prefilled ✅ 
- DX Hunter button :arrow_right: QSO is prefilled ✅ 
- Cycle through relevant spots (this happen if 2 or more spots are near each center marker)  :arrow_right: QSO is prefilled ✅
- Dialing radio frequency :arrow_right: QSO is NOT prefilled ❌ 
- Clicking on water fall to change frequency :arrow_right: QSO is NOT prefilled ❌ 
- Entering frequency manually to the input field :arrow_right: QSO is NOT prefilled ❌ 
- No frequency change, but DXCluster comes in the background  :arrow_right: QSO is NOT prefilled ❌ 

What do you guys think ?

HB9HIL:
IMHO prefilling only on explicit action. Automated prefills from various triggers will cause issues and will make people go crazy if something is accidentially overwritten. 

So in my opinion. No ***automatic*** prefiling. Only on explicit action.

HB9HIL:
My conter proposal:

    Keyboard arrows :right ➡️ QSO is NOT prefilled ❌ (will cause confusion)
    Clicking the prev/next spot buttons ➡️ QSO is NOT prefilled ❌ (will cause a lot of load)
    Mouse clicking at spot's label ➡️ QSO is prefilled ✅
    DX Hunter button ➡️ QSO is prefilled ✅
    Cycle through relevant spots (this happen if 2 or more spots are near each center marker) ➡️ QSO is prefilled ✅
    Dialing radio frequency ➡️ QSO is NOT prefilled ❌
    Clicking on water fall to change frequency ➡️ QSO is NOT prefilled ❌
    Entering frequency manually to the input field ➡️ QSO is NOT prefilled ❌
    No frequency change, but DXCluster comes in the background ➡️ QSO is NOT prefilled ❌

HB9HIL:
Cycle through relevant spots (this happen if 2 or more spots are near each center marker)

what does this mean exactly?

szporwolik:
@HB9HIL Please jump in the place where there are two spots at the same frequency or really close each other - the new button will be shown in the top left corner with the text [1/3] or similar, this will let you cycle through those spots.

HB9HIL:
Ah yes I see. Well imho normal click on the waterfall should not trigger a prefill. Imagine a big instance like DARC with a few hundred users. And at a weekend they click on the cluster. This triggers a load which is extreme since the lookup in the qso windows is one of the biggest and load-intensive ones which we have. It's mandatory then that two spot labels do not block each other

Keyboard arrows :right ➡️ QSO is NOT prefilled ❌ (will cause confusion)
Clicking the prev/next spot buttons ➡️ QSO is NOT prefilled ❌ (will cause a lot of load)
Mouse clicking at spot's label ➡️ QSO is prefilled ✅
DX Hunter button ➡️ QSO is prefilled ✅ ( I think we can live with that one)
Cycle through relevant spots (this happen if 2 or more spots are near each center marker) ➡️ QSO is prefilled ✅ (this one is okay only if the spot was clicked directly and not when just clicking into the waterfall near those spots.)
Dialing radio frequency ➡️ QSO is NOT prefilled ❌
Clicking on water fall to change frequency ➡️ QSO is NOT prefilled ❌
Entering frequency manually to the input field ➡️ QSO is NOT prefilled ❌
No frequency change, but DXCluster comes in the background ➡️ QSO is NOT prefilled ❌

The main goal should be: Load where load is necessary. Automated prefilling is not necessary and can cause serious issues (accidentially overwriting already prefilled qso data for example).

HB9HIL:
PS: I just tested. I clicked on a spot, the callsign was prefilled. I edited some details in the QSO... suddenly the automation kicks in and overwrites all my data I added to the QSO.. This is a no go

szporwolik:
> PS: I just tested. I clicked on a spot, the callsign was prefilled. I edited some details in the QSO... suddenly the automation kicks in and overwrites all my data I added to the QSO.. This is a no go

Please don't bother with that for some time, I have actually some commits waiting to be pushed to integrate with websockets and some additional fixes.

I will let you guys know here, when the last code is at the github.

int2001:
@HB9HIL is absolutely right. this won't work. once a QSO has been filled / looked up, it's up to the user what should happen here.

take your time regarding a fix/solution - but this is really important

szporwolik:
- Resolved merge conflicts after recent PRs into `dev`.
- WebSocket support should now be functional.
- Spot logic updated according to @HB9HIL's proposal.
- Various race conditions have been addressed (still worth double-checking and probably... still broken in a few places).

**General comment:**  
The console is currently spammed with debug output (this will be removed before merge).  
If you notice any strange behavior, please let me know — the console logs will help with debugging.  
This has grown into a large and complex piece of code with many execution paths; I’d really appreciate if you all spend some time trying to break it. There were dozens of race conditions that had to be handled — I’m not 100% sure all of them are fully covered.

---

@int2001 — three points to discuss:

1. Please review https://github.com/szporwolik/wavelog/blob/8c93692eced9bf58409ec445f4165ca3c8c552e7/application/views/interface_assets/footer.php#L1548-L1550 — what should be the correct logic there when operating in `ws` mode?  
2. I’ve slightly modified the WebSocket code in the footer. Could you confirm this doesn’t break contesting or bandmap features? (My tests show no issues.)  
3. **Out of scope for this PR**, but: should we consider making WebSocket the default mode for users?  
   - If yes, where should this setting be configured — in *user account*, *hardware interfaces*, or *elsewhere*?  
   - Since WebSocket works reliably, there seems to be little reason to continue using the legacy pull-via-database method; I got your note of the back compatibility, I'm talking about defaulting it instead of removing.

int2001:
Currently doing a contest-shift at DL3O. Will check during upcoming week

int2001:
> @int2001 — three points to discuss:
> 
> 1. Please review https://github.com/szporwolik/wavelog/blob/8c93692eced9bf58409ec445f4165ca3c8c552e7/application/views/interface_assets/footer.php#L1548-L1550 — what should be the correct logic there when operating in `ws` mode?

You implemented it correct. even in ws mode the QSY should (and is) triggered by a "normal" request. Only thing i noticed: it's triggering twice, when i click a spot. See network-tab - two requests are made. pse check.

> 2. I’ve slightly modified the WebSocket code in the footer. Could you confirm this doesn’t break contesting or bandmap features? (My tests show no issues.)

just tested old-cluster, live-qso, live-contesting with classic and websocket Radio, from the user-perspective = everything works. 

But you modified it, that the radio-name is now shown instead of the "Websocket-String". this could be missleading. 
Background: It's important for the user to see if he's using a keydefined-radio (which may run on another host) or websocket (which is always the computer you're browsing wavelog with). in my opinion the little star is not enough / not intuitive for the user. we're introducing that websocket-feature with the next version. People need to see on 1st sight if its websocket/localhost or a classic radio. The Users don't open the debug-pane ;)

> 3. **Out of scope for this PR**, but: should we consider making WebSocket the default mode for users?
>    
>    * If yes, where should this setting be configured — in _user account_, _hardware interfaces_, or _elsewhere_?
>    * Since WebSocket works reliably, there seems to be little reason to continue using the legacy pull-via-database method; I got your note of the back compatibility, I'm talking about defaulting it instead of removing.

Not sure - perhaps in another PR. Our Userbase is very mixed. Many many FT8-People. and for FT8-Transportation the key is still neded (we can think about putting that to websocket as well - but that'd mean that WL has to be open - tricky). if we put it to default there'd be complains about "why isn't my FT8 working?". imho we should leave it as it is for now


Disclaimer: Didn't look into the code, yet. hopefully i find a bit sparetime during the week.

szporwolik:
Thanks!

> See network-tab - two requests are made. pse check.

Yes that was a fix for the Yaesu behavior discussed in the WL Gate repository, I have removed it yesterday as you merged the fix there. 
For now Yaesu users can have some bugs when cycling between CW and SSB spots, solution will be to upgrade to the latest WLGate version once it's released.

> But you modified it, that the radio-name is now shown instead of the "Websocket-String" 

Right, I forgot about it! Will update in the evening, anyhow I think once this one is merged one day, I will craft some PR to get the radio/cat logic out of footer and unify usage across the whole WL.

> imho we should leave it as it is for now

Got it!

> Didn't look into the code, yet

No worries, I would say - please focus on the functional tests first, once we all agree that this works as expected without critical bugs, there will be time for final code refactor and detailed review.

One more question: I've been playing with websockets and this feature yesterday, it feels nicely with the decent feeling of responsivity, there is no comparison to CAT pooling which feels "laggy" even with 500ms between pools. 
I would remove the variable which I added to config.php and set the CAT pooling to default 3000ms - what do you think ? Leave it with the option definition in config.php or revert back to hardcoded value ?

int2001:
> One more question: I've been playing with websockets and this feature yesterday, it feels nicely with the decent feeling of responsivity, there is no comparison to CAT pooling which feels "laggy" even with 500ms between pools. I would remove the variable which I added to config.php and set the CAT pooling to default 3000ms - what do you think ? Leave it with the option definition in config.php or revert back to hardcoded value ?

revert to hardcoded value with 3000ms, please. better safe than sorry on huge instances

szporwolik:
> Yes that was a fix for the Yaesu behavior discussed in the WL Gate repository, I have removed it yesterday as you merged the fix there. For now Yaesu users can have some bugs when cycling between CW and SSB spots, solution will be to upgrade to the latest WLGate version once it's released.

Errata: seems I only though I fixed it yesterday, double tune fixed now for sure. 😉

int2001:
From my point of view it works very good now.

szporwolik:
Tested, verified, and refactored — this will still require extensive testing while in the ```dev``` branch.

I understand this review won’t be easy — it’s a massive one (sorry 🫨 ).  My suggestion: focus mainly on the common file changes.  
The main `dxwaterfall.js` file can be treated as an external library — no need to dive deep into it.  I’ll handle any initial bugs that appear there.

**Note:**
To be completely honest, there are still some rare race conditions causing occasional strange UX behavior.  
They don’t block the user, but may not look the nicest.

Tuning and band changes are asynchronous, with no instant confirmation from the radio that tuning succeeded.  
Making this component “perfect” would require dropping polling support and moving fully to WebSockets, plus some work on **WLGate** to ensure proper confirmations are sent.

Considering all the great discussions above — and keeping backward compatibility and WLGate simplicity in mind — I decided to release it as it is.  In my humble opinion, it’s more than *“good enough”*, though not *“perfect”*.

As we often say in Poland: _Lepsze jest wrogiem gorszego_ / _Le mieux est l’ennemi du bien_ / _Das Bessere ist der Feind des Guten_ / _Perfect is the enemy of good_ - so, let’s review, fix and go with it. 🙂

application/views/qso/index.php
```diff
@@ -350,9 +383,9 @@
...
               <select class="form-select radios" id="radio" name="radio">
                 <option value="0" selected="selected"><?= __("None"); ?></option>
-		<option value="ws"<?php if ($this->session->userdata('radio') == 'ws') { echo ' selected="selected"'; } ?>><?= __("WebSocket (Requires WLGate>1.1.10)"); ?></option>
+		            <option value="ws"<?php if ($this->session->userdata('radio') == 'ws') { echo ' selected="selected"'; } ?>><?= __("Live - ") . __("WebSocket (Requires WLGate>1.1.10)"); ?></option>
                 <?php foreach ($radios->result() as $row) { ?>
-                  <option value="<?php echo $row->id; ?>" <?php if($this->session->userdata('radio') == $row->id) { echo "selected=\"selected\""; } ?>><?php echo $row->radio; ?> <?php if ($radio_last_updated->id == $row->id) { echo "(".__("last updated").")"; } else { echo ''; } ?></option>
+                  <option value="<?php echo $row->id; ?>" <?php if($this->session->userdata('radio') == $row->id) { echo "selected=\"selected\""; } ?>><?= __("Pooling - ") . $row->radio; ?> <?php if ($radio_last_updated->id == $row->id) { echo "(".__("last updated").")"; } else { echo ''; } ?></option>
```
int2001: it's called "polling". i gues only a typo here ;)

application/views/qso/index.php
```diff
@@ -350,9 +383,9 @@
               <label for="radio"><?= __("Radio"); ?></label>
               <select class="form-select radios" id="radio" name="radio">
                 <option value="0" selected="selected"><?= __("None"); ?></option>
-		<option value="ws"<?php if ($this->session->userdata('radio') == 'ws') { echo ' selected="selected"'; } ?>><?= __("WebSocket (Requires WLGate>1.1.10)"); ?></option>
+		            <option value="ws"<?php if ($this->session->userdata('radio') == 'ws') { echo ' selected="selected"'; } ?>><?= __("Live - ") . __("WebSocket (Requires WLGate>1.1.10)"); ?></option>
```
int2001: should be: `>=1.1.10`, not `>1.1.10`

AndreasK79:
@szporwolik sorry for the merge conflict. Please fix :) It will be good to have this merged into dev, to avoid too many of these conflicts. I will try to test this tonight.

szporwolik:
@AndreasK79 no problem an easy one!
@int2001 fixed

phl0 (approved):
I just tested this PR in my dev environment (which does not have CAT control). Waterfall display works and all the logic with clicking callsigns, changing bands / modes works. Thumbs up from my side.

AndreasK79 (approved):

AndreasK79:
Tested this a bit, and it worked here. I say it's ready to merge into dev for further use and testing.

szporwolik:
🎉 Thanks for support and - again - please excuse it's size.

DL2CC:
On my screen the background is constantly kind of flickering, like noise in a real waterfall.
Lacking real signal information there is no sense in showing noise flickering, it is just straining the eyes.
Can there be an option to permanently turn this off and just show a black background when there is no real spectrum signal source?

PR #2433, opened 2025-10-23
Introduce resubmission avoidance for contest logging

goodspeed34:
The contest logging page doesn't have resubmission avoidance so users could click the save button multiple times and get duplicated QSOs. What's more, this happends more commonly in contest because of the high pace.

The fix was borrowed from the implimentation in `assets/js/sections/qso.js'.

Ref: https://github.com/wavelog/wavelog/blob/e4d17a2a07be52b12420cb351b35b72ab4c9e8fa/assets/js/sections/qso.js#L158

However, this looks great but could be annoying for slow servers. It could also be implemented by blanking the qso info earlier so that the user wouldn't be able to click again. But this may trick users to think the QSO was already uploaded when it really isn't.


int2001 (approved):
good finding. works. tnx 4 contrib!

PR #2434, opened 2025-10-23
Websocket radio

int2001:
Adds the functionality to receive CAT-Information (QRG, Mode, Power) via Websocket from Wavelog-Gate, instead of polling Wavelog over and over (which can cause performance-problems).

- [x] Add Basic WebSocket-functionality to QSO-Form
- [x] Error-Catching, captions, etc.
- [x] Handling of formatted qrg and units
- [x] testing "old roundtrip" as well as new functionality
- [x] Adding the WebSocket functionality as well to Contest
- [x] Adding the WebSocket functionality as well to DXCluster

to be fixed:
- [x] Remember Websocket like other radio when switching pages

Needs WLGate >=V1.1.10 [WavelogGate](https://github.com/wavelog/WaveLogGate/releases/tag/v1.1.10) to test it. But: Should (and must) work also (the polling-way) with other 3rd-party-tools or older Versions of the Gate.

Important new Feature:
- if using the websocket, no key / url / whatever is needed to do CAT since Wavelog gets the CAT-Data directly from the client (without roundtrip via Wavelog). Localhost is doing the work.
- Affects ONLY CAT. If one wants to log - e.g. - wsjt-x QSOs via the Gate, key, url & Co. are still needed
- reduces the polling-load for CAT significant. Instead of fetching the CAT-Data from WL-Server every 3 seconds, WL-Page gets notified via the Websocket which QRG/Mode is on TRX
- Backwards-Compatibility. CAT-Data is posted (as before) to WL-Server. This PR is able to handle "classic CAT" as well as "Websocket"

### DIFF
--- a/application/views/interface_assets/footer.php
+++ b/application/views/interface_assets/footer.php
@@ -1468,7 +1468,7 @@ mymap.on('mousemove', onQsoMapMove);
 		    });
 
 		    cat2UI($('#frequency_rx'),data.frequency_rx,false,true,function(d){$("#band_rx").val(frequencyToBand(d))});
-		    cat2UI($('.mode'),data.mode,false,false,function(d){setRst($(".mode").val())});
+		    cat2UI($('.mode'),catmode(data.mode),false,false,function(d){setRst($(".mode").val())});
 		    cat2UI($('#sat_name'),data.satname,false,false);
 		    cat2UI($('#sat_mode'),data.satmode,false,false);
 		    cat2UI($('#transmit_power'),data.power,false,false);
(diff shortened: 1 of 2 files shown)


AndreasK79:
Tested with and without anything in the hardware interface.
Once I swap to websocket for radio, it now works. It also translates CWU/CWL modes to CW.
So far it's working good!

szporwolik:
I have some time over this weekend, if there are no critical bugs in this one I would kindly recommend to merge it, so I can work on DX Waterfall and Websocket integration. 👍🏻

PR #2438, opened 2025-10-24
Ignore keyup/input event when composing with IME, and trigger them when finished

goodspeed34:
The current keyup/input didn't handle composing events, which leads to:
  1. Repeating input when composing in IME with other language (qso.js);
  2. Incorrect sync between the callsign field and the search field (contesting.js).

The commit solves it by not doing anything when the user compsoing in IME and firing the event later when the user finished typing.

This isn't a big deal on PC but may be troublesome for mobile users, where they often have a default IME on everytime typing in a <input>.

A screen record of how this was behaving was attached in the thread. 

https://github.com/user-attachments/assets/078fb70f-1916-495a-873b-af853c4943af

### DIFF
--- a/assets/js/sections/contesting.js
+++ b/assets/js/sections/contesting.js
@@ -385,8 +385,17 @@ $("#callsign").on( "blur", function() {
 
 var scps=[];
 
+$("#callsign").on("compositionstart", function(){ this.isComposing = true; });
+$("#callsign").on("compositionend", function(e){
+	this.isComposing = false;
+	$(this).trigger("keyup");
+});
+
 // On Key up check and suggest callsigns
 $("#callsign").keyup(async function (e) {
+	// Prevent checking when the user's composing in IME
+	if (this.isComposing || e.isComposing) return;
+
 	var call = $(this).val();
 	if ((!((e.keyCode == 10 || e.keyCode == 13) && (e.ctrlKey || e.metaKey))) && (call.length >= 3)) {	// prevent checking again when pressing CTRL-Enter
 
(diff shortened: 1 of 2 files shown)


int2001:
Sorry, tried to understand that composing thing (and tried it on my iphone as well, with Chars like é or ß or ü). 

Could you explain a bit more whats happening here? Sorry, but i wasn't even able to understand the problem or to reproduce it.

goodspeed34:
> Sorry, tried to understand that composing thing (and tried it on my iphone as well, with Chars like é or ß or ü).
> 
> Could you explain a bit more whats happening here? Sorry, but i wasn't even able to understand the problem or to reproduce it.

On my iOS, when inputing the callsign in the default Chinese IME, it will put the text into the field before selecting a suggestion or entering the text as it.

Because of how the language works, the IME will put a space between the romanized input separating each Chinese word, this triggers the regex that trims the spaces making the IME discarding the input, casuing the spam in the first part of screen record.

Also it seems, when the user decided to input the text as it, as shown in the second part, the search field won't sync properly with the callsign field, because the keyup/keydown wasn't fired alongwith compositionend event.

This happens with Chinese language, not sure how other languages work 😳

int2001:
QSL.
since it seems nothing to destroy in "normal" use, i am going to approve it. 
but would be nice, if someone else of our team (@HB9HIL ? @AndreasK79 ? @phl0 ) could also have a look at it

phl0:
> QSL. since it seems nothing to destroy in "normal" use, i am going to approve it. but would be nice, if someone else of our team (@HB9HIL ? @AndreasK79 ? @phl0 ) could also have a look at it

Need some assistance on what to check exactly. No an apple pro here ... :)

PR #2459, opened 2025-10-28
[Location list] Listing all locations in it's own view

AndreasK79:
New button in station setup, which takes you to a new view.
The view lists all locations with most of the columns. Useful for our users with many locations.


int2001 (approved):
working here

Issue #2392, opened 2025-10-07
SOTA CSV Export option to filter out repeater QSOs


DJ3CE:
**What Feature is Missing?**
The CSV-Export function ("Other Export Options" - "SOTA CSV Export") gives the user some options to filter QSOs to export.

The SOTA-guidelines exclude QSOs via Repeaters, but the selection-box for propagation mode so far only features selecting a specific propagation mode for export, not excluding one.

**Why is This Feature Important?**
SOTA is a widespread program, which is recognised by wavelog by featuring their CSV format as an export option. But currently, the existing SOTA CSV Export does not allow users to safely export a valid file to upload to SOTA.

**Is This Feature Personal or Beneficial to Others as Well?**
Beneficial for all SOTA users with repeater-qsos during their activation.

I've prepared a possible [PR](/DJ3CE/wavelog/tree/sota_csv).

HB9HIL:
For easier comparison: 

https://github.com/wavelog/wavelog/compare/dev...DJ3CE:wavelog:sota_csv

int2001:
there is no Propagation-Mode "SOTARULES": https://www.adif.org.uk/316/ADIF_316.htm#Propagation_Mode_Enumeration

imho the only possible thing would be: Filtering out PropagationModes which are not suitable for SOTA.
But tbh: besides SAT and the SHF-Guys the people don't seem to use the Propmodes

DJ3CE:
I'd appreciate suggestions for a better wording. I see no reason to limit the GUI dropdown to the ADIF Enumeration, especially since there are cases in which you'd need a combination or (as in this case) a negation.

Filtering out propagation modes is exactly what the PR aims to do (cf. `Csv_model.php`) - maybe I'm missing your point here.

int2001:
Sorry, but we're adif compliant and we've no plans to change that.

What are you setting to PROP_MODE if you do meteor-scatter on a SOTA-Activation.

DJ3CE:
So you'd be fine with a checkbox to select s.th. like 'Filter out propagation modes not valid for SOTA'?

int2001:
the "not allowed prop_modes" for SOTA are static, aren't they?
if they're static simply add the filter to the SOTA-CSV-Export and everything is great.

or didn't i understand? Either there are prop_modes which aren't allowed for SOTA, or any PROP_MODE is allowed. why add a checkbox?

DJ3CE:
[...]

PR #2460, opened 2025-10-28
SOTA CSV Export

DJ3CE:
See #2392


DJ3CE:
Sure - how about adding 'All' as well, like done on the 'Accumulated Statistics'-Page for Sat?

int2001:
good idea. go ahead.

PR #2465, opened 2025-10-29
[QSO Entry] Added info to map where position comes from

AndreasK79:
From DXCC:
image

From Gridsquare:
image

- [x] Change text when grid is changed
- [x] Grid is not always fetched (race condition in code. Hopefully fixed by https://github.com/wavelog/wavelog/pull/2465/commits/347e9da27aae41e1fd5c7c4617c2873e372e2978?)
- [x] Resetting QSO entry does not remove text.
- [x] Needs translation
- [x] WWFF, POTA, SOTA, IOTA (?) has some logic to set location -> need to handle that too -> it is handled by the current state except IOTA (See https://github.com/wavelog/wavelog/pull/2465#issuecomment-3463709233)
- [x] Remove markers from maps where no grid is set (pan map to DXCC position but without marker)
- [x] Need to add on DXCC change
- [x] Fixed an error in get_entity in logbook_model, wrong kind of ticks

### DIFF
--- a/application/views/interface_assets/footer.php
+++ b/application/views/interface_assets/footer.php
@@ -68,12 +68,14 @@
     var lang_notes_duplicate_confirmation = "<?= __("Duplicate this note?"); ?>";
     var lang_notes_duplication_disabled_short = "<?= __("Duplication Disabled"); ?>";
     var lang_notes_not_found = "<?= __("No notes were found"); ?>";
-	var lang_qso_note_missing = "<?= __("No notes for this callsign"); ?>";
-	var lang_qso_note_toast_title = "<?= __("Callsign Note"); ?>";
-	var lang_qso_note_deleted = "<?= __("Note deleted successfully"); ?>";
-	var lang_qso_note_created = "<?= __("Note created successfully"); ?>";
-	var lang_qso_note_saved = "<?= __("Note saved successfully"); ?>";
-	var lang_qso_note_error_saving = "<?= __("Error saving note"); ?>";
+    var lang_qso_note_missing = "<?= __("No notes for this callsign"); ?>";
+    var lang_qso_note_toast_title = "<?= __("Callsign Note"); ?>";
+    var lang_qso_note_deleted = "<?= __("Note deleted successfully"); ?>";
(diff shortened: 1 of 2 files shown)


szporwolik:
Love it! ❤️  As I was working in the similar area of the code yesterday, please note that there is a logic implemented to override grid basing of WWFF, POTA, SOTA, IOTA (?) references; I would consider handling that case as well, so user knows exactly where the location came from.

AndreasK79:
> Love it! ❤️ As I was working in the similar area of the code yesterday, please note that there is a logic implemented to override grid basing of WWFF, POTA, SOTA, IOTA (?) references; I would consider handling that case as well, so user knows exactly where the location came from.

Thanks for the input. I'll add those to the list above!

phl0:
Changed code so that markers are only set if we have a grid. Changed maps on dashboard and logbook pages as well. Prevents stacking of markers for QSOs which don't have any grid set.

szporwolik:
I actually have small fix to suppress console warnings when leaflet is used at QSO page, please take a look here: https://github.com/wavelog/wavelog/commit/01284e7475255549f91ae1764fc1ccd170ebf5ad

You can patch your code with it if you want. 👍🏻

phl0:
> I actually have small fix to suppress console warnings when leaflet is used at QSO page, please take a look here: [01284e7](https://github.com/wavelog/wavelog/commit/01284e7475255549f91ae1764fc1ccd170ebf5ad)
> 
> You can patch your code with it if you want. 👍🏻

What warnings were shown in console pls?

szporwolik:
Latest Chrome at Windows 11:
image

You can rise additional ones by clicking Gridsquares checkbox off and on.

phl0:
Location updates from SOTA, POTA and WWFF work. IOTA functions do not set a locator.

phl0 (approved):

AndreasK79:
> Latest Chrome at Windows 11: image
> 
> You can rise additional ones by clicking Gridsquares checkbox off and on.

Please make a separate PR only for this fix :)

PR #2466, opened 2025-10-29
Remove CFD-Export, since it's deprecated by DARC

int2001:
Remove CFD-Export, because it's deprecated by DARC.
People who want to use that List can still use [mostwanted.exe](http://www.dl7ucx.de/mwl.html)


phl0:
Removed more stuff that was added back when https://github.com/wavelog/wavelog/pull/117 was merged.

phl0 (approved):

PR #2468, opened 2025-10-30
Fix get plaincall logic

### LINKED ISSUE #1227: Strange glitch
### Discussed in https://github.com/wavelog/wavelog/discussions/1200

<sup>Originally posted by **MikleBas** November 13, 2024</sup>
Today I wanted to call a station, I enter its callsign in the window, well, the result can be seen in the screenshot
it shows that I already had QSO with him, although it is clear that there are no matches      ¯\_(ツ)_/¯

Screenshot</div>

AndreasK79:
Fixes #1227.

In some rare cases, the plaincall logic would fail, like R4FBJ/0/P. The plaincall would return as 0, which is not correct.
With this fix, R4FBJ is returned.

Hopefully I didn't break anything while refacoring this. I've tested with several callsigns, but I might have overlooked some cases.

### DIFF
--- a/application/libraries/Callbook.php
+++ b/application/libraries/Callbook.php
@@ -170,19 +170,20 @@ class Callbook {
 
 	function get_plaincall($callsign) {
 		$split_callsign = explode('/', $callsign);
-		if (count($split_callsign) == 1) {				// case F0ABC --> return cel 0 //
-			$lookupcall = $split_callsign[0];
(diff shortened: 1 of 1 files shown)


int2001 (approved):
Made a couple of tests. e.g.:
- HB9/DF2ET/P

Still shows all other DF2ET QSOs

Tested some known special calls with `/3` etc. as well.

Looks good.

PR #2463, opened 2025-10-28
[QSO page] Minor fixes, tweaks, UX improvements and one noticeable bug fixed

szporwolik:
As I needed a short break from DX Waterfall, I decided to address a few smaller issues on the QSO page that personally bothered me.  These are mostly minor improvements, with one change that might require some discussion — see the note below.

---

### Improvements
- Replaced some notifications with Bootstrap toasts — this prevents the entire GUI from shifting when notifications appear.  
  [`68f43b33e5ac2780f579fbf60669792f27a17128`](https://github.com/wavelog/wavelog/commit/68f43b33e5ac2780f579fbf60669792f27a17128)
- The badges below **Callsign** and **Distance/Bearing** have now space reserved, avoiding unwanted GUI shifts.  
  [`05ba6609a06ce4ddfc10a0f1ae228cdd57daf1f8`](https://github.com/wavelog/wavelog/commit/05ba6609a06ce4ddfc10a0f1ae228cdd57daf1f8)
- Added an `mailto:` button that appears when an email address is provided.  
  [`05ba6609a06ce4ddfc10a0f1ae228cdd57daf1f8`](https://github.com/wavelog/wavelog/commit/05ba6609a06ce4ddfc10a0f1ae228cdd57daf1f8)
- Improved the **Callsign Notes** expand/collapse behavior — the whole bar is now clickable, and the icon was updated.  
  [`0549d276501c29de25989bbf5f63c3fe3a93950d`](https://github.com/wavelog/wavelog/commit/0549d276501c29de25989bbf5f63c3fe3a93950d)  
  [`592592a866774fe2232b5d6f28055c3118be376e`](https://github.com/wavelog/wavelog/commit/592592a866774fe2232b5d6f28055c3118be376e)
- The Leaflet map now features rounded corners for visual consistency with the rest of the page.  
  [`c55e2de6a2b7fea3cea4068b8dfef5d0b3cefb5b`](https://github.com/wavelog/wavelog/commit/c55e2de6a2b7fea3cea4068b8dfef5d0b3cefb5b)
- Added several previously hardcoded strings to the translation table.  
  [`5f9494c2d7cb2d43182d25df89cce8dae340fcc2`](https://github.com/wavelog/wavelog/commit/5f9494c2d7cb2d43182d25df89cce8dae340fcc2)

### Bug Fixes
- Fixed an issue where logging a QSO during an active callsign data fetch caused multiple errors.  
[...]

PR #2469, opened 2025-10-30
[Callbook] Option to prioritize 3rd party callsign lookup over database search

szporwolik:
As discussed in https://github.com/wavelog/wavelog/issues/2463#issuecomment-3466520786 - by default Wavelog's database (previous QSOs) is used to fetch call-sign information. 

This PR gives users possibility to choose which of the mechanism is the preferred one to be used as 1st priority.

There is a change in the user options, so per code's comment: @int2001 @phl0 please feel notified about. 😄 

Technically:
- new option with the proper description was added to the profile edit page
- handling of the new option was added in the User model & Controller
- Logbook controller was updated to support new possibility

By default option is set to "Yes", meaning the current "Use database" approach is preffered.


int2001:
tnx fer notifying ;)
Background is: mantaining a huge instance with a special SSO. Everytime the function gets a new param i'll have to adjust the SSO-Controller there.

Going to check this PR after all else PRs are merged

AndreasK79 (approved):

AndreasK79:
Works on testing here.

Setting options on/off works.
Test lookups in QSO entry. Fetches from callbook when callbook is priority. DB when that is a priority.

phl0:
My last comment was a mistake and deleted. Instead fix proposed in https://github.com/wavelog/wavelog/pull/2483.

PR #2472, opened 2025-10-30
[CAT] Moved javascript into seperate js file

szporwolik:
After recent works on QSO page, implementation of WebSockets and major refactor of the CAT code in the ```footer.php``` this PR is a natural next step.

This may look big, but's it's basically relocation of 716 lines of code, so no worries. 😅 

### Changes
- CAT (inc. WebSocets) code was moved into separate ```cat.js``` - this is mainly to improve code structure
- CAT functionality is enabled only when we are in the LIVE data input (Live QSO & Live Contents logging)

### Further TODO (not in the scope of this PR)
Once https://github.com/wavelog/wavelog/pull/2436 and this one are merged, we need to re-work ```bandmap/list```, so it uses ```cat.js``` .

### DIFF
--- a/application/views/interface_assets/footer.php
+++ b/application/views/interface_assets/footer.php
@@ -1401,24 +1401,25 @@ mymap.on('mousemove', onQsoMapMove);
 
 <?php } ?>
 <?php if ( $this->uri->segment(1) == "qso" || ($this->uri->segment(1) == "contesting" && $this->uri->segment(2) != "add")) { ?>
+	<!--- Frequency input functionality --->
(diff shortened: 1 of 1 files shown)


AndreasK79:
CAT still works :)

szporwolik:
> CAT still works :)

🤣 that was the idea! Thank you review!

But seriously please check contenting, I have never used it other than testing.
The Band List has CAT hard-coded, this will be handled in https://github.com/wavelog/wavelog/pull/2479 and unified with QSO/contesting pages.

AndreasK79:
Tested contesting, and that works too!

PR #2473, opened 2025-10-31
Prevent Re-Lookup not only on Reset-Click but also on ESC

int2001:
When pressing ESC (with a pre-typed call), a lookup is triggered (which is wrong / overhead).
This problem was already solved for pressing the Reset-Button, but not for ESC.

this fixes it


phl0:
Tested. works.

PR #2474, opened 2025-10-31
[DX Cluster] Backend Performance Optimization

szporwolik:
## The Problem
Loading the bandmap/DX Waterfall band was slow - each request made **600+ database queries** for just 100 spots. This caused multi-second page loads and high server load.

### 1. Batch Database Queries
Instead of checking each spot individually (6 queries × 100 spots), we now check all spots in a single query.
- Created `get_batch_spot_statuses()` - checks worked/confirmed status for all spots at once
- Created `get_batch_last_worked()` - gets "last worked" info for all callsigns in one query

### 2. Smart Caching
Results are cached for 59 seconds (matching typical DXCache server's cache time), so repeat requests are instant.

### 3. Additional Optimizations
- DXCC lookups are memoized to avoid duplicate processing
- Removed duplicate cache loading
- Added early filtering to skip unnecessary processing
- Park references (SOTA/POTA/IOTA/WWFF) now extracted on backend (this will improve DX Waterfall, but.. not only)
- Contest spot detection added server-side (more to come...)

### 4. Error Handling
Added proper validation and error handling for:
- Network failures when fetching spots
- Invalid JSON data
- Missing spot properties
- Empty database results

## Performance Impact
- **Before**: 600+ queries per request (100 spots), several seconds load time; see lines 133-140 in current ```Dxcluster_model``` 
- **After**: 2-3 queries per request, instant loading
- **With cache**: 0 queries, instant response

## Additionaly
This is for ca. 250 spots fetch - current master branch and this PR:
dx_current
dx_new

The gain will be of course more visible on slower/loaded instances.

The cache is build like ```spotlist_{$band}_{$maxage}_{$de}_{$mode}_{$user_id}_{$logbook_id}```, so it's seperate per user, per logbook even at the same instance.

In regards ```enrich_spot_metadata``` I need this for further development (please trust me on that 😄 ) and after minor change this will also move some of the logic does in DX Waterfall to the cached server side. But the most optimal way would be to have similar logic implemented at DX Cluster Cache instance one day and treat this code as a fallback.

This is how the enriched spot looks like when returned to frontend:
```
[...]
        "message": "FT4 POTA @ GB-0265 South Downs National Park (GB-ENG)",
[...]
        "dxcc_spotted": {
            "cont": "EU",
            "entity": "England",
            "flag": "\ud83c\udff4\udb40\udc67\udb40\udc62\udb40\udc65\udb40\udc6e\udb40\udc67\udb40\udc7f",
            "dxcc_id": "223",
            "lotw_user": false,
            "lat": "51.5",
            "lng": "-0.1",
            "cqz": "14",
            "pota_ref": "GB-0265",
            "pota_mode": "FT8",
            "sota_ref": "",
            "iota_ref": "",
            "wwff_ref": "",
            "isContest": false
[...]
 ```

### CODE REVIEW

File: application/models/Logbook_model.php [hunk 1]
```diff
@@ -2693,6 +2693,252 @@ function check_if_dxcc_cnfmd_in_logbook($dxcc, $StationLocationsArray = null, $b
...
+	 */
+	function get_spot_status($callsign, $dxcc_id, $continent, $logbooks_locations_array, $band = null, $mode = null) {
+		$user_default_confirmation = $this->session->userdata('user_default_confirmation');
+		$qsl_where = $this->qsl_default_where($user_default_confirmation);
+
+		// Build band filter
+		$band_sql = '';
+		$band = ($band == 'All') ? null : $band;
+		if ($band != null && $band != 'SAT') {
+			$band_sql = " AND COL_BAND = " . $this->db->escape($band);
```
int2001: please don't do this. don't put variables directly into SQL, even if they're 'PHP-Escaped'.
use binding instead. most of our queries are already refactored to bindind

application/models/Logbook_model.php
```diff
@@ -2693,6 +2693,252 @@ function check_if_dxcc_cnfmd_in_logbook($dxcc, $StationLocationsArray = null, $b
...
+		// Escape parameters
+		$station_ids = implode(',', array_map('intval', $logbooks_locations_array));
+		$callsign_escaped = $this->db->escape($callsign);
+		$dxcc_escaped = $this->db->escape($dxcc_id);
+		$continent_escaped = $this->db->escape($continent);
+
+		// Single optimized query to get all statuses
+		$sql = "
+			SELECT
+				MAX(CASE WHEN COL_CALL = {$callsign_escaped} THEN 1 ELSE 0 END) as worked_call,
```
int2001: same here. please use binding

application/models/Logbook_model.php
```diff
@@ -2693,6 +2693,252 @@ function check_if_dxcc_cnfmd_in_logbook($dxcc, $StationLocationsArray = null, $b
...
+		}
+
+		$user_default_confirmation = $this->session->userdata('user_default_confirmation');
+		$qsl_where = $this->qsl_default_where($user_default_confirmation);
+
+		// Build band filter
+		$band_sql = '';
+		$band = ($band == 'All') ? null : $band;
+		if ($band != null && $band != 'SAT') {
+			$band_sql = " AND COL_BAND = " . $this->db->escape($band);
```
int2001: use binding (please grep full PR for that issue - many places)

int2001 (changes requested):
so far ONLY reviewed from the technical side of the PR.

Regarding the logical one: not touched/not reviewed so far. only thing i noticed: yes, should be faster, but redundant to already implemeneted functions. we try to reuse functions whereever possible, to keep the (future) change-effort low. better use (and perhaps optimize) existing functions than creating new ones.

application/models/Dxcluster_model.php
```diff
@@ -25,6 +34,19 @@ public function __construct() {
 	public function dxc_spotlist($band = '20m', $maxage = 60, $de = '', $mode = 'All') {
 		$this->load->helper(array('psr4_autoloader'));
 
+		// Load cache driver once
+		$this->load->driver('cache', array('adapter' => 'file', 'backup' => 'file'));
```
int2001: question:
i already thought about the same during the past, but discarded my idea.
reason: will it scale on huge instances (always keep in mind: there are instances with >1000 Users out there)

szporwolik: I was wondering the same. In theory this is a native Codeigniter mechanism developed to do exactly this what we are doing here. We are using this in a few places already (this was the final trigger to put it in for me):

image

Various places were saying to use this for small/medium instances and migrate to Redis cache if things need to be scaled - but usually they are referring to millions, not thousand of requests. The transition is quite simple.

To be honest I simply don't know - I can tell you that this PR will significantly reduce load of your server coming from DX cluster calls, the question is how the disk usage will rise and if you will get into troubles.
I finally understood what large scale instance you are always referring to, having this is mind since. 😄 

I have no feelings about this cache we can leave it in or remove - it's up to you. The rest of the code is anyhow making everything much faster, cache was a cherry at the top of the cake. 😉

application/models/Logbook_model.php
szporwolik: Yes, completly forgot about - please excuse; fixed!

application/models/Logbook_model.php
szporwolik: Yes, completly forgot about - please excuse; fixed!

application/models/Logbook_model.php
szporwolik: Fixed

application/models/Dxcluster_model.php
szporwolik: I had a coffee, fixed the binding stuff and looked at the cache topic again.

This is how it looks in my app/cache after a lot of testing:
image

If 4k users will start to play with the new code, you will definitely end up with a thousands of files for sure; but the server load will be much smaller, the trade off will be the disk usage (reads/writes, not the space - that's usually not an issue these days 😉).

On the other hand - see the Bandplan List with this PR, it's fast, very fast. My main dev instance is runing on my NAS, it's slow, but I prefer it this way - on fast servers, you don't always catch what shall be improved.
Prior this PR pull of DX Spots for new band can go up to 15-30 seconds, with the PR, I'm usually < 1 sec., it's crazy;

What about config.php switch to use the filecache for DX Cluster or not? From strictly engineering curiosity I would give this a try at your large instance and if things would go not as planned, the admin could quickly switch off this function.
What do you think ?

int2001: Good idea. I'd like to test it. Have admin on one of the hugest instances.

szporwolik: Perfect! Added config.php settings for that : ```enable_dxcluster_file_cache``` .

After very short testing of ```/dxcluster/spots/20m/120/EU/All?_=1762021324733``` - 13 spots:
- cache disabled: ca. 700ms on every page load
- case enabled: ca. 700ms on initial load and 300ms for following ones

szporwolik: OK, made some more quick tests; tested in short time at index.php/bandmap/list assuming that the same data was received from DX Cluster cache:

| Version / Code Branch | Test Environment | Cache Status | Run 1 | Run 2 | Notes / Reference |
|------------------------|------------------|---------------|--------|--------|-------------------|
| Current master         | NAS (Apache + PHP8) | N/A | 1.8s | 1.8s | Baseline performance |
| Current dev            | NAS (Apache + PHP8) | N/A | 1.5s | 1.5s | Includes first improvements ([commit 69db074](https://github.com/wavelog/wavelog/commit/69db0741862a9ec4253cf9d540178d0fc2c2e308)) |
| This PR | NAS (Apache + PHP8) | Cache off | 0.7s | 0.7s | Improved performance even without cache |
| This PR | NAS (Apache + PHP8) | Cache on | 0.7s | 0.3s | Best performance with cache enabled |

6x speed improvement when we have cache hit - hard to believe, too good to be true!? 🫨 

@int2001: what are your experiences with this large instance, what's the usual time you would expect to have band-plan page filled with spots ? Or when changing band?

application/models/Logbook_model.php
```diff
@@ -2693,6 +2693,278 @@ function check_if_dxcc_cnfmd_in_logbook($dxcc, $StationLocationsArray = null, $b
...
+	 *
+	 * @param string $callsign Spotted callsign
+	 * @param int $dxcc_id DXCC ID
+	 * @param string $continent Continent code
+	 * @param array $logbooks_locations_array Station IDs
+	 * @param string $band Band (e.g., '20m', 'All')
+	 * @param string $mode Mode (e.g., 'SSB', 'CW', 'All')
+	 * @return array ['worked_call', 'worked_dxcc', 'worked_continent', 'cnfmd_call', 'cnfmd_dxcc', 'cnfmd_continent']
+	 */
+	function get_spot_status($callsign, $dxcc_id, $continent, $logbooks_locations_array, $band = null, $mode = null) {
```
int2001: never used?

application/models/Logbook_model.php
```diff
@@ -2693,6 +2693,278 @@ function check_if_dxcc_cnfmd_in_logbook($dxcc, $StationLocationsArray = null, $b
...
+		$callsigns_array = array_keys($callsigns);
+		$dxccs_array = array_keys($dxccs);
+		$continents_array = array_keys($continents);
+
+		$callsigns_placeholders = implode(',', array_fill(0, count($callsigns_array), '?'));
+		$dxccs_placeholders = implode(',', array_fill(0, count($dxccs_array), '?'));
+		$continents_placeholders = implode(',', array_fill(0, count($continents_array), '?'));
+
+		// Add values to bind params
+		$bind_params = array_merge($bind_params, $callsigns_array, $dxccs_array, $continents_array);
```
int2001: Here's something fishy. seems like the order of the bindings won't fit.
Have a look at the resulting SQL.

int2001 (changes requested):
started to check this beauty.
doesn't work so far. see my code-comments

AndreasK79:
@szporwolik this is of course much faster, however something is wrong.
The DXCC-status doesn't really work, as all my spots are green.

I set a breakpoint in get_batch_spot_statuses, and checked query + bindings.
Seems they are shifted, and that might be the reason for this not working as it should.

```

			SELECT
				COL_CALL as callsign,
				COL_DXCC as dxcc,
				COL_CONT as continent,
				MAX(CASE WHEN (COL_QSL_RCVD='Y' OR COL_LOTW_QSL_RCVD='Y') THEN 1 ELSE 0 END) as is_confirmed
			FROM logbook
			WHERE station_id IN (1)
			AND (COL_CALL IN (?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?)
				OR COL_DXCC IN (?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?)
				OR COL_CONT IN (?,?,?,?,?,?))
			 AND COL_BAND = ?
			
			GROUP BY COL_CALL, COL_DXCC, COL_CONT
```
While params are like this:

```
array (
  0 => '20m',
90 => '',
```

I didn't post the whole array, but it seems band is wrong, and the first question mark is for COL_CALL in,
and the last one is for COL_BAND = ?.

I used band = ALL here, and spots DE = Europe.

szporwolik:
Yes, I failed miserably with this one. 

Please excuse not testing it properly, I was so focused on having this running as quick as possible that forgot about what this is really about (cache did not helped during testing). Rand() generator of the data would probably run even faster... 😞 

Back to Workshop! 🧑‍🏭

application/models/Logbook_model.php
szporwolik: Yes, orphaned in the process. Will check and get rid of it.

application/models/Logbook_model.php
szporwolik: Sir! Good catch, it should be ```$bind_params = array_merge($callsigns_array, $dxccs_array, $continents_array);```

szporwolik:
Ok, refactored and hopefully better tested this time. I'm getting some meaningful data now:
image

### Controller (Dxcluster.php)
- Added input cleaning with `xss_clean()` for the band, mode, and continent filters - basically stopping anyone from injecting anything
- Made sure the age parameter is actually a number instead of whatever else
- Fixed error response `'{ "error": "not found" }'` to use proper JSON encoding
- Ditched `JSON_PRETTY_PRINT` because it just adds white-spaces 😄 
- Better check for empty results so we don't pretend empty arrays are successful responses

### Model (Dxcluster_model.php)
- Fixed a dumb bug where mode filtering was comparing a string to true/false instead of actually checking the boolean - oops!
- Added CQ Zone info to both the spotted station and spotter data (this is something I need for https://github.com/wavelog/wavelog/pull/2479)

### Database Layer (Logbook_model.php)
- **BIG FIX**: Closed a SQL injection hole where station IDs weren't being properly bound
- Completely rewrote `get_batch_spot_statuses()` because the original logic was broken - you can't mix OR conditions with GROUP BY like that and expect it to work
- Fixed the parameter order in `get_batch_last_worked()` - they were all mixed up and didn't match the SQL
- Added handling for when multiple QSOs have the exact same timestamp (rare but it happens)

Bottom line: I hope this will be working now during your review. 😅

AndreasK79:
@szporwolik as far as I can tell and test, this works as it should here. I know @int2001 wanted to do some more testing regarding cache etc, so we will wait for him,.

szporwolik:
That's perfectly fine. Can you see any performance improvement or this is not visible at your testing instance ?

AndreasK79:
Yes, I see some improvement. Perhaps half a second when looking in devtools, going back and forth on bands. I am guessing this will help even more on larger instances.

PR #2475, opened 2025-10-31
Clear refs before filling spot data

phl0:
We need to clear reference fields on spot click to prevent stale data from being logged if the new spot clicked does not provide ref information. Maybe @szporwolik can check here as well.

### DIFF
--- a/assets/js/dxwaterfall.js
+++ b/assets/js/dxwaterfall.js
@@ -1204,6 +1204,8 @@ var DX_WATERFALL_UTILS = {
                 wasPreventLookupSet = true;
             }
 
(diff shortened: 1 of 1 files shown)


szporwolik:
Let me check and work on that - the solution by @phl0 will work, but probably we will lose fetched name and location of the POTA and SOTA reference.

phl0:
Oh sorry. It was merged to dwv already. But we can fix it anyway if needed. Let us know what you find.

szporwolik:
No worries: https://github.com/wavelog/wavelog/pull/2476 :)

PR #2471, opened 2025-10-30
[Notification] Moved toast notification code to general files

szporwolik:
Moved toast container to ```footer.php``` and javascript code to ```common.js```. 
This will make toast notification available at all pages with the centralized creation code for unification. 

### Usage
Examples are included at the function definition to ensure common usage:
```
	showToast('Saved', 'Your data was saved!', 'bg-success text-white', 3000);
	showToast('Error', 'Failed to connect to server.', 'bg-danger text-white', 5000);
	showToast('Warning', 'Please check your input.', 'bg-warning text-dark', 4000);
	showToast('Info', 'System will restart soon.', 'bg-info text-dark', 4000);
```

Above will display:
image

### How to test
After page is loaded (any) you can copy example into browser's console to trigger notification or check Add QSO page.

PR #2479, opened 2025-11-01
[DX Cluster] Bandplan List Update

szporwolik:
## Goal
The main goal is to update the **Bandplan list** functionality, addressing multiple user requests and aligning this view with other DX Cluster tools.  
It should be convenient to use, easy to read, and—thanks to the API backend—significantly faster than the current implementation.

### Prototype
screenshot

Note: what you see at my screens is my own tweaked DX Cluster cache, I proposed the changes in https://github.com/int2001/DXClusterAPI/issues/31 to improve SOTA spot creation.

## Tasks
- [x] Merge CQ zone logic
  - [x] Wait for https://github.com/wavelog/wavelog/pull/2474 to be merged  
- [x] Implement toasts
  - [x] Wait for https://github.com/wavelog/wavelog/pull/2471 to be merged
- [x] Re-do CAT handling at this page using common libraries
  - [x] Wait for https://github.com/wavelog/wavelog/pull/2472 to be merged
- [x] Bugs and features
  - [x] Sending log to QSO window shall also populate references (if we have any)
  - [x] User settings - band and modes shall affect page behavior
  - [x] Contest Only behavior is wrong, this shall behave like LoTW filter 
  - [x] Small highlight of new spots (as in the old code)
  - [x] Frequency marker (as in the old code)
  - [x] Table broken when using translations (German, Polish etc.)
  - [x] Css breaking DXCC quick lookup table
- [x] Re-work how/what filters are send to API calls
  - [x] Filter by continent logic shall be improved
  - [x] Reduce server load by possibly limiting the query size to minimum (ex. when tied to radio only one band shall be pulled)
- [x] Review feedback
  - [x] Button to turn on and off  - CAT band tracking; expected behavior:
   - ON (default): Follow CAT // Only CAT-Band shown
   - OFF: Follow CAT // All Bands shown
  - [x] Check font in the filters popup
  - [x] Disable CAT pooling when CAT control is OFF
  - [x] Contest flag - contest name in the tooltip and improved detection
  - [x] Date of previous QSO - broken at some instances see https://github.com/wavelog/wavelog/pull/2479#issuecomment-3478016809
  - [x] Clear text search icon
  - [x] Add DX quick filter
  - [x] Add ITU zone popup
  - [x] Add Continent popup
  - [x] Sorting of the table - freq is always numeric
  - [x] Consider using leaflet.geodesic.js
  - [x] Send spot to : "Post QSO" or "Live QSO" - decision
- [x] Ready for review
  - [x] Remove cache prevention tricks
  - [x] Finalize UI prototype
  - [x] Translation support
  - [x] Code refactoring (this will be massive task)
  - [x] Remove debug logs
  - [x] Complete all previous tasks
  - [x] Final testing

### DIFF
--- a/application/controllers/Bandmap.php
+++ b/application/controllers/Bandmap.php
@@ -74,15 +74,22 @@ class Bandmap extends CI_Controller {
 
 	// Get user's favorite bands and modes (active ones)
 	function get_user_favorites() {
+		session_write_close();
+
 		$this->load->model('bands');
 		$this->load->model('usermodes');
 
 		// Get active bands
 		$activeBands = $this->bands->get_user_bands_for_qso_entry(false); // false = only active
 		$bandList = [];
-		foreach ($activeBands as $group => $bands) {
-			foreach ($bands as $band) {
-				$bandList[] = $band;
+
+		if (is_array($activeBands)) {
+			foreach ($activeBands as $group => $bands) {
+				if (is_array($bands)) {
+					foreach ($bands as $band) {
+						$bandList[] = $band;
+					}
+				}
 			}
 		}
 
@@ -94,14 +101,16 @@ class Bandmap extends CI_Controller {
 			'digi' => false
 		];
 
-		foreach ($activeModes as $mode) {
-			$qrgmode = strtoupper($mode->qrgmode ?? '');
-			if ($qrgmode === 'CW') {
-				$modeCategories['cw'] = true;
-			} elseif ($qrgmode === 'SSB') {
-				$modeCategories['phone'] = true;
-			} elseif ($qrgmode === 'DATA') {
-				$modeCategories['digi'] = true;
+		if ($activeModes) {
+			foreach ($activeModes as $mode) {
+				$qrgmode = strtoupper($mode->qrgmode ?? '');
+				if ($qrgmode === 'CW') {
+					$modeCategories['cw'] = true;
+				} elseif ($qrgmode === 'SSB') {
+					$modeCategories['phone'] = true;
+				} elseif ($qrgmode === 'DATA') {
+					$modeCategories['digi'] = true;
+				}
 			}
 		}
 
--- a/application/views/interface_assets/footer.php
+++ b/application/views/interface_assets/footer.php
@@ -58,6 +58,7 @@
     var lang_general_word_please_wait = "<?= __("Please Wait ..."); ?>";
     var lang_general_states_deprecated = "<?= _pgettext("Word for country states that are deprecated but kept for legacy reasons.", "deprecated"); ?>";
     var lang_gen_hamradio_sat_info = "<?= __("Satellite Information"); ?>";
+
     var lang_notes_error_loading = "<?= __("Error loading notes"); ?>";
     var lang_notes_sort = "<?= __("Sorting"); ?>";
     var lang_notes_duplication_disabled = "<?= __("Duplication is disabled for Contacts notes"); ?>";


int2001:
good idea.
put an extra-button to it. users are already requested to see "All Bands" while CAT is turned on.

Idea a button to toggle: 
- ON (default): Follow CAT // Only CAT-Band shown
- OFF: Follow CAT // All Bands shown

BTW: Its again very huge. we need time for proper review on those things.

szporwolik:
> BTW: Its again very huge. we need time for proper review on those things.

Yes, but this time it was controlled. I have divided API and UI changes to keep those separate. 
It huge but it's complete overhaul of the page, there was not really how to do it step by step. I really do try to keep everything organized this time. 😉 

Let me work on that, code is still total spaghetti don't even bother with it yet. 
Now it's time for the functionality and behavior discussion. 

Thanks @int2001 for initial ideas - will do. 

If somebody would like to have a cup of coffee and click through it, all feedback is highly appreciated.

AndreasK79:
I gave this a test, and I like it very much. Well done!

While testing, I made a note of a few things:

- The font size in the filter form is smaller than the rest of the page. Any reason for that? Perhaps keep it the same as the rest of the text/page?
- imageShow which contest in mouseover? I didn't see it in the comments
- imageSeems date for previous QSO is wrong. Perhaps an issue with dateformat?
- When disabling CAT control, should it stop polling the db? Yes, I know about websocket, but for the people not using that :)

int2001:
i just tested it. What a beautiful update to my old UI. it needed that ;)
atm no things found besides those andreas already found. it's draft - so i'll wait for review until you tell us: "further tests needed"

szporwolik:
Thanks! Still a lots to do.

@int2001 not sure how the SAT spots are handled by the cache and there is not many to check. Shall we care, if so - any ideas how to handle it ?

int2001:
SAT is an own playground. Even if i also do a lot of LEO/GEO-SAT, @phl0 is our "SAT-Guru".
my 2 Ct. on SAT-Spots: SAT is most of the time done via special socialmedia-groups and not via Clusters.

Of course there are SAT-Spots from time to time. but i personally know no SAT-OP (and i know a lot of them ;) who is using a classic cluster.

phl0:
Same view as @int2001. LEO satellites have visibility of max. 15 - 20 min. Guess no one is watching the cluster while working the birds as you usually have one hand less than needed to operate satellites anyway...

szporwolik:
LOL, thanks - that makes things easier (or at least does not add additional complexity 😉 )

int2001:
Came just to my mind: 
What about selecting more than one band or selecting "worked, not confirmed" and "not worked" in parallel?

Nothing urgent. Just as idea

szporwolik:
@AndreasK79 
- reverted the text size to default, the text in multi selects is smaller - i'm trying to keep this small to not overflow on smaller monitors; per my experience some hams have tendency to use surprisingly low-res displays ✅ 
- well contest detection is done in the ```enrich_spot_metadata``` in ```Dxcluster_model``` I will try to improve it in the future, but TBH it always will be simple regex, not sure how good we will get with this; I need to think about and try to understand how others are doing that, I know that other software is detecting that accurately not sure how -> TBD
- will check the dates (thanks!) not really sure at which step It's messed up - it's shown properly on my dev instance, but I usually switch it to run proper YYYY-MM-DD format 😄-> TBD 
- When disabling CAT control, should it stop polling - done ✅ 

SAT logic completely gone

@int2001 just pushed the last commits with the new set of "quick filters" :
image
Most of you guys are probably much more seasoned HAM ops then myself, whatever you want will get there. 👌🏻 From user perspective, I would like just to keep this in this single line together with text field as I think this is most convinient.

> What about selecting more than one band or selecting "worked, not confirmed" and "not worked" in parallel?

You can already achieve this: more bands can be selected at once either by clicking or using Advanced Filters popup. Same goes for DXCC. Also the star icon will apply your active bands/modes from settings

Per @AndreasK79 comment - once CAT is enable, radio updates come and you are tied to band selected at TRX.

AndreasK79:
@szporwolik leave the filter font size like this. I discovered it was my own custom theme that makes the fonts smaller.

int2001:
> > What about selecting more than one band or selecting "worked, not confirmed" and "not worked" in parallel?
> 
> You can already achieve this: more bands can be selected at once either by clicking or using Advanced Filters popup. Same goes for DXCC. Also the star icon will apply your active bands/modes from settings

Yes, seen it. I had to press ALT while clicking "worked, not confirmed" and "not worked". Works so far. Didn't expect the ALT as modifier key for checkboxes. but once you know... ;)

int2001:
just oberved: clicking on a Spot opens "Post QSO" and not "Live QSO". intended?

szporwolik:
Intended for now - to be further discussed/tested. The issue is that if one has open both pages with radio control enabled, I'm not sure if there will be no collisions. Need to test and check - with the new DX Waterfall the issues are less likely (freq field is not controlling radio anymore).

PR #2481, opened 2025-11-02
Drop LoTW login errors to error log and simplify code

phl0:
Add dropping of LoTW login errors to error log in order to ease debugging. Also simplified code as filling an (extra) array with credentials seems useless here.


int2001 (approved):
good one. makes debugging a bit simpler

PR #2483, opened 2025-11-02
Option cannot be written to the auth table on user creation

phl0:
https://github.com/wavelog/wavelog/pull/2469 introduced a bug by trying to write user_qso_db_search_priority option to the auth table. As this is a user_option it fails.


int2001 (approved):

szporwolik:
@phl0 thanks! I even remember making this mistake and wondering why some other options like solar is not there... 🤦🏻‍♂️

phl0:
> @phl0 thanks! I even remember making this mistake and wondering why some other options like solar is not there... 🤦🏻‍♂️

Never mind. The code is not perfect at all. We find bugs every day =)

PR #2486, opened 2025-11-04
[Station Location] Import / export

AndreasK79:
Seems many people have many locations, and move between instances.

Added export / import buttons on the station list page. Uses a simple json format.
Dupe check is implemented.
Checks on the following columns:
- station_profile_name
- station_callsign
- station_gridsquare
- station_city
- station_iota
- station_sota
- state
- station_cnty
- station_dxcc
- station_wwff
- station_pota
- station_sig
- station_sig_info
- user_id

What this does not do (manual job)
- Create a logbook
- Set the logbook

Todo:
- [x] Added xss_clean
- [x] load user option model
- [x] 1 k limit
- [x] Make import work when having no locations

### DIFF
--- a/application/controllers/Stationsetup.php
+++ b/application/controllers/Stationsetup.php
@@ -572,8 +572,8 @@ class Stationsetup extends CI_Controller {
 			// Data for station_profile
 			$dbdata = [
 				'station_active'        => 0,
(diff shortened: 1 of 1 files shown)


phl0 (changes requested):
The code obviously only works if at least one station exists. Would make sense to also allow at least importing if no location exists at all?

AndreasK79:
> The code obviously only works if at least one station exists. Would make sense to also allow at least importing if no location exists at all?

You are absolutely right. Removed the check on empty array.

Issue #2485, opened 2025-11-04
Strange Behaviour at QSY/Split since new Cluster


int2001:
I am absolute unsure, if there's a problem with the new CAT-Logic, @szporwolik implemented.

Have a look at the Video (not my setup) at second 15ff.

Whats happening here:
1.) OP was in Split Mode on his Yaesu (RX 17m // TX 15m)
2.) OP clicked on a 17m Spot
3.) Radio QSYed to 17m - so far everything okay
4.) at 00:15 he removes split, and the whole radio jumped to 15m (doesn't happen if FLRig/WLGate is disabled)

expected behaviour: Stay at least for RX on 17m

since i own no yaesu, i wasn't able to repro it. My icom behaves different here.

any idea @szporwolik ?

https://github.com/user-attachments/assets/32c317cc-f38e-473c-875f-53ed96f9d525

int2001:
There's more strange.
The function to QSY from the QSO-Form is new.

Old behaviour: Change Band at QSO-Form, no QSY happens - one had to do that manual
New: Change Band at QSO-Form, TRX QSYs to band.

in general a good idea. but there are some race-conditions. Still unsure if same "bug".

When i qsy from 40m to - e.g. - 20m (on the radio), i often switch to FM to tune my antenna.
Since last update the first press on "FM" which seems to be instantly overwritten by CAT, so it stays at USB. So i've to press "FM" again.

i guess the bug is somewhere here: https://github.com/wavelog/wavelog/blob/3c087176a274797af8e0d108afc5eab9f430dbf0/assets/js/cat.js#L522

tried to dig a bit in. complex stuff.
It looks like @szporwolik accidently built some kind of "loop" regarding my FM-Issue.
Analyzed a bit:

- OP QSYs on radio
- UI follows and sets QRG/Band
- QSY-onChange at UI is triggered and emits a NEW QSY (with the previous band, since it wasn't updated / i pressed "FM" to fast)
- Results to correct Frequency but wrong mode (and a redundant/not nescessary CAT-Request)

szporwolik:
There are two items, that we need to separate:
- I confirm split behavior reported by the user at Yeasu - I have seen the same on Sunday while working on https://github.com/wavelog/wavelog/pull/2479 ; the current CAT command sets both VFO A and VFO B at the same freq; I will come to that later.
- band change need further discussion - see below

[...]

PR #2487, opened 2025-11-04
Don't clear on Bandchange!

int2001:
People are complaining that complete form is cleared on Bandchange.
This never happened before, but since #2429 

This one reverts that logic. Please have a look if something at the new Cluster fails @szporwolik

### DIFF
--- a/assets/js/sections/qso.js
+++ b/assets/js/sections/qso.js
@@ -2238,9 +2238,6 @@ $('#band').on('change', function () {
 		return;
 	}
 
-	// Clear the QSO form when band is manually changed
-	$('#btn_reset').click();
-
 	// Set flag to prevent waterfall from auto-reverting the band change
 	if (typeof dxWaterfall !== 'undefined') {
 		dxWaterfall.userChangedBand = true;


szporwolik:
That's tricky - then we will again run into a troubles discussed in  https://github.com/wavelog/wavelog/pull/2476

This is issue with the architecture, not only code. 

Please see https://github.com/wavelog/wavelog/issues/2485#issuecomment-3486626071

int2001:
we need a solution b4 our next release here. current situation is not good / ready 4 release

HB9HIL:
I double that. This has to be resolved. The DX Waterfall has way too much impact on the form data which is the most important thing. Clearing users formdata must not happen except in really special and carefully considered cases. 

There is no way around

szporwolik:
I would say, let's cancel this PR - more complex update to come within: https://github.com/wavelog/wavelog/pull/2491

int2001:
i'll leave this PR open for now, since it mitigates things "for now" (not perfect, but it works for the moment).

PR #2491, opened 2025-11-05
[DX Waterfall] QSY behaviour

szporwolik:
## Summary
Refactored DX Waterfall from bidirectional coupling (form ↔ radio ↔ waterfall) to unidirectional data flow (Radio → CAT → catState → Waterfall/Form), eliminating race conditions and simplifying the codebase.

This shall address:
- https://github.com/wavelog/wavelog/issues/2485
- https://github.com/wavelog/wavelog/pull/2487

## Problem Statement
The original implementation had bidirectional coupling between form, radio, and waterfall components, causing:
- Race conditions when form and CAT tried to update simultaneously
- Band selector changes triggering unwanted radio frequency updates
- Form changes affecting waterfall display during CAT updates

## Solution
Implemented unidirectional data flow with radio as the single source of truth:

```
Radio → CAT updates → window.catState → Waterfall reads catState
↓
Form reads catState (display only)
```


phl0:
You may want to change the base branch to dev here?

int2001:
just checked this from user-perspective in a quick-run. things found so far:
- ❌ The "Recheck if wked/cnfmd" doesn't trigger on Bandchange. only on mode-change (QSO-Form / no matter if CAT enabled or not)
- ~~❌ Manual QSY to other Band (on TRX) still triggers back to radio (At Radio doing: 40m LSB -> 20m -> FM // instant overwrite of FM to USB at radio via CAT)~~ // my fault: was running two different instances in 2 browser-windows. this one seems to be solved.
- ✅ Manual QSY to other Band (on TRX) doesn't triggers back to radio, Frequency/Mode is written to form
- ✅ Changing Band at Form doesn't trigger CAT (good!)
- ✅ Changing Band (form or cat) doesn't empty qso-form (good!)

szporwolik:
Mentioned issue was fixed, but... the code was massively refactored - some of the code was moved to ```radiohelpers.js``` , some to ```common.js``` for reuse and the whole logic was changed to state machine. 2nd change will significantly improve or even solve most of the possible race conditions. It's more predictable than initial PR and - most important - it shall be much easier to maintain and understood.

Downside - again, massive piece of code for the review. 😞 

## Important Note
As the code was refactored in all possible aspects, please consider it UNTESTED - whatever was tested/was working in the past, may not be working now and new bugs might be introduced. I tried my best to test it, but it's massive piece of code.
In addition, we have 2 major paths - no CAT and with CAT, both are different, so I would recommend to test both.

PR #2493, opened 2025-11-06
Add QSO count to station location list

phl0:


int2001 (approved):
works.
checked:
- Locations with QSOs
- Locations without QSOs

AndreasK79 (approved):

PR #2494, opened 2025-11-06
[Advanced Logbook] Added continent fix and distance fix

AndreasK79:
Added 2 new menu items under action.
- Fix continent - set continent based on DXCC where continent is not set, or is wrong.
- Update distances - update distances where distance is not set

Added many translations which were hardcoded in the JS.

Todo:

- [x] Add backend and handling for continent
- [x] Add dialog for distances
- [x] Add backend for distances
- [x] Refactor the flow with the dialog and the updates
- [x] Tweak distance update


phl0 (approved):
Works. Removed cont and distance information for some QSOs and they were updated as requested.

PR #2495, opened 2025-11-06
Refactor translations

phl0:
<del>Refactored translations where text included stray whitespaces (at beginning or end) to prevent translators from crashing the text format.</del>

This is not applicable as weblate automatically cares for leading and trailing whitespace. So reverted that commit. Remaining changes:

- Some white space replacements in application/views/interface_assets/footer.php
- Removed unereferenced translation string from application/views/qso/index.php
- Refactored two strings with sprintf in application/controllers/Stationsetup.php and application/views/stationsetup/stationsetup.php

### DIFF
--- a/application/views/awards/wae/index.php
+++ b/application/views/awards/wae/index.php
@@ -6,7 +6,7 @@
             var lang_awards_info_button = "<?= __("Award Info"); ?>";
             var lang_award_info_ln1 = "<?= __("WAE Award"); ?>";
             var lang_award_info_ln2 = "<?= __("The oldest and most renowned of all DARC certificates is awarded for contacts with amateur radio stations in European countries and on islands listed in the WAE country list on different bands."); ?>";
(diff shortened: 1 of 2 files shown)


phl0:
Pls carefully spend a second set of eyes on https://github.com/wavelog/wavelog/pull/2495/commits/c5aca6fc463770354ee5745e9d40b1222b1e361d. I did not find any place where this text is used but might have missed something.

HB9HIL (approved):
Ok for me like this. The commit c5aca6fc463770354ee5745e9d40b1222b1e361d was from the old DXCC summ in the qso view. I gues sit's okay to remove it. I couldn't find any reference either. 👍

phl0:
The stray translation string re the DXCC sum was forgotten to be removed with https://github.com/wavelog/wavelog/commit/a313d261a700f44d7b67f15bc34f83099be0a623#diff-9410f73c9b8edff1bd4b0dc1c972ea4e8a538f3a3be3712fb5b8a79dced9124b when the code referencing this was removed.

PR #2498, opened 2025-11-07
Enable ColReorder

DB8DT:
This enables the feature ColReorder in Datatables used in Section logbookadvanced


AndreasK79:
Hi, while the thought is good, we can't accept this PR.

At least these things will break in the advanced logbook:
- row reloads will be wrong (a functionality which is done several times when editing/updating)
- some quicksearch functionality will break

We would also need to have a way to restore column order for users.

PR #2507, opened 2025-11-10
show more specified permission paths in the installer

HB9HIL:
we can show the permission status of `application/config`, `application/cache` and `application/logs` instead just `application/` in the installer.


HB9HIL:
Before 
grafik


After
grafik

PR #2509, opened 2025-11-10
[DX Cluster] Change of the CAT marker/table sorting added information about timezone

szporwolik:
Changed behavior based on user feedback — the table sorting will now be locked only when the special mode is activated.
image

Changed time of last fetch to be UTC with "Z" marker:
image


phl0 (approved):
This works for me (non-CAT). May a CAT-enabled one also have a check here pls :)

AndreasK79 (approved):

PR #2510, opened 2025-11-11
Fixes QRG and RST/CAT handling

HB9HIL:
Fixes two issues:

1. Commit: 7149f3403f6ed8368cdeb5c44cfaf47ea8bfd1d1
The qrg input was broken when clicking on a frequency in the dxwaterfall. the input frequency was in the wrong unit and the users selected unit was not respected. this fixes this

2. Commit: c344ff1994d895410f40676bbbb74d39da55267e
When a new frequency arrived by cat (no matter by polling or live) the user RST was overwritten every time the cat2ui() functionw as triggered. It's not necessary to set RST here as this already is handled by mode handling 8rst gets overwritten only when mode changes, there is a listener in place).

please test and report back any special cases (especially @szporwolik) . but this makes it usable for the majority of people

### DIFF
--- a/assets/js/dxwaterfall.js
+++ b/assets/js/dxwaterfall.js
@@ -6248,9 +6248,10 @@ function setFrequency(frequencyInKHz, fromWaterfall) {
     }
 
     // Update frequency field with value in Hz
-    $('#frequency').val(frequencyInKHz * 1000).trigger('change');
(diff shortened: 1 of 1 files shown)


int2001 (approved):
confirm. fixes the bugs

PR #2511, opened 2025-11-11
Use global options in dxwaterfall

HB9HIL:
Each instance has default values in global options defined for the dxcluster. These values should also be used in the dxwaterfall.

### DIFF
--- a/assets/js/dxwaterfall.js
+++ b/assets/js/dxwaterfall.js
@@ -3297,9 +3297,9 @@ var dxWaterfall = {
         var age = 60; // minutes
         var de = this.currentContinent; // Use current continent (may have been cycled)
 
-        // Check if dxwaterfall_maxage is defined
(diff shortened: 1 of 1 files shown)


szporwolik:
I think this is partially redundant:
https://github.com/wavelog/wavelog/blob/6166d2407c3af93bfe8840bf7905655760f54861/application/views/components/dxwaterfall.php#L59-L60

But this PR is justified - it looks that we have lost the logic behind during recent re-factoring.

HB9HIL:
> I think this is partially redundant:
> 
> https://github.com/wavelog/wavelog/blob/6166d2407c3af93bfe8840bf7905655760f54861/application/views/components/dxwaterfall.php#L59-L60
> 
> But this PR is justified - it looks that we have lost the logic behind during recent re-factoring.

oh yes. looks like a redundancy. but tbh I would rather load the values in the controller then in the view component. This would be a better fit to the rest of wavelogs strcuture. I gonna fix this

PR #2517, opened 2025-11-12
[DX Waterfall] Mode classification

szporwolik:
Moving mode classification to backend - single source of truth.
Fixing the issue when DX Waterfall was showing correct sideband, but tunned radio to other frequency.

**Base idea:**
Backend will always return ```spot.submode``` and ```spot.mode``` - this can be used in all UI elements without the need of recalculation at the frontend side.


AndreasK79:
As far as I can tell, this is working. Mode ID still works, and info is sent to QSO Form.

int2001 (approved):
tested a lot of constellations. looks good so far

PR #2518, opened 2025-11-12
Only validate grid if existent

phl0:
PR https://github.com/wavelog/wavelog/pull/2504 introduced a bug. Searching for a prefix/part of call there are some errors because grid cannot be validated. We need to check if gridsquare array key exists.

Repro:
- Enable debug mode
- Search for prefix / part of call

Result:

image


AndreasK79:
Fixes error :)

PR #2521, opened 2025-11-12
[DX Cluster] Translations

szporwolik:
Changes:
- Fixed missing translations
- Compacted some strings, so layout is consistent when using narrow themes and more descriptive languages

### DIFF
--- a/application/views/bandmap/list.php
+++ b/application/views/bandmap/list.php
@@ -64,6 +64,7 @@
 	var lang_bandmap_new_continent = "<?= __("New Continent"); ?>";
 	var lang_bandmap_new_country = "<?= __("New Country"); ?>";
 	var lang_bandmap_worked_before = "<?= __("Worked Before"); ?>";
+	var lang_bandmap_worked_details = "<?= __("Worked on %s with %s"); ?>";
 
 	// Bandmap filter prefixes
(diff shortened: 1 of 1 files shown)


application/views/bandmap/list.php
```diff
@@ -78,6 +83,15 @@
 	var lang_bandmap_see_details = "<?= __("See details for"); ?>";
 	var lang_bandmap_worked_on = "<?= __("Worked on"); ?>";
 	var lang_bandmap_not_worked_band = "<?= __("Not worked on this band"); ?>";
+	var lang_bandmap_lotw_last_upload = "<?= __("LoTW User. Last upload was"); ?>";
+	var lang_bandmap_days_ago = "<?= __("days ago"); ?>";
```
HB9HIL: This is not really good practice as it rips strings apart and makes it really hard for users to translate (missing context). Especially in some languages where grammar depends on context this is really not nice. 

I know you had it working with %s somewhere and I would like to see that here (and some other places aswell.

szporwolik: Right, sorry - the javascript ```translation_string.replace('%d', value_days)``` shall do the trick and will be in fact much easier for translators.

AndreasK79 (approved):

PR #2523, opened 2025-11-12
[QSO Page] Fix for non-positive frequency value when band is not selected

szporwolik:
Problem statement - there is wrong logic implemented:
- open QSO page
- ensure CAT is OFF
- enter out of band frequency like 225kHz - band dropdown will be null
- enter 0 or negative frequency - this will be allowed

Why this is important:
- DX Waterfall expects the frequency to be non-negative (more than 0)

Solution:
- the current ajax query at line 82 in qrg_hanlder.js is incorrect, null value is being send in given test case
- we want to run the mentioned query only when band is selected or default to some non-negative number

### DIFF
--- a/assets/js/sections/qrg_handler.js
+++ b/assets/js/sections/qrg_handler.js
@@ -76,7 +76,8 @@ async function set_new_qrg() {
 
 	// If field is empty or parsing failed, fetch default frequency for current band/mode
 	if (!new_qrg || new_qrg === '' || isNaN(parsed_qrg) || !isFinite(parsed_qrg) || parsed_qrg <= 0) {
-		if (typeof base_url !== 'undefined') {
+		// Check if band is selected before attempting to fetch frequency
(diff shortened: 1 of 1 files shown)


phl0 (approved):
Tested a few situations around. Imho prevents freq from being 0 or negative. So works on my end.

assets/js/sections/qrg_handler.js
```diff
@@ -89,12 +90,22 @@ async function set_new_qrg() {
 				return;
 			}
 		}
+		// If band is empty or base_url is undefined, set frequency to 1 Hz
+		$('#frequency').val(1);
+		$('#qrg_unit').html('kHz');
```
HB9HIL, line 95: we should not change user set qrg unit manually (overwrite)

HB9HIL (changes requested):

szporwolik:
Replaced by https://github.com/wavelog/wavelog/pull/2537

PR #2527, opened 2025-11-13
use correct syntax for set_header()

HB9HIL:
seems like this was forgotten in commit e18abd3962a41a69b0aeed146192948261f8e070


HB9HIL:
Ref to https://github.com/wavelog/wavelog/pull/2357

PR #2528, opened 2025-11-14
Prevent user_callsign to have Ø in DB

phl0:
User can set/edit his user callsign in profile to contain 'Ø'. This does obviously not have any impact on daily logging operations. But Adding a user to a club station's authorized users fails because user_callsign contains a crappy zero in DB and user does not search for slashed zero but normal zero. Thus user account can simply not be found.

### DIFF
--- a/application/controllers/Api.php
+++ b/application/controllers/Api.php
@@ -263,6 +263,19 @@ class API extends CI_Controller {
 					if(count($record) == 0) {
 						break;
 					}
+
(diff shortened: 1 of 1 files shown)


int2001 (approved):
Read code - from that perspective everythin okay.
Tested as well.

ApprØve

phl0:
> ApprØve

Thank y︀øu

PR #2537, opened 2025-11-15
fixing various bug which were introduced during dxwaterfall implementation

HB9HIL:
see commit messages

PR is based on #2523 and replaces this one

@szporwolik You have to change your dxwaterfall stuff regarding the call of `frequencyToBand()` function and the valuecheck of `null` instead of `All`


HB9HIL:
worked out some bugs with @szporwolik

int2001 (approved):
works.
bugs solved / tested:
- Change Band at Post-QSO --> QRG will change as well
- CAT per default disabled at Post-QSO
- Code much better, because it uses existing functions

PR #2545, opened 2025-11-16
[Logbook Advanced] Error in translation string

szporwolik:
Found during other development work. 

Fixed language variable that was causing JavaScript error preventing Advanced Logbook from loading correctly.


HB9HIL (approved):
good find

PR #2550, opened 2025-11-17
fix the order of clearing and populating data

HB9HIL:
We need to first clear the fields before populating new data. 

also moved the pendingReferences = null to the top of the resetFields function as I noticed on slow javascript the clearing of the selectize fields can cause minimal race conditions in this case. Another way would be to make reset_fields() async which would need a bit more work and is kinda cumbersome. So I think this light fix should work for the majority of setups (testing with a real slow machine here, should be fine)


szporwolik (approved):
Checked seems to be solving the old issue

PR #2551, opened 2025-11-17
[DX Cluster] Final fine-tuning and bugfixing

szporwolik:
This shall initially go with https://github.com/wavelog/wavelog/pull/2541 

- Feature: changes of the behavior - purple mode as per feedback @int2001 
- Feature: of the layout (new columns) as per feedback @int2001 
- Feature: code pulling the spots only for active band in purple mode (also proposed by @int2001)
- Feature: introduced 2nd cache config setting - we want to use band cache, even if we don't want worked cache
- Feature: spot with no entity will display pirates' flag
- Bugfix: fix for backed not properly utilizing Cluster API data
- Bugfix: fixed map not loading correctly
- Bugfix: some warnings suppression observed by @int2001 
- Bugfix: removed debug code, improved js loading (thanks! @HB9HIL)
- Bugfix: fixed slow table rendering (multiple renders)

### DIFF
--- a/assets/js/sections/bandmap_list.js
+++ b/assets/js/sections/bandmap_list.js
@@ -90,8 +90,30 @@ $(function() {
 	let applyFiltersTimer = null;
 	function debouncedApplyFilters(delay = 150) {
 		if (applyFiltersTimer) clearTimeout(applyFiltersTimer);
-		applyFiltersTimer = setTimeout(() => applyFilters(false), delay);
+		applyFiltersTimer = setTimeout(() => {
+			// Safety check: only call if applyFilters is defined
+			if (typeof applyFilters === 'function') {
+				applyFilters(false);
+			}
+		}, delay);
 	}
+
+	// ========================================
+	// MAP VARIABLES (declared early for drawCallback access)
+	// ========================================
+
+	let dxMap = null;
+	let dxMapVisible = false;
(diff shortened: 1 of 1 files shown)


szporwolik:
Also recommended to pull this one for DX Cluster API [2.0 Alpha ](https://github.com/szporwolik/DXClusterAPI/commit/e7528752fa84ae7b02d44132ce56c5d2243b2570)

application/models/Logbook_model.php
```diff
@@ -2707,262 +2709,554 @@ function get_batch_spot_statuses($spots, $logbooks_locations_array, $band = null
...
+			$worked_bind_params = array_merge($worked_bind_params, $logbooks_locations_array, $continents_array);
+
+			$confirmed_queries[] = "
+				SELECT 'cont' as type, COL_CONT as identifier, COL_BAND as band, COL_MODE as mode
+				FROM {$this->config->item('table_name')} FORCE INDEX (idx_HRD_station_id)
+				WHERE station_id IN ({$station_ids_placeholders})
+				AND COL_CONT IN ({$continents_placeholders})
+				AND ({$qsl_where})
+				GROUP BY COL_CONT, COL_BAND, COL_MODE
+			";
```
HB9HIL: Not 100% sure but you may should use bindings here.

PR #2553, opened 2025-11-19
[X-mas cleaning] Found some leftovers

AndreasK79:
While browsing the code, I found some code and files that we do not use.


HB9HIL (approved):
nice! checked and couldn't find any references

PR #2554, opened 2025-11-19
Make OnAir-Widget dynamic

int2001:
QRZ is very very picky when it comes to JS / interaction.
This PR is compatbile with the old widget, but adds two functionalities:

1. Dynamic refresh if JS is allowed (every 60s - should be enough for a page)
2. Dynamic refresh if JS is disallowed via special qrz-tag.
3. If called like before nothing is changed (compatibility)

Testing at QRZ:
```
<iframe name="iframe" src="[YOUR WAVELOG URL]/widgets/on_air/[PUBLIC SLUG]?nojs=1" height="140" width="640" frameborder="0" align="top" data-refresh="60">
```

going to update wiki after merge here

### DIFF
--- a/application/controllers/Widgets.php
+++ b/application/controllers/Widgets.php
@@ -119,7 +119,15 @@ class Widgets extends CI_Controller {
 	 * @return void
 	 */
 	public function on_air($user_slug = "") {
-		// determine theme
+		// Check for nojs parameter (for QRZ.com compatibility)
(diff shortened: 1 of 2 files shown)


phl0:
Fixed this one by https://github.com/wavelog/wavelog/pull/2554/commits/bf51e45167e28421bba0bb0c772bec61043b1e0e:

image

phl0 (approved):

PR #1548, opened 2025-01-24
QSO Edit Modal fix

HB9HIL:
Same issue as in #1237

Bad HTML ID's cause issue in the QSL Tab of the QSO edit modal.

image

Reproduce:

- Open DXCC Award
- Choose any QSO from the table
- Open QSO Edit Modal
- See failing tab control in QSL tabs

This PR fixes this issue by renaming the tab control ID's

thx to DL2COM for reporting this issue

PR #2561, opened 2025-11-20
Use ID mode instead of class

int2001:
when editing a previous QSO from QSO-Page (History),
one wasn't able to change the Mode, because CAT always overwrites it.

this patch targets ONLY the element `id="mode"` instead of every field which has the class `mode`.

pse check carefully for side-effects.

things to test:
Play with / without CAT and look carefully for MODE and default-RST

### DIFF
--- a/assets/js/cat.js
+++ b/assets/js/cat.js
@@ -725,7 +725,7 @@ $(document).ready(function() {
         var $band = $('#band');
         var $frequencyRx = $('#frequency_rx');
         var $bandRx = $('#band_rx');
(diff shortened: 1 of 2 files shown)


HB9HIL:
Check https://github.com/wavelog/wavelog/pull/1548 and https://github.com/wavelog/wavelog/pull/1237 for reference and fixes of the same issue at other fields

int2001:
slightly different bug than only `edit_` vs. non-`edit_`.
but tnx for hint

PR #2562, opened 2025-11-20
Check LoTW certificates against CRL

phl0:
In the current state we have no option to check whether a LoTW certificate has been superseded before its expiry. We found a CRL status API at ARRL that lets us query a certificate for its validity. The current implementation checks all certs when calling the LoTW cert page as the API only allows one query per call.


phl0:
The current implementation checks each cert against the CRL on page load. Having 12 active LoTW certificates this is how the timing roughly increases.

Current dev (no CRL):
Screenshot from 2025-11-21 07-29-43

This branch (with CRL check):
Screenshot from 2025-11-21 07-34-17

If this proofs to be unfeasible in the wild the code might be changed to a button to (manually) check each cert against the CRL.

PR #2563, opened 2025-11-20
[QSO Edit] Fix for CAT updating

szporwolik:
Problem:
When CAT is active QSO edit form is updated with CAT data, this is a bug.

Solution:
Per my understanding this shall be enough to fix the edit form issues (see code).

Dev note:
There are still some #ids same as QSO form, but I would focus on mode with this one.

### DIFF
--- a/assets/js/cat.js
+++ b/assets/js/cat.js
@@ -756,6 +756,10 @@ $(document).ready(function() {
                     // Auto-update band based on frequency
                     if ($band.val() != newBand) {
                         $band.val(newBand).trigger('change'); // Trigger band change
+                        // Update callsign status when band changes via CAT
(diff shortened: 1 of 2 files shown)


int2001:
For documentation: Default-RST isn't set correctly

int2001:
Edit-Bug: Gone.
Modechange-Bug with RST: Gone

but:

no recheck on ModeChange which was triggered by cat (checked latest master = works there)

int2001:
tested w. CAT enabled:
- QSY - works
- ModeChange - works
- edit - still working
- changing RST to 5/8 and turing VFO - works

tested wo CAT enabled:
- QSY - works
- ModeChange - works
- edit - still working
- changing RST to 5/8 and changing Freq - works

PR #2564, opened 2025-11-20
[DX Cluster] DX and spotter callsign click behaviour

szporwolik:
Problem:
Inconsistency, as in the rest of Wavelog we expect that clicking on the callsign would open contact history per callsign.

Solution:
Used existing function/js lib to display callsign history. 

Additionaly - cluster's font was changed to cover slashed zeros by font.


int2001:
looks good so far. will test a bit more

PR #2574, opened 2025-11-23
set qrg from calculated

HB9HIL:


int2001 (approved):
tested:
- switching forth and back (None / CAT)
- clicking at waterfall
- reloading page with and without CAT default
- Changing Bands manual
all working

PR #2577, opened 2025-11-23
Refactor QSL colors for arrows

phl0:
The (unusal) colors for eQSL need to be refactored as they only use red and green atm.


phl0:
Other QSL methods need to be reviewed as well ....

phl0:
This is how it looked before:

image

image

We adjusted all colors so that the statuses are shown correctly. Also we refactored the QSL statuses to be edited for QRZ and CLUBLOG as those are different according to ADIF (3.1.6). See https://adif.org.uk/316/ADIF_316.htm#QSODownloadStatus_Enumeration which is different for the status of other QSL services.

PR #2583, opened 2025-11-24
Remove accidentially checked-in CSS

phl0:
Seems we accidentially checked in some CSS that was meant for debugging/development purposes and is never used. Probably gone unnoticed with monster squash merge in https://github.com/wavelog/wavelog/commit/c70c2ec5cd829fe86c8a81ed36979ded7a9dc204.


int2001 (approved):

phl0:
A bit more of digging. Introduced with eed8ef4bf25e6706e1f7f2186a4dbb8b2e73fc15 into club stations PR and with https://github.com/wavelog/wavelog/commit/c70c2ec5cd829fe86c8a81ed36979ded7a9dc204 into dev and master subsequently. As club was a large squash commit and branch is deleted we cannot determine with commit introduced this leftover. Stopping here.

Issue #2548, opened 2025-11-17
Upoaded ADIF from Nov SS SSB (and CW) doesn't bring in full contest exchange


Reid-n0rc:
**Describe the bug**
Wavelog does not save all the contest exchange information for the ARRL November Sweeptakes SSB (and CW). The ADIF output from N1MM contains the complete contest exchange that is sent and received. This data should be retained when importing into Wavelog. 

**Version Information**
We need some information about your versions. You can get all Information in the Admin > Debug Menu
- OS:               Ubuntu
- PHP:              8.3.6
- MySQL/MariaDB:     	10.11.13-MariaDB-0ubuntu0.24.04.1
- Wavelog:          2.1.2

**Steps to Reproduce**
1. Export ADIF file from N1MM after a contact is logged in the ARRL November Sweep Stakes SSB
2. Import ADIF to wavelog
3. Open imported contact.
4. Click on Edit QSO
5. Go to the Contest tab.
6. See that only the RS and Serial number has been imported. The Precedence, Check, and Section have been ignored.
 
**Expected Behavior**
When importing the ADIF file for the Nov SS contest, all contest data should be retained. The Precedence, Check, and Section should be in the Exchange field.

**Screenshots**

Sample ADIF data:
[...]

PR #2584, opened 2025-11-24
Refine CBR import and make it more obvious it exists

DB4SCW:
This PR adds 4 things:
- If you have a multi-op station, some logs like N1MM add the TRX number as the last element of the CBR file. Added a new switch to indicate if this number is present to prevent it from being imported as part of the received exchange
- Explicitely skip QTC records. This worked previously because the code was strict enough to not include them in the qsos and fail, but this will make that more obvious in the code and save a bit of memory during parsing.
- ~~The CBR import feature seems to be overlooked quite often (last occurence in  #2548 ). So I changed the badge of the info text after importing a contest ADIF from "information" to "warning" to make it more obvious.~~
- Support contest CBRs without signal rapport exchange (like ARRL Sweepstakes)
- Support exchanges where the received exchange can be different length (like TRC-DIGI)

### DIFF
--- a/application/libraries/Cbr_parser.php
+++ b/application/libraries/Cbr_parser.php
@@ -241,12 +241,15 @@ class CBR_Parser
                 case '144':
                     $band = '2m';
                     break;
+                case '220':
                 case '222':
                     $band = '1.25m';
                     break;
+                case '430':
                 case '432':
                     $band = '70cm';
                     break;
+                case '900':
                 case '902':
                     $band = '33cm';
                     break;


int2001:
Hi Stefan,
not a friend of making it more visible. Many people already asked: "Whats this? Do i need it?".
People who use contesting are aware of the hint. no need for "danger" instead of "info"

DB4SCW:
So... remove the warning but keep the rest?

int2001:
think so. need to validate the rest

DB4SCW:
Added more and better error handling and support for CBRs without signal rapport exchanges (see #2548 )

int2001 (approved):
tested with:
- classic contest (works)
- ARRL-SS-SSB contest (works)

good improvement

DB4SCW:
Added graceful handling for contests, where the received exchange can be different length depending on the qso partner (like TRC-DIGI)

int2001 (approved):

DB4SCW:
merged @phl0 's PR, good find

PR #2586, opened 2025-11-24
[DX Cluster] Post 2.2 release fixes

szporwolik:
Problem:
When CAT is on the filters follows band at the TRX (via CAT polling or websockets), this shall not happen untill user entered the "purple mode".

Solution:
Changed the logic a little to match the expected behaviour.

Dev note:
Tested a lot, but please test carefully - there is a lot of logic behind after purple mode was seperated from normal CAT operation.

Errata:
- fixed also not correct CAT status box (thanks SP9WSS!)
- fixed not proper content of the columns when working on narrow screen (thanks OM0AAO!)

### DIFF
--- a/assets/js/cat.js
+++ b/assets/js/cat.js
@@ -745,6 +745,12 @@ $(document).ready(function() {
             }
         }
 
+        // Validate that we have frequency data before proceeding
+        if (!data.frequency || data.frequency == 0 || data.frequency == null) {
+            console.warn('CAT: No valid frequency data received');
+            return;
+        }
+
         // Force update by clearing catValue (prevents cat2UI from blocking updates)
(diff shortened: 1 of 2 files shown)


szporwolik:
Please wait with this one, I think there is another issue (thanks SP9WSS!)

szporwolik:
Please: test, test, test.

Scenarious:
- General CAT behaviour, on/off/purple mode, changing of the radio, working offline, stale data etc.
- Monitor: filters, purple markers, but also CAT div behaviour
- Resize the screen, play with filters

int2001 (commented):
- Non-EPIC-CAT-Mode: Filter is still there after clicking a spot: great
- Tested with None, Polling and Socket

PR #2591, opened 2025-11-24
[DX Waterfall] Fixes for not proper closure of the component

szporwolik:
Problems:
- when user turned on DX Waterfall and then turned it off, changing band/mode was still triggering fetch of spots
- rapid turning the component on and off could lead to unpredicted behaviour

Solutions:
- the whole DX Waterfall object is now properly destroyed
- added short cool off timer preventing user for rapid turning the component on/off

Dev notes:
Tested, but please test carefull this was some serious ingeration in Init/Destroy of the DX Waterfall.

### DIFF
--- a/assets/js/dxwaterfall.js
+++ b/assets/js/dxwaterfall.js
@@ -1,7 +1,7 @@
 // @ts-nocheck
 /**
  * @fileoverview DX WATERFALL for WaveLog
- * @version 0.9.5 // also change line 32
+ * @version 0.9.6 // also change line 32
  * @author Wavelog Team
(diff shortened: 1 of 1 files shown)


int2001 (approved):
Confirm.

Wild requests are gone.

Repro in dev:
- Turn Waterfall on, observe Requests in Network-Tab
- Turn Waterfall off, observe the requests. Turn VFO.
- After a time there are wild requests.

This Patch fixes it

phl0 (approved):
No issues found on my (small) instance

Issue #2588, opened 2025-11-24
New DX Cluster ignores Bands and Modes settings


xyz667:
### I have checked the following

- [x] I am running the latest version of Wavelog
- [x] I have cleared my browser cache
- [x] This is a bug, not a setup question or feature request

### Bug Description

Fist of all, thank you very much for the cool new DX cluster! I really like it a lot!

But one thing annoys me a little:
In the "Bands" and "Modes" menus of Wavelog I have deactivated quite a few bands (which I'm not licensed to use) and the CW mode (which I'm too dumb for). DX Cluster ignores those settings, however.
As the cluster always auto-selects all bands and modes (and does not store the last filter setting), I have to filter out the unwanted bands and mode(s) every time I call the cluster.

Ok, I guess I can live with it 😅  but I'd still call this behavior inconsistent.

### Steps to Reproduce

1. Deavtivate stuff in "Bands" and "Modes"
2. Open DX Cluster

### Expected Behavior

DX Cluster does not show deactivated bands and modes, or at least does not auto-select them.

### Wavelog Version

2.2

### PHP Version

8.2.29

### MySQL/MariaDB Version

10.11.14-MariaDB-0+deb12u2

### Server OS

DietPi v9.17.2

### Browser

Safari

### Device Type

Desktop

lycanananas:
DXCluster do not work on Firefox Ubuntu works on Edge only :/ 

I got:

```
Uncaught ReferenceError: getBandsInGroup is not defined
```

gx1400:
@lycanananas looks an awful lot like my bug report from yesterday that looks like it may have been a cached js file that didn't clear

https://github.com/wavelog/wavelog/issues/2579

xyz667:
@int2001 I don’t think this should be closed. 
The user-defined band and mode settings are still ignored throughout the new cluster in 2.2.1.
Yes, there is a new filter favorites functionality, but it does not take the user settings into account. 
At the very least preset the filter with the user’s band (and mode?) settings?

int2001:
handing over to @szporwolik

brstiefel1:
I'm having a similar but different issue with the new DX cluster. I recently upgraded from version 2.1.2 to 2.2.1

On Chrome it won't allow me to enable CAT control or the lock to follow the radio's tuned band. It does work on Edge, but it won't stay selected. Every time I return to the DX cluster page I have to re-enable the CAT control.

[...]

PR #2607, opened 2025-11-27
[DX Cluster] UI improvements post 2.2 release

szporwolik:
Changes:
- added the favorites following QSO page fav logic - this shall address some of the initial users' requests
image

- reworked CAT button - this shall be more intuitive 

image
image

- added a few icons - this shall help to compact layout when working on narrow screens

image

- added My Submodes filter - this is the way for users to filter out specific sub-modes like FT4, PSK etc. while keeping some in like FT8 for Digi mode - this shall address some of the initial users' requests

image

- improved responsiveness, focused on typical half screen operation (around 1000px)

image

- introduced hide menu button, for users wanting to use max of the screen

image

- seperated 6m from VHF bands
image

image

### DIFF
--- a/assets/js/sections/bandmap_list.js
+++ b/assets/js/sections/bandmap_list.js
@@ -32,7 +32,8 @@ const BAND_BUTTONS = [
 	{ id: '#toggle17mFilter', band: '17m' },
 	{ id: '#toggle15mFilter', band: '15m' },
 	{ id: '#toggle12mFilter', band: '12m' },
-	{ id: '#toggle10mFilter', band: '10m' }
+	{ id: '#toggle10mFilter', band: '10m' },
+	{ id: '#toggle6mFilter', band: '6m' }
 ];
 
 const BAND_GROUP_BUTTONS = [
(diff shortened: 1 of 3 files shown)


assets/js/sections/bandmap_list.js
```diff
@@ -1960,11 +2026,163 @@ $(function() {
...
+		if (keys.length === 0) {
+			$menu.append('<span class="dropdown-item-text text-muted"><em>' + lang_bandmap_no_filter_presets + '</em></span>');
+			return;
+		}
+
+		keys.forEach(function(key) {
+			// Build the menu item with data attribute on the parent div for easier click handling
+			let $item = $('<div class="dropdown-item d-flex justify-content-between align-items-center dxcluster_fav_item" style="cursor: pointer;"></div>').attr('data-fav-name', key);
+			let $nameSpan = $('<span></span>').text(key);
+			let $deleteBtn = $('<span class="badge bg-danger dxcluster_fav_del ms-2" title="Delete"><i class="fas fa-trash-alt"></i></span>').attr('data-fav-name', key);
```
HB9HIL: Untranslated string "Delete". You can use `lang_general_word_delete` which is already available through https://github.com/wavelog/wavelog/blob/ea8d895e8f1184bd0ce80e4b81ddb0ed8e962845/application/views/interface_assets/footer.php#L65

HB9HIL (changes requested):
Looks good so far. However the initial page load and the behaviour if the Submode button is not good yet. We discuss that internally.

xyz667:
Hi @szporwolik , I can't try the dev branch and PRs easily, but I guess [issue 2588](https://github.com/wavelog/wavelog/issues/2588) has not been incorporated yet?
Wouldn't it make sense to take the user's already pre-set band and mode settings from Wavelog into account for filters?

assets/js/sections/bandmap_list.js
szporwolik: rr

szporwolik:
> Hi @szporwolik , I can't try the dev branch and PRs easily, but I guess [issue 2588](https://github.com/wavelog/wavelog/issues/2588) has not been incorporated yet? Wouldn't it make sense to take the user's already pre-set band and mode settings from Wavelog into account for filters?

There is a plan to have this addressed - not directly, but you shall have this functionality back via Favorites and My submodes working together initially.

PR #2616, opened 2025-11-28
[Advanced Logbook] Update grid if callbook accuracy is greater

AndreasK79:
image

This PR also contains a fix for ADIF export, mention in #2617


int2001 (approved):
Tested:
- Update Distance via Bulkedit
- Update from callbook with checkbox ticked
- Update from callbook without checkbox ticked
- Edit QSO

phl0 (commented):
Does a nice job and works flawlessly :) Sri for late feedback here.

PR #2618, opened 2025-11-30
Club adif

int2001:
Introduce a new Role "Clubmember ADIF" (Clublevel 6) which is allowed to import ADIFS (for their own Call) and Export ADIFS (from their own Call / they made)

- [x] Introduce new role
- [x] Refactor ADIF-Views and Controller for more granular permission-management
- [x] Check Imported ADIF in case of Clubmember ADIF uploaded it. // If Clubmember ADIF uploads the OP-Call is strictly overwritten with the one from the Operator.
- [x] Export only QSOs from Operator
- [-] ~Adjust LBA in similar way.~ // Will be done in extra-PR just for LBA by @AndreasK79 

Testhints:
- Edit Clubstation, set new permission ("Clubmember ADIF")
- Test ADIF Functionality (everything below adif-menu) for normal user and admin.
- Test ADIF Functionality (everything below adif-menu) for ClubOfficer on Clubstation.
- Test ADIF Functionality (everything below adif-menu) for ClubMember on Clubstation.
- Test ADIF Functionality (everything below adif-menu) for ClubMember ADIF on Clubstation

Expected outcome:

- Normal User and Admin: As b4
- ClubMember: as b4 - no export via ADIF-Menu (ADIF Menu-entry not even visible)
- ClubMember ADIF: ADIF Menu visible.
  * Import: All ADIF Entries imported are strictly overwritten with Operators Call. No option to mark ADIF as already sent to 3rd party services, since OP != Owner of the call.
  * Export: Only QSOs the OP made will be exported. no "Mark exported QSOs as exported to LoTW"
 - Club Officer: As b4

### DIFF
--- a/application/controllers/Adif.php
+++ b/application/controllers/Adif.php
@@ -28,7 +28,7 @@ class adif extends CI_Controller {
 
 	private function determine_allowed_tabs() {
 		if (clubaccess_check(6) && !clubaccess_check(9)) {
-			// Only ClubMember Plus and NOT ClubOfficer
+			// Only ClubMember ADIF and NOT ClubOfficer
 			$this->allowed_tabs = ['import', 'export'];
 		} else {
 			// Default: show all tabs (backward compatible)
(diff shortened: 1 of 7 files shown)


HB9HIL (approved):

phl0 (approved):
Works for me. Only issue is that if I want a (new) user with export permissions I have to add him and then edit separately. If that is intended then idc. Otherwise please consider PR https://github.com/int2001/wavelog/pull/24. Wasn't able to add as suggestion because this area of the code is not in scope of this PR (yet).

PR #2620, opened 2025-12-01
[Awards] Extract of hardcoded geojsons

szporwolik:
**Problem:**
With introduction of the states geojsons, keeping the geodata in other js seems redundant. We shall threat those geojsons as single source of truth.

**Changelog:**
- RAC - Extracted
- WAJA - Extracted
- H26 - Extracted
- WAP - Extraced, fixed states geojson, fixed color maps to match new approach, fixed bug with call history not shown properly
- WAS - Extracted, added DC to geojson, but also created states files for Alaska and Hawaii

**Additionaly:**
- extracted Corsica from French geojson

**Testing:**
- Checked awards to properly draw maps 
- Fix state for France - especially Corsica 
- Fix state for USA - especially DC, Hawaii, Alaska (both seperate ADIF entity)  
- Fix state for Netherlands - especially GE, LI, ZE


szporwolik:
Update: per discussion fixed or changed the logic for WAS, a little the DC will be counted as part of Maryland (acc. award rules). Map coloring will follow that.

phl0 (changes requested):
Basically working ok. But found two issues with wrong counting:

- WAS -> https://github.com/szporwolik/wavelog/pull/5
- H26 -> https://github.com/szporwolik/wavelog/pull/6

image

AndreasK79 (approved):

Issue #2619, opened 2025-11-30
Station Location create/edit state/gridsquare fields broken


kc9uhi:
### I have checked the following

- [x] I am running the latest version of Wavelog
- [x] I have cleared my browser cache
- [x] This is a bug, not a setup question or feature request

### Bug Description

-- When creating/editing a station location, the "station state" field is non functional when a DXCC with states is selected (for example, United States of America -K). 'county' is also broken as a result.
-- Station gridsqure->get gridsquare button does not work (javascript error: function not found)

Issues are reproducible using wavelog demo site.

Seems to have been broken with pull request 2553. Restoring these changes restores broken functionality.

### Steps to Reproduce

1) Go to station setup -> create a station location.

2a) Select a Station DXCC with 'states' (such as USA).
2b) Click 'Get Gridsqure'

### Expected Behavior

2a) Station State drop-down populates with states.
2b) Gridsquare field populates

### Wavelog Version

2.2

### PHP Version

8.1.2

### MySQL/MariaDB Version

8.0.44

### Server OS

Ubuntu 22

### Browser

Chrome

### Device Type

None

AndreasK79:
@kc9uhi thank you for letting us know. I've reverted it and merged the code into the dev branch. Fix is there, or wait for next release.

PR #2622, opened 2025-12-01
Revert "[X-mas cleaning] Found some leftovers"

AndreasK79:
Reverts wavelog/wavelog#2553


AndreasK79:
As mentioned in #2619, some of it was in use.

PR #2624, opened 2025-12-01
Fix code to show the active logbook

phl0:
The Polska Awards page shows false information regarding station location. Locations and Logbooks where mixed up here:

image

As the code uses the active station logbook we changed the description and fixed the code to fetch the name of the active logbook.


phl0:
@szporwolik please leave your review here as well.

szporwolik (approved):
Yep, seems that this is in fact a bug. 👍🏻

PR #2627, opened 2025-12-02
Fix Bug with deprecated options and implement #2626

int2001:
This PR contains
- seperation for date and time in Export
- Column-Headers more userfriendly
- additional cols for CSV: Prop_mode, Gridsquare and MY_Gridsquare

- BugFix for a PHP-deprecation-warning (fputcsv required escape and seperator-options)

Implements solution for discussion [2626](https://github.com/wavelog/wavelog/discussions/2626)

No change for via-Field, since it is already empty if not provided


phl0 (approved):
Works. Maybe add PROP_MODE? https://github.com/int2001/wavelog/pull/23

application/controllers/Qslprint.php
```diff
@@ -92,42 +92,54 @@ public function exportcsv()
...
+						"DATE_ON",
+						"TIME_ON",
+						"MODE",
+						"SUBMODE",
+						"FREQ",
+						"BAND",
+						"RST_SENT",
+						"SAT_NAME",
+						"SAT_MODE",
+						"PROPAGATION",
```
phl0:
```suggestion
						"PROP_MODE",
```

Better use the ADIF name here? (though ROUTING is not either)

phl0 (changes requested):

phl0 (approved):

PR #2631, opened 2025-12-02
[UI] Fix for copy to clipboard

szporwolik:
# Problem:

There are three places offering copy-to-clipboard functionality:
- api page
- debug page
- cron page

When working on non-https connection (usually installation / debug / some local instanced) - the copy to clipboard function was not working.

https://github.com/wavelog/wavelog/pull/2524 fixed the code for api page, but still debug was not fixed.

# Solution:

This PR introduces shared function to common.js and updates are places to use the common function. This will eliminate redundant code and make the common approach.


HB9HIL:
That's not working on https is a security feature by the browser. Not sure if we should override that. Just hide the copy fields on non https if you want to improve UX

PR #2633, opened 2025-12-03
[DX Calendar][Contest Calendar] Backend optimization

szporwolik:
### Description

This PR implements a cron-based RSS feed caching system for the DX Calendar and Contest Calendar pages, following the existing pattern for other cron jobs.

### Why?

Cleaner architecture and most imporant - posibility to quickly display part of the data in other parts of the website in the future. As a side benefit - on my test instance DX Calendar load time was reduced from 7s to 1.5s.

### Problem

Previously, both calendar pages fetched RSS data from external sources (ng3k.com and contestcalendar.com) on every page load. This caused:
- Slow page loads when external sources were unresponsive
- Unnecessary load on external servers
- Not perfect user experience during network issues
- For DX Calendar: additional performance issues due to 4 database queries per DXpedition entry

#### Backend RSS Feed Caching
- **New cron job** (`update_rss_feeds`) that runs every 12 hours to fetch and cache RSS feeds - this can be simply improved to add more RSS sources in the future
- **Lockfile mechanism** (`/tmp/.update_rss_feeds_running`) to prevent concurrent executions - following other cron jobs
- **XML validation** before caching to ensure only valid RSS data is stored - simplified the display code
- **Debug page integration** for manual trigger and status monitoring

#### DX Calendar Performance Optimization
- **New batch query function** `get_batch_dxcalendar_statuses()` in Logbook_model
- Reduces database queries from **4×N** (4 queries per DXpedition) to just **4 total queries**
- Uses `IN` clauses to check all callsigns/DXCCs at once

#### Frontend Improvements
- Calendar pages now read from cache only (no external fetches)
- Role-based error messages:
  - **Admins**: See instruction to trigger update from Debug page
  - **Regular users**: See message to contact administrator

### Migration

Migration **268** adds the new cron job:
- **ID**: `update_update_rss_feeds`
- **Schedule**: `0 */12 * * *` (every 12 hours)
- **Description**: Update RSS feeds for DX Calendar and Contest Calendar

### Testing

1. Run migration to add the cron job
2. Navigate to Debug page and trigger "RSS Feeds" update
3. Verify DX Calendar and Contest Calendar load from cache
4. Check that cache files are created in `application/cache/`

### DIFF
--- a/application/config/migration.php
+++ b/application/config/migration.php
@@ -22,7 +22,7 @@ $config['migration_enabled'] = TRUE;
 |
 */
 
-$config['migration_version'] = 267;
+$config['migration_version'] = 268;
 
 /*
 |--------------------------------------------------------------------------


HB9HIL:
Usually here is caching implemented. Imho it's better to only load the data on the first load and cache it instead loading data you may never need. If the caching does not work on your side you may suffering a bug or something.

int2001:
same view as fabian here.
Contest- and DXPed-Cal already have a caching logic.
yes - 1st user of day may wait a bit longer.

but cron doesn't solve a possible issue - if there's a network issue - either.

PR #2635, opened 2025-12-03
[SimpleFLE] Fix date since we tightened up the checking

AndreasK79:


int2001 (approved):
works. tested:
- English and German
- different Date-Formats

PR #2637, opened 2025-12-04
Catch error for calling a deleted qso

int2001:
# Scenario:
Two tabs.
- One with Overview/Dashboard/LBA (something where one can click on a made QSO)
- Second tab: Delete a QSO

After deletion go back to Overview/Dashboard/LBA and click on the previous removed QSO. Results in a 500.
This patch fixes it, and shows a message "QSO not found" instead.


AndreasK79:
Instead of no visible error (other than devtools), now at least gives the user a message :)

PR #2642, opened 2025-12-05
Exp chunked

int2001:
## Problem
Even if really much memory is delegated to the application, it may happen that in runs into Memory-Exhaust when trying to export huge amounts of QSOs
Designproblem is: Select demanded QSOs into mem. write them to client

## Solution
Chunkwise export and streaming

## affected
- [x] export_all
- [x] export_custom
- [x] API / especially the API which was implemented in #664 if the client doesn't respect limits.

Testing Hints:
- Export ADIF in all flavours. Compare to export without patch (Caution: Ordering may be different, but content should be the same)
- 3rd-party exports. Not 100% sure / but i tried to leave them on the old path (fixing this in further PR)

### DIFF
--- a/application/controllers/Adif.php
+++ b/application/controllers/Adif.php
@@ -103,11 +103,7 @@ class adif extends CI_Controller {
 		header('Content-Disposition: attachment; filename="'.$filename.'"');
 
 		// Output ADIF header // No chance to use exportall-view any longer, because of chunking logic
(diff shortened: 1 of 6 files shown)


DB4SCW:
Tested ADIF export API. Works flawlessly on my side.

phl0 (approved):
Export functions still working :)

PR #2644, opened 2025-12-05
Fix club edit

int2001:
Accidently built in a bug with the new Role, where even a normal clubmember wasn't able to edit his own QSOs.

Testing Hints:
- Log in as a normal User, which is granted for a clubstation (incognito session / real clubmember-user, no impersonante)
- Switch to Clubstation into the clubstation and log a QSO.
- Try to edit it

without patch: no way

with patch: possible (because OP is the same)


phl0 (approved):
Tested and fixes the issue.

PR #2653, opened 2025-12-08
Show stationlist for ClubMembers

int2001:
Problem:

Clubmembers are not able to see valid station_ids for the clubstation they're granted for (except API).

Solution:
Show Stationlist (List only) for Clubmember and Clubmember ADIF without "import function".

Test:
Create a User, grant it to:
a) Clubmember
b) Clubmember ADIF
c) ClubOfficer

"Impersonate" to ClubStation with those three Users, and test:
- User a and b should see "Station Setup" (Since it shows the Setup) in their Usermenu. Page is only the List without Import-Function
- User c should see "Station Setup" as well, but in a normal manner.
- Adminuser should see/do everything as before

### DIFF
--- a/application/controllers/Station.php
+++ b/application/controllers/Station.php
@@ -7,15 +7,15 @@
 class Station extends CI_Controller
 {
 
-	function __construct()
-	{
+	function __construct() {
 		parent::__construct();
 		$this->load->helper(array('form', 'url'));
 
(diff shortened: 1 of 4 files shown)


phl0 (approved):
Tested and works.

phl0 (approved):

PR #2665, opened 2025-12-09
Implementing callbook failover logic

iu2frl:
This MR implements a very simple callbook failover logic.

## Summary

- If the `$config['callbook']` is set to a single string (default), nothing is changed in the logic, a single query is performed and result is evaluated accordingly
- If the `$config['callbook']` is set to an array, then the different providers are sequentially tested to see if a valid response is returned

## Testing

I tested the following scenarios:

### Back compatibility

I let the default `config.php` config (or with any single providers set):

```php
$config['callbook'] = 'qrz';
```

And the following sequence was returned (single callbook -> single request). The single provider was tested, nothing different from existing behavior.

```php
DEBUG - 2025-12-09 16:32:56 --> Callbook lookup for R2VA using qrz: Not found: R2VA
```

### New behavior

I set the `config.php` config to an array of providers, and the following sequence was returned (providers are tested in the same sequence as defined in the config):

```php
// $config['callbook'] = ['qrz', 'qrzcq'];
DEBUG - 2025-12-09 16:37:49 --> Callbook lookup for R2VA using qrz: Not found: R2VA
DEBUG - 2025-12-09 16:37:50 --> Callbook lookup for R2VA using qrzcq: Errore QRZCQ: Premium Required
// $config['callbook'] = ['qrz', 'qrzcq', 'hamqth'];
DEBUG - 2025-12-09 16:38:30 --> Callbook lookup for R2VA using qrz: Not found: R2VA
DEBUG - 2025-12-09 16:38:31 --> Callbook lookup for R2VA using qrzcq: Errore QRZCQ: Premium Required
DEBUG - 2025-12-09 16:38:32 --> Callbook lookup for R2VA using hamqth: Callsign not found
// $config['callbook'] = ['hamqth', 'qrzcq', 'qrz'];
DEBUG - 2025-12-09 16:51:40 --> Callbook lookup for R2VA using hamqth: Callsign not found
DEBUG - 2025-12-09 16:51:40 --> Callbook lookup for R2VA using qrzcq: Errore QRZCQ: Premium Required
DEBUG - 2025-12-09 16:51:44 --> Callbook lookup for R2VA using qrz: Not found: R2VA
```


HB9HIL:
Before we are going to check this PR:

Really well done Pull Request! Simple, Short, Single Feature, Necessary Changes only 

Really Really good!

application/libraries/Callbook.php
```diff
@@ -18,7 +18,40 @@ public function __construct() {
...
-		switch ($this->ci->config->item('callbook')) {
+		// Load callbook configuration from config.php
+		$source_callbooks = $this->ci->config->item('callbook');
+
+		// Check if the source callbook is a single element or an array
+		if (is_array($source_callbooks)) {
+			// Parse each callbook in the array until we get a valid result
+			foreach ($source_callbooks as $source) {
+				$callbook = $this->queryCallbook($callsign, $source);
+				if (!isset($callbook['error'])) {
```
phl0:
```suggestion
				if (!isset($callbook['error']) || $callbook['error'] == '') {
```

At least hamqth lib gives empty error message back. We should cover this case as well as error being undefined.

application/libraries/Callbook.php
```diff
@@ -46,15 +79,8 @@ public function getCallbookData($callsign) {
...
-		if (! array_key_exists('geoloc', $callbook)) {
-			$callbook['geoloc'] = '';
-		}
-		// qrz.com gives AA00aa if the user deleted his grid from the profile
-		$this->ci->load->library('qra');
-		if (!array_key_exists('gridsquare', $callbook) || !$this->ci->qra->validate_grid($callbook['gridsquare'])) {
-			$callbook['gridsquare'] = '';
-		}
+
+		log_message('debug', 'Callbook lookup for '.$callsign.' using '.$source.': '.(isset($callbook['error']) ? $callbook['error'] : 'Success'));
```
phl0:
```suggestion
		log_message('debug', 'Callbook lookup for '.$callsign.' using '.$source.': '.((($callbook['error'] ?? '' ) != '') ? $callbook['error'] : 'Success'));
```

Same as above. We need to catch empty errors as well.

phl0 (changes requested):
The basic functionality works for me. But hamqth lib returns error being empty so the check for success will always fail for hamtqth. Added two suggestions to change this behavior.

application/libraries/Callbook.php
iu2frl: Thanks @phl0 I will check that

iu2frl:
@phl0 I implemented your suggestions

phl0:
> @phl0 I implemented your suggestions

Seen that. Will re-check.

phl0 (approved):
This works for me. We should put an example and some description into config.sample.php (if not in this PR I can add it separately).

iu2frl:
@phl0 ok, I will update the wiki later tonight

AndreasK79 (approved):

AndreasK79:
Tested working here :)

iu2frl:
I started documenting the config file, with some notes on this feature: [Config.php](https://github.com/wavelog/wavelog/wiki/config.php-Configuration-file)

phl0:
> I started documenting the config file, with some notes on this feature: [Config.php](https://github.com/wavelog/wavelog/wiki/config.php-Configuration-file)

And I added some description to config.sample.php 😇

PR #2669, opened 2025-12-10
Fix user option missing

phl0:
User reported that gridmap was not working. Bug was that some colors were not defined in user profile. Reason is *probably* migration from Cloudlog. Also user profile settings page breaks. So we set some default fallback values in case every thing else fails / is not set.

Repro: Delete map_custom - icon - qsoconfirm from user_options.

image

image


int2001:
GM. Checked a few Maps/Awards, and had to stop at the first Award (DXCCMap)
Still buggy. e.g.: dxccmap still fails because confirmed icon isn't checked

PR #2670, opened 2025-12-10
Fix hamqth lib

phl0:
Adds an error handler to hamqth lib to gracefully fail if hamqth is not reachable. Also fixes an issue with image width for hamqth profile pictures:

image


int2001:
testing commenting. will check PR later

PR #2680, opened 2025-12-13
Check Club-Keys more carefully on API

int2001:
The best Grant-concept doesn't work if it's bypassable by API.

This PR fixes the following things:
- [x] Force Operator to Club-OP if it's a Member or ADIF-Member at logging. Officer and normal User aren't affected
- [x] more coming


phl0 (approved):
Tested and works. Location can only be created for auth level club officer.

PR #2681, opened 2025-12-13
[DXCluster] CAT QRG line is diffcult to see in light themes

AndreasK79:
image


int2001:
just, that it won't be lost, goal should be:
take color for QRG-Bar (and gradient) from CSS-Theme.

int2001 (approved):

PR #2688, opened 2025-12-17
[DXCC id] Tweaking the DXCC id logic

AndreasK79:
While playing around, I noticed two callsign that would not id as Bahamas (WJ7R/C6A and 9H5G/C6A).
I added some variants to my calltester, which assists in testing.

This PR fixes this problem, and also adds a few extra additions to callsigns, that should be ignore, namelig /J /L /SK.

### DIFF
--- a/application/controllers/Calltester.php
+++ b/application/controllers/Calltester.php
@@ -225,6 +225,27 @@ class Calltester extends CI_Controller {
     function call() {
         $testarray = array();
 
+		$testarray[] = array(
+            'Callsign'  => 'WJ7R/C6A',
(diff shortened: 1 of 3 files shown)


phl0 (approved):
Tested and works. Correctly identifies the callsigns as per calltester.

phl0:
Just an idea: Maybe move `$csadditions` to some central place/lib to get rid of redundancies?

AndreasK79:
@phl0 good idea. We have the dxcc class. I'll leave it for a future refactor, but it is possible.

PR #2691, opened 2025-12-19
Hide marker if grid is not set and DXCC is NONE

phl0:
Hide marker in QSO view if grid is empty.

image

### DIFF
--- a/assets/js/sections/qso.js
+++ b/assets/js/sections/qso.js
@@ -1650,7 +1650,7 @@ $("#callsign").on("focusout", function () {
 
 				// Set Map to Lat/Long
 				markers.clearLayers();
-				mymap.setZoom(8);
+				var zoom = 8;
 				// Remove previous banner (if any)
 				if (window.mapBanner) {
 					mymap.removeControl(window.mapBanner);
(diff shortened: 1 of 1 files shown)


int2001:
Hmm. Fair for Stations like NA1SS or /MM stations.
but marker is also gone for "normal" stations without grid. intended?

phl0:
> Hmm. Fair for Stations like NA1SS or /MM stations. but marker is also gone for "normal" stations without grid. intended?

Where to put the marker in case grid is not existent?

int2001:
Ideas:
- filter for /mm /am in combination with no grid. Maybe a blacklist on top?!
- list the QSOs without grid as transparent overlay but with the standard-colors?
- add a checkbox to the small "show grid"-box, called: "show stations without grid?"

I mean: we added the function to place QSOs without grid to the center of the DXCC. This PR would remove that functionality completely, and, tbh: on Shortwave nobody cares about grid...

phl0:
Seems it affects more than just the map in QSO view. So drafting it for now. Will check more closely.

phl0:
Found a(nother) solution to this. Marker is only hidden if grid is not given AND DXCC is none as that was the case I wanted to cover. So code was refactored to deal with this.

NA1SS with no grid and DXCC set to NONE:
image

Other (HF) QSOs without grid but DXCC:
image

AndreasK79 (approved):

PR #2693, opened 2025-12-19
Respect BandEdges for Cluster

int2001:
## Problem:
if using bandlist (List DXCluster), neither the user defined Bandedges nor the default ones are respected.
which means: If there are spots out of Hamradio-Bands, they're shown.

## Solution:
Filter spots against BandEdges (User defined w. fallback to default).

## Testing:
- Filter on 30m
- Wait e.g. for a spot at 10.489 (common error, because QO100-Spots are mapped to 30m band. Rootcause is a different one: Handling of huge uints in clusters - but good to test for this one)
- now try with patch: Should be gone.

other testing-approach: Play around with userdefined BandEdges. they should be respected with the patch

### DIFF
--- a/application/models/Dxcluster_model.php
+++ b/application/models/Dxcluster_model.php
@@ -167,6 +167,11 @@ class Dxcluster_model extends CI_Model {
 				// Ensure frequency is always a number (not a string)
 				$singlespot->frequency = floatval($singlespot->frequency);
 
+				// Validate against amateur band allocations (skip non-amateur frequencies)
+				if (!$this->isFrequencyInAmateurBand($singlespot->frequency)) {
+					continue;
+				}
+
 				$spotband = $this->frequency->GetBand($singlespot->frequency*1000);			// Apply band filter early (before expensive operations)
 			if (($band != 'All') && ($band != $spotband)) {
 				continue;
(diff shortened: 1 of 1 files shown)


xyz667:
Why not respect all band settings - not only the edges? See #2588 .

int2001:
Simple: bands are different over the world. 
If it "annoys" you, ignore those Bands or remove them at your BandEdge-Setup.

xyz667:
But isn’t this the whole point of the „Bands“ configuration menu?
To ignore bands!
Not sure what you mean with „ remove them at your BandEdge-Setup“ - removing bands from the Cluster is the content of issue #2588 ?

int2001:
This PR solves the issue showing Spots which are outside the Bands you defined/changed.

Removing whole Bands from the cluster is a total different story. Workaround, if it's so annoying for you: tighten the Bandedges of those Bands to 1Hz, and it's done.

Or did I miss something?

PS: we're open for PRs. So if you have a good idea, with a simple UI/UX to solve the "problem": go ahead and contribute. We appreciate every new contribution and help.

AndreasK79 (approved):

AndreasK79:
@int2001 these do not show with this PR:
image

PR #2694, opened 2025-12-20
[DBTools] Some fixes

AndreasK79:
Fixes so far:

- [x] Custom date in QSOs missing DXCC dialog
- [x] Enable callbook grid fixer as default
- [x] Speedup state fixing
- [x] Added batch update for callbook grid updater (though each lookup to callbook takes time)


phl0 (approved):
Tested and works fine (found a bug in qrzru lib. Will fix separately).

PR #2698, opened 2025-12-22
Fix: Decode a few Swedish strings that shows up encoded in the UI

sm6srw:
There are a few swedish strings that shows up encoded in the UI that I know about.  

These are in the:
* new Waterfall
* Dx-Cluster
* Days with QSOs view
* Statistics view

This PR fixes these by using the new utility function `decodeHtml()` or by using the `html` property instead of the `text` property.

Are we OK with fixing these on a case by case basis? Other languages must have the same problem in other places? Or is it possible to fix this in a generic way?


phl0:
Thanks for this addition. Much apprechiated :)

PR #2699, opened 2025-12-22
[Advanced Logbook] Added a dialog with parameter for duplicate search

AndreasK79:
image

### DIFF
--- a/application/views/logbookadvanced/dupesearchdialog.php
+++ b/application/views/logbookadvanced/dupesearchdialog.php
@@ -31,7 +31,7 @@
                         <label class="form-check-label fw-semibold" for="band_check">
                             <i class="fas fa-wave-square me-2 text-warning"></i><?= __("Band"); ?>
                         </label>
-                        <small class="d-block text-muted"></small><?= __("Match QSOs on the same band"); ?></small>
+                        <small class="d-block text-muted"><?= __("Match QSOs on the same band"); ?></small>
                     </div>
                 </div>
                 <div class="col-md-6">


AndreasK79:
@iu2frl got some time to implement this. Think I got it working as it should now.

iu2frl:
> @iu2frl got some time to implement this. Think I got it working as it should now.

I haven't tested it but it looks awesome!

application/views/logbookadvanced/dupesearchdialog.php
```diff
@@ -0,0 +1,49 @@
...
+                        <small class="d-block text-muted"><?= __("Match QSOs with the same mode (SSB, CW, FM, etc.)"); ?></small>
+                    </div>
+                </div>
+                <div class="col-md-6">
+                    <div class="form-check form-check-lg border rounded p-3 bg-light h-100">
+                        <input class="form-check-input" type="checkbox" name="band_check" id="band_check" checked>
+                        <label class="form-check-label fw-semibold" for="band_check">
+                            <i class="fas fa-wave-square me-2 text-warning"></i><?= __("Band"); ?>
+                        </label>
+                        <small class="d-block text-muted"></small><?= __("Match QSOs on the same band"); ?></small>
```
int2001: the small is double closed. intended?

phl0 (approved):
Code works. HTML tag could be removed as @int2001 outlined.

application/views/logbookadvanced/dupesearchdialog.php
phl0:
```suggestion
                        <small class="d-block text-muted"><?= __("Match QSOs on the same band"); ?></small>
```

phl0: Added @int2001's change as a suggestion :)

AndreasK79: @int2001 oops, guess I didn't see that one.

PR #2705, opened 2025-12-23
Store CAT-Radio to My_rig

int2001:
Once a CAT-Radio (Station Tab) is chosen, the name of the radio is saved at ADIF "MY_RIG".

image

Furthermore the Radio-Name is shown when viewing the QSO at Station-Tab.

image

if no CAT-Radio is chosen no radio-name is saved and no radio-name will be shown

### DIFF
--- a/application/views/view_log/qso.php
+++ b/application/views/view_log/qso.php
@@ -709,6 +709,13 @@
                     </tr>
                     <?php } ?>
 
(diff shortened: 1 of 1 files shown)


AndreasK79:
Tested cat polling. That works. Websocket does not save.

PR #2714, opened 2025-12-25
Dbtools dxcc fix

AndreasK79:
Instead of blindly trusting the dxcc updater, you can now see the dxcc differences in a table, and choose which QSOs to update.

Advanced Logbook -> dbtools -> dxcc checker


int2001 (approved):
tested a bit forth and back. good job. works!

don't forget to mention config-switch for gridsquare-thingy within wiki

Issue #2713, opened 2025-12-25
eQSL card thumbnail preview in eQSL interface


int2001:
#### Needed Feature:
- Thumbnails at eQSL-List-View for the already downloaded QSOs

#### Things to check b4 implmentation:
- Resize on Client or on Server (last one could be performance-breaking)?

HB9HIL:
I vote for resize on server

int2001:
was also my first impulse. but client is very quick.
eQSL-Images are already very small, so traffic isn't the problem. building a PoC atm.

Edit: Turns out: rescaling on serverside is more quick

PR #2715, opened 2025-12-25
Eqsl thumbnails

int2001:
Implements a thumbnail-view to "View eQSL" at Logbook-Menu.
Serverside-Scaling, because files are smaller and doesn't take that time

#2713


phl0 (approved):
Page loading time increasing with No of pictures and renders unusable with 100s of cards. But basically gives a nice view.

PR #2727, opened 2025-12-27
Remove redundant function to mark QSOs as uploaded to LoTW

phl0:
We have a redundant function to mark QSOs as uploaded to LoTW. It seems that at least the function on the ADIF (import/export) page is misleading and people tend to think this is the LoTW upload function. As the LBA offers a much more convenient way to mark QSOs as uploaded we gonna remove this one. (Closes DARC Ticket #20383767).


phl0:
> perhaps alter this text here as well and add an "already" to the caption?

Added.

PR #2729, opened 2025-12-28
[QSO Map] Tweaks, features, refactor

AndreasK79:
Done:
- [x] Moved JS to it's own file
- [x] Added CQ Zone layer
- [x] Added ITU Zone layer
- [x] Added gridsquare layer
- [x] Rewrote to plain sql
- [x] Removed redundant function for fetching locations
- [x] Added mouseover for grid/distance/cq/itu/bearing
- [x] Made map height responsive
- [x] Added prefix in country dropdown, and sorted by prefix
- [x] Add mouseover to display info about region/county/state

Test:
Goto wavelog-url/map/qso_map

### DIFF
--- a/assets/js/sections/qso_map.js
+++ b/assets/js/sections/qso_map.js
@@ -186,7 +186,7 @@ function initMap() {
                                 color: '#ff0000',
                                 weight: 2,
                                 opacity: 0.5,
-                                fillOpacity: 0.1
+                                fillOpacity: 0.2
                             },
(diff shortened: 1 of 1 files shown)


phl0 (changes requested):
Needs two things fixed:

Check geojson boundaries for some country's states:

Screenshot from 2025-12-28 22-12-28



Need to fix error with additional legends appearing on map change:

image

PR #2730, opened 2025-12-28
Show info about source callbook

phl0:
Now as we allow multiple callbooks to be queried it might be useful to show a hint about in which callbook the information about a callsign was found.

Test: Search for or prepare logging of an un-worked callsign.


AndreasK79 (approved):

AndreasK79:
Works :)

PR #2731, opened 2025-12-29
Geojson updated maps

AndreasK79:
This is based on #2729, so it needs to be merged first.

I updated the map for DL (230).


AndreasK79:
Something strange started happening while I edited the geojson. Suddenly coordinates within states got reported as not being inside. Closing this.

PR #2741, opened 2025-12-30
Auto update only for single ITU zones

phl0:
The update function for ITU zones will set the (single one) ITU zone matching the DXCC in dxcc_entitites table. In case the DXCC covers multiple ITU zones this will go wrong. So we enhanced the auto updater to only do the auto change for DXCC with single ITU zone. The rest will have to be updated manually (for now).


AndreasK79 (approved):

AndreasK79:
Works. Found reason for those not updating. That will be done in another PR.

PR #2744, opened 2025-12-31
Dbtools another round with fixes

AndreasK79:
Relies on #2741, which must be merged first.

1. QSOs with DXCC not set didn't make it into the list
2. Removed "Check QSOs missing DXCC data", as it is redundant
3. Added logic for the force multi update to remove filer and checked lines for table in CQ/ITU zone


phl0 (approved):
All fine

PR #2747, opened 2025-12-31
[DBTools] Fix for hanging loop

AndreasK79:
If you have several thousand entries to correct, browser hangs.

Moving draw for datatable outside loop.

Removed backend stuff not needed anymore.


int2001 (approved):
works

PR #2751, opened 2026-01-01
Tune qso save

int2001:
At QSO-Saving there are multiple things done in a redundant manner.
this PR optimizes them


AndreasK79:
Nice with speed :)

Issue #2707, opened 2025-12-23
Login to LoTW fails


ea5wa:
### I have checked the following

- [x] I am running the latest version of Wavelog
- [x] I have cleared my browser cache
- [x] This is a bug, not a setup question or feature request

### Bug Description

Not posible to connect to LoTW with Test Connection button:

<img width="737" height="651" alt="Image" src="https://github.com/user-attachments/assets/ba3ff2e8-808d-47e1-93c0-f734c56bdef7" />

Testing these credentials with the web browser are ok.

In a Cloudlog installation in the same WAMP64 server is working perfectly

### Steps to Reproduce

1. Go to Account Settings
2. Go to Thir Party Services
3. Enter user and password and click "Test Connection"

### Expected Behavior

Error message "Error: LoTW currently not available. Try again later." is shown

### Wavelog Version

2.2.1

### PHP Version

8.3.6

### MySQL/MariaDB Version

MySQL 8.3.0

### Server OS

WAMP64 Windows 11

### Browser

Edge

### Device Type

Desktop

phl0:
For some reason it seems that the cURL call receives an empty content. Could you please add this debug code:


```
diff --git a/application/controllers/Lotw.php b/application/controllers/Lotw.php
index 90686c458..50169ca22 100644
--- a/application/controllers/Lotw.php
+++ b/application/controllers/Lotw.php
@@ -849,6 +849,7 @@ class Lotw extends CI_Controller {
                        curl_setopt($ch, CURLOPT_RETURNTRANSFER, true);
                        curl_setopt($ch, CURLOPT_CONNECTTIMEOUT, 30);
                        $content = curl_exec($ch);
+                       log_message('debug', 'Content: '.$content);
                        if ($content) {
                                if(curl_errno($ch)) {
                                        $ret['status']='failed';
```

And have a look at the log file if it dumps any content?

ea5wa:
Where do I have to add this debug code?

phl0:
> Where do I have to add this debug code?

In the file application/controllers/Lotw.php in line 852.

ea5wa:
Inside the IF? Where? Keep in mind that I'm not a programmer.....

phl0:
[...]

PR #2755, opened 2026-01-01
Make LoTW login test code more bullet-proof

phl0:
Based on https://github.com/wavelog/wavelog/issues/2707 we reworked the LoTW code a little. If curl fails (because of SSL issues for example) $content will be empty and the user sees the misleading error message that LoTW is not available. This is just wrong in this case. Instead it is an issue with local SSL installation in this case for example.
Also adds some debug log output in case connection fails.

### DIFF
--- a/application/controllers/Lotw.php
+++ b/application/controllers/Lotw.php
@@ -847,8 +847,8 @@ class Lotw extends CI_Controller {
 			$ch = curl_init();
 			curl_setopt($ch, CURLOPT_URL, $lotw_url);
 			curl_setopt($ch, CURLOPT_RETURNTRANSFER, true);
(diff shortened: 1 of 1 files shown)


application/controllers/Lotw.php
```diff
@@ -849,20 +849,22 @@ public function check_lotw_credentials () {
 			curl_setopt($ch, CURLOPT_RETURNTRANSFER, true);
 			curl_setopt($ch, CURLOPT_CONNECTTIMEOUT, 30);
```
int2001: perhaps add a "normal" timeout as well. not only the connect one
```suggestion
curl_setopt($ch, CURLOPT_CONNECTTIMEOUT, 10);
curl_setopt($ch, CURLOPT_TIMEOUT, 20);
```

int2001 (approved):

PR #2758, opened 2026-01-02
Add last modification time to LBA

phl0:
This exposes the last modification time per QSO in LBA if column is selected to be shown. This can be used for debugging purposes (e.g. upon update of POTA refs or DCL import).


AndreasK79:
Works

PR #2759, opened 2026-01-02
Last modified should not be shown if not configured at all

phl0:
Relates to https://github.com/wavelog/wavelog/pull/2758. The column should not be shown if not configured at all.


AndreasK79 (approved):

AndreasK79:
Hidden by default

PR #2760, opened 2026-01-02
[Zonechecker] Added an internal tool

AndreasK79:
Added link to internal tools, but they need to activated in the config:

$config['internal_tools'] = true;

They are disabled by default.

image

This PR also adds the zonechecker.


phl0 (changes requested):
Should add some explanation and a config item to application/config/config.php.sample. Rest of the code works fine :)

AndreasK79:
> Should add some explanation and a config item to application/config/config.php.sample. Rest of the code works fine :)

Done

phl0 (approved):

phl0:
was about to have other icons. These ones are better suited :)

PR #2766, opened 2026-01-03
Catch LoTW download rate limit and log accordingly

phl0:
Just distinguish this rate limit error from generic error message saying that the report was invalid. In case of the rate limit we just don't get any kind of report at all. This should be logged accordingly. Option: Delete LoTW credentials as well to prevent further polling. This needs some more investigation of how this rate limit is reached and how to release the root cause. So this is left out of this PR for now but may be added later.

### DIFF
--- a/application/controllers/Lotw.php
+++ b/application/controllers/Lotw.php
@@ -766,7 +766,7 @@ class Lotw extends CI_Controller {
 						break;
 					}
 					continue;
-				} else if(str_contains($content,"Username/password incorrect</I>")) {
+				} else if(str_contains(substr($content,0 , 2000),"Username/password incorrect</I>")) {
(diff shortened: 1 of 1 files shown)


application/controllers/Lotw.php
```diff
@@ -775,6 +775,10 @@ function lotw_download($sync_user_id = null) {
 						log_message('error', 'Deleting LoTW credentials for user '.$user->user_name.' failed');
 					}
 					continue;
+				} else if (str_contains($content,"Page Request Limit!</B>")) {
```
int2001: perhaps add a substr here to check only the first 2kB ?!
Otherwise the whole ADIF would be grepped, if it is an ADIF... (performance?!)

phl0: Pushed a change to check the first 2000 chars. Ok?

PR #2767, opened 2026-01-04
Relax fetch of DX Cluster data

user00265:
DX Cluster data is being fetched with a minimally-initialized cURL instance, and the parameters do not accept any redirects. I personally do some proxy trickery to get this to work and allowing a redirect or two would help make this work normally instead of having to default to exposing dxcache directly.

### DIFF
--- a/application/models/Dxcluster_model.php
+++ b/application/models/Dxcluster_model.php
@@ -99,6 +99,8 @@ class Dxcluster_model extends CI_Model {
 			curl_setopt($ch, CURLOPT_TIMEOUT, 10);
 			curl_setopt($ch, CURLOPT_HEADER, false);
 			curl_setopt($ch, CURLOPT_RETURNTRANSFER, true);
+		    curl_setopt($ch, CURLOPT_FOLLOWLOCATION, true);
+    		curl_setopt($ch, CURLOPT_MAXREDIRS, 3);
 			$jsonraw = curl_exec($ch);
 			$curl_error = curl_error($ch);
 			$http_code = curl_getinfo($ch, CURLINFO_HTTP_CODE);


phl0 (approved):

HB9HIL:
@user00265 

I don't really understand why you need this. `followlocation` can be an security issue in some cases. Therefore there would be a benefit for all Wavelog Users on the world when accepting this PR.

user00265:
Benefit? Some advanced setups, or non-standard setups where one chooses to redirect. For example, a user enters http://[..] and the server redirects to https://[..] -- this would be completely broken with the current code.

I do hear you on the security concerns in certain cases, but those cases are marginal at best, unless you have examples I am unaware of. Almost all scenarios I can work with involve control of the network, DNS server, application itself, or the target host the user selected for the URL. All of which are quite severe and compromise of their network as a whole is the bigger concern here.

May I know your specific concerns so that I may inform myself, or perhaps learn something new?

int2001:
Help us.
The DXCLuster-URL has to be set by the instance admin within WL, not by the user. So instance admin should know what he's doing there.
Very very unusual to provide an API which isn't reachable directly but returns a redirect instead.

Of course it wouldn't destroy anything, but whats "advanced" with 301ing an API-Endpoint?

user00265:
In my mind an end user is either: A user that either uses the application with no administration privileges, or one that has both. I do not assume the user with administrative privileges is trustworthy and (to an extent) knowledgeable. I assume they are someone who intends to use an application, and their level of knowledge varies. What I am counting on is human nature; Most users prefer or use commands copied & pasted from somewhere, whether that be a friend, a README, a manual, or an article online. I assume none of those are trustworthy by default.

Most users want things to work, and then they don't, then there is either confusion or dissatisfaction. If I tell @int2001 as the admin of his instance to use `http://dxc.example.com` but I forgot (or am not aware) that either HTTP redirects to HTTPS or that the host is not listening on HTTPS, or the he types http instead of https on his own, then there's an issue.

What I am saying is I am making a conscious choice of a possibility for ease of use instead of focusing on a possibility which would have to be thoroughly exploited by compromising the network, host, or network connection this application resides in, at which point the system is compromised either way...

I already provided one use case: HTTP to HTTPS redirection (I am choosing the more secure approach), which is common in environments that use Cloudflare or use domain TLDs that force HTTPS like `.dev` for example. I have one use case where I send the request to an edge proxy and redirect with a key hash in the URL that gets then sent to another host for access -- I rotate the backends. I use this for testing.

In summary, a very low possibility of abuse with a sensible limit (I would be okay with even 1 redirect, but normal seems to be 2-3) was my choice to relax it just enough to make it useful, but not enough to cause issues.

On another note, debugging these issues requires one to set debug logging and watch the log file because on the UI it is not immediately obvious there is an issue. I spent several hours understanding the code of Wavelog to figure out that it fails silently to the UI.

int2001:
Good arguments. But the main question is still: why provide an API which doesn't answer on a request but throws a 301? 

Very unusual and imho bad practice. If a tool exposes an api it should provide the real api behind the endpoint and not a 301. the only (half legitimate) use would be: new version, redirect for compatibility reasons. But tbh: even this is bad practice.

IMHO Hard to find the balance between "yolo, I've no clue what I am doing" and "well architecture".

I am totally with you when it comes to browser redirection. Common practice there. But not for APIs.

Or are we talking about different things?

If - e.g. - a WL-Admin inputs http instead of https for the cluster, and the cluster isn't reachable there, it's crystal clear. For fresh installations there's even a hint for the default cluster...

Best practice: configure your cluster to listen on http and https instead of redirecting. Easy one with all kind of Webservers/reverse proxies. I mean: instance owner was able to setup WL and perhaps cluster. Hobby is about learning how things work.

user00265:
In a perfect world, I totally agree with you and this would not be a PR I'd consider, but we do not live in a perfect world.

The http/https redirect argument which is the most obvious example of not a perfect world, user error is another, as we've both touched on, and then there's usability. Is it perfect would practice? No. Is it a corner case? Not as much as you'd expect. User error, from experience, makes this more common.

Unless you provide feedback that the URL won't work and refuse to save (ideal), or provide visual feedback that the URL is not working when the endpoint is queried instead of `{"error": "not found"}` from the API and the UI just sits there with nothing going on like nothing is going on (if you combine this for temporary issues with the previous for the save-time check would be ideal), or relax it a little you are basically hurting usability from a UI perspective.

In a perfect world, the DX cache client would do a fetch with exactly 1 redirect as the maximum accepted, validate, then show the user the end result.

By validating I mean attempting to fetch the URL, report any errors, validating whether it is a simple http -> https redirect and if so, we would allow it since the redirect limit being 1 would take care of it, look to make sure the output looks like valid JSON, then allow if all works. We would also look at any non-200 return codes, zero content body returned, or any other general failure. The whole ides behind this being to let the application user (admin or otherwise) know that something is wrong, and the admin know when trying to save that the URL will not work.

I could actually work on this, thinking about it, though this being a longer project as I'd have to acquaint myself with the codebase and how to notify the UI from the backend the way Wavelog does.

int2001:
Sorry, going to close this here.
Reason:
for 301 the Admin/OP has to implement explicitly the 301. it doesn't come "alone". if one has the opinion to expose his API behind a 301 he knows what he's doing.

PR #2771, opened 2026-01-05
Emit ADIF to Websocket (if in qso) on logging

int2001:
If CAT via Websocket is enabled, emit a logged QSO to Websocket (details and adif).
This one is for preparing more interfaces.

The UDP/2333-Standard for example. With a further development of WavelogGate or other tools it would be possible to emit QSOs as ADIF via UDP/2333 to other applications.

Testing Hints:
- Enable CAT/Websocket
- inspect (Debug-Console) network-tab / websocket message
- Log a QSO
- a meesage with the type "qso_logged" should appear. it includes a lot of informations about the QSO.


phl0 (approved):
image

Basically works. I can see the QSO ADIF being transmitted.

PR #2777, opened 2026-01-05
CQ and ITU zone fixes

AndreasK79:
While doing some digging, I found a few issues around CQ and ITU zones.

1. The advanced logbook doesn't set CQ zone from callbook if the zone is blank.
2. When editing a QSO, you can't set the CQ zone to blank.
3. When doing a batch edit in the advanced logbook, you can't set CQ or ITU zone to blank.

This PR will fix these issues.


phl0 (approved):
Tested with editing zones back and forth. Setting zone as well as deleting zones works.

PR #2780, opened 2026-01-06
Emit Az/Ele - when working SAT - via Websocket

int2001:
This PR implements pushing Az/Ele, when working a SAT, via Websocket.

Testing:
- Enable Websocket (Needs WavelogGate or WavelogStoat running)
- Chose a SAT in QSO-Window
- Inspect Websocket-Data
- Should stop, if CAT is disabled or SAT is removed from QSO

it's preparation for Rotator-Control.
i'll add the same for HF / Beamdirection in another PR.

todo's (outside this PR): Catch data in Stoat/Gate and forward it to rotctld

### DIFF
--- a/assets/js/cat.js
+++ b/assets/js/cat.js
@@ -510,6 +510,54 @@ $(document).ready(function() {
     window.sendQSOViaWebSocket = sendQSOViaWebSocket;
 
     /**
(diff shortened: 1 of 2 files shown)


phl0 (approved):
Works as expected (forgot to make a screenshot here) :)

PR #2783, opened 2026-01-07
Fetch callbook for api private_lookup if requested

int2001:
This one adds an optional parameter to the [private_lookup API ](https://github.com/wavelog/wavelog/wiki/API#apiprivate_lookup)called "callbook". if set to true, the endpoint delivers the callbookinformation along with the worked/cnf-informations

Should help at #2782 

Testing (with and without qrz-field):

`curl -X POST -H "Content-Type: application/json" -d '{"key":"[WL_KEY]","callsign":"DJ7NT","callbook":"true"}' https://[YOUR_WL]/index.php/api/private_lookup`

expected result (with callbook=true and a bit patience):
```
{
    ...
    "callbook": {redacted}
    }
}
```

### DIFF
--- a/application/controllers/Api.php
+++ b/application/controllers/Api.php
@@ -843,7 +843,7 @@ class API extends CI_Controller {
 			$user_id=$this->session->userdata('user_id');
 		}
 
-		if (($raw_input['qrz'] ?? '' == 'true') && (($raw_input['callsign'] ?? '') != '')) {
+		if (($raw_input['callbook'] ?? '' == 'true') && (($raw_input['callsign'] ?? '') != '')) {
 			$this->load->library('callbook');
 			$this->load->model('logbook_model');
 			$lookupcall = $this->callbook->get_plaincall($raw_input['callsign']);


application/controllers/Api.php
```diff
@@ -843,6 +843,17 @@ function private_lookup() {
 			$user_id=$this->session->userdata('user_id');
 		}
 
+		if (($raw_input['qrz'] ?? '' == 'true') && (($raw_input['callsign'] ?? '') != '')) {
```
phl0:
```suggestion
		if (($raw_input['callbook'] ?? '' == 'true') && (($raw_input['callsign'] ?? '') != '')) {
```

phl0: I would rather name the parameter "callbook" or similar because it is (not only) qrz.com which is queried. Depends on user configuration. Can even be unpredictable with multiple callbooks configured (which works as I just tested. callbook->source parameter in results tells which lookup was successful).
Apart from that the code works fine for me. Tested several callsigns.

int2001: done

PR #2785, opened 2026-01-07
Introduce (optional) Ratelimiting on API

int2001:
Introducing Ratelimiting on APIs.

Reason: WL is getting more huge. Currently there's nearly no way to protect the APIs (only coarse via reverse-proxy) from being DoSd

see config.sample.php how to enable it. Could be enabled for single endpoints with different thresholds.
It checks the amount of calls made with a specific api_key. if there's no api-key it takes the session_key, if there's no session_key it takes the IP. Most of our API-Endpoints use the key.

if no `api_rate_limits`-key is present in `config.php`, WL behaves as b4 (no ratelimiting... :) )

The threshold is checked in a sliding window. Once exceeded the caller receives an error "Exceeded - try again in nn seconds".

Testing:
- add rate_limit keys to config.php (as in config.sample.php)
- Call endpoints. Stress-testing example: `seq 1 100 | xargs -n1 -P10 -I{} curl -X POST -H "Content-Type: application/json" -d '{"key":"[WL_KEY]","callsign":"DJ7NT"}' https://[WL_URL]/index.php/api/private_lookup` - will call private_lookup 100 times

### DIFF
--- a/application/controllers/Api.php
+++ b/application/controllers/Api.php
@@ -42,7 +42,7 @@ class API extends CI_Controller {
 		$result = $this->rate_limit->check($endpoint, $identifier);
 
 		if (!$result['allowed']) {
-			log_message("Debug","Rate limit for ".$endpoint." and ID:".($identifier ?? '')." exceeded");
+			log_message("Debug","Rate limit for endpoint ".$endpoint." and ID: ".($identifier ?? '')." exceeded");
 			$this->rate_limit->send_limit_exceeded_response($result['retry_after']);
 			return false;
 		}


application/controllers/Api.php
```diff
@@ -28,6 +28,27 @@ function help() {
...
+	 */
+	protected function check_rate_limit($endpoint, $identifier = null) {
+		if (!$this->load->is_loaded('rate_limit')) {
+			$this->load->library('rate_limit');
+		}
+
+		$result = $this->rate_limit->check($endpoint, $identifier);
+
+		if (!$result['allowed']) {
+			log_message("Debug","Rate limit for ".$endpoint." and ID:".($identifier ?? '')." exceeded");
```
phl0:
```suggestion
			log_message("Debug","Rate limit for endpoint ".$endpoint." and ID: ".($identifier ?? '')." exceeded");
```

PR #2786, opened 2026-01-07
feat: prevent user select lotw badge when copy

tallcode:
When I try to copy contact information to a friend, or copy an entire row to Excel, an extra "L" always appears after the callsign. This is because the callsign and the LoTW badge are in the same cell. Adding this style prevents the user from selecting the LoTW badge.

Like this.
image
2026-01-07	05:26	VK9DXL	FT8	+01	+02	15m	Norfolk Island


phl0 (approved):

phl0:
Thank you for the contribution. I extended the code to also cover the results in Logbook advanced. Code has been merged and will be available with the next release.

PR #2791, opened 2026-01-07
Remove unneeded LoTW stuff from the code

phl0:
The export function is not linked/referenced anywhere so probably not needed.


int2001:
PS: perhaps reduce/remove the webadif-controller too?

phl0:
> PS: perhaps reduce/remove the webadif-controller too?

Nothing stale found in webadif controller. Those functions are in use for uploads to QO-100 Dx Club. So not touching that controller.

Issue #2804, opened 2026-01-12
AO-123 QSOs are not uploaded to LoTW


ea5wa:
### I have checked the following

- [x] I am running the latest version of Wavelog
- [x] I have cleared my browser cache
- [x] This is a bug, not a setup question or feature request

### Bug Description

After trying to upload to LoTW my last QSOs, status is shown in gray. Any error message has been displayed.

Previously, a QSO has been recorded in RS-44 sat without any problem to be uploaded.

<img width="1450" height="512" alt="Image" src="https://github.com/user-attachments/assets/1b64a99e-f655-49e5-9038-34ea4a8a6f5d" />

### Steps to Reproduce

1. Select AO-123 as satellite
2. Record a QSO
3. Upload this QSO to LoTW

### Expected Behavior

Should be Up arrow displayed in green

### Wavelog Version

2.2.2

### PHP Version

8.4.15

### MySQL/MariaDB Version

8.4.7

### Server OS

Windows 11

### Browser

Edge

### Device Type

Desktop

phl0:
Cannot confirm a bug here. Can you please show the QSO details? Is the PROP_MODE set to SAT?

ea5wa:
Yes, Propagation Mode was SAT

<img width="1543" height="1330" alt="Image" src="https://github.com/user-attachments/assets/b9627494-dcf8-4acc-83c8-0e27c1e6e20c" />

<img width="1028" height="910" alt="Image" src="https://github.com/user-attachments/assets/fac0651b-6d4d-4db7-a9da-91957a3d17f9" />

phl0:
Hhmmm that looks about correct. Can you please check the list of satellites

ea5wa:
I syncronized the sat list after uploading these QSOs.

I'm going to re-enter these QSOs now and check if it works

phl0:
Just set the LoTW sent/received state back to "No" and sync LoTW. That should work. No need to re-enter the QSOs.

ea5wa:
Now, it's OK

<img width="2328" height="413" alt="Image" src="https://github.com/user-attachments/assets/080bee14-6946-4ab4-95ec-6accd6863ea3" />

Maybe some message "Sat not included" or something like this is needed

phl0:
We made the code check against valid satellites for LoTW. So the list has to be synced first. Upload with invalid SAT names is useless and prevented by setting LoTW sent/rcvd to "Ingnore". So works as designed. Closing here.

PR #2809, opened 2026-01-13
Show info about SAT support by LoTW

phl0:
As per https://github.com/wavelog/wavelog/issues/2804 we should show info if entered satellite is supported:

image

Test:

- Enter SAT name :)

### DIFF
--- a/assets/js/sections/flightpath.js
+++ b/assets/js/sections/flightpath.js
@@ -719,7 +719,7 @@ function plot_sat() {
 	}
 
 	$.ajax({
-		url: base_url + 'index.php/satellite/get_tle',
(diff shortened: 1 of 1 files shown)


application/controllers/Satellite.php
```diff
@@ -227,16 +227,25 @@ public function create_ics($raw_sat,$raw_aos,$raw_los) {
 		$this->load->view('satellite/schedule',$data);
 	}
 
-	public function get_tle() {
+	public function get_sat_info() {
```
int2001, line 230: breaks flightpath-page (see ./assests/js/sections/flightpath.js )

PR #2810, opened 2026-01-14
Order labels by main-prefix instead of ADIF-DXCC-Number

int2001:
Should fix #2807


AndreasK79 (approved):

AndreasK79:
This should hopefully fix the sorting issue here.

PR #2812, opened 2026-01-15
Add callsign to gridmap filter

phl0:
Add callsign as optional filter for gridmap:

image


AndreasK79:
@phl0 what about putting the searched callsign in the legend?

phl0:
> @phl0 what about putting the searched callsign in the legend?

Good idea. Will have a look how this can be achieved.

phl0:
Callsign has been added to legend. Also fixed translation for "fields"

PR #2820, opened 2026-01-17
[DBTools] Added a map to show map for incorrect grids

AndreasK79:
Go to LBA and click on dbtools.
Run invalid grids.
In the table, you will get a link to a map that shows the incorrect gridsquare, and the correct gridsquares for the dxcc.


phl0:
This works nicely on my dev system. Gosh I have many wrong grids in the log ... :D

AndreasK79:
@phl0 that is exactly why I made the tool in the first place.

PR #2821, opened 2026-01-17
Handle (new) satellites without LoTW name correctly

phl0:
Working with satellites which do not have a LoTW name (yet) causes a few bugs. Satellite path simulator works by accident as long as there is only one satellite without LoTW name.
Satellite pass predictor shows empty select dropdown and upon selection and calculation shows passes for **all** satellites instead of selected one.

Screenshot from 2026-01-17 20-38-25

Screenshot from 2026-01-17 20-38-34

Test: Create a satellite (e.g. QMR-KWT-2) without LoTW name and configure transponder and TLEs. Then use it in pass prediction or flightpath

### DIFF
--- a/src/predict/Predict/Sat.php
+++ b/src/predict/Predict/Sat.php
@@ -28,7 +28,6 @@ class Predict_Sat
 
     public $name     = null;
     public $nickname = null;
-    public $displayname = null;
     public $website  = null;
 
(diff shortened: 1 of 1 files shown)


phl0:
Created a new PR to collect all these changes instead of handling them separately. See https://github.com/wavelog/wavelog/pull/2828

PR #2822, opened 2026-01-18
Allow for more than one satellite without LoTW name

phl0:
Adding more than one satellite without LoTW name silently fails because the add function checks if there is a satellite with that (empty) (LoTW) name and if there is skips.

Patch: Check for name **and** displayname and log an error message if a dupe is to be added

Test: Add satellites without LoTW name. First one goes ok. Upon adding the second one nothing happens in the GUI.


phl0:
Created a new PR to collect all these changes instead of handling them separately. See https://github.com/wavelog/wavelog/pull/2828

PR #2825, opened 2026-01-18
add 'd M y' date format

kc9uhi:
add short form 'd M y' date format


HB9HIL:
There is another dependency in the QSO/Contest Logging form. The date there is quite tricky in handling. Did you check this properly working?

phl0:
It seems that contest logging is working. But the display in the QSO table does not match this format:

image

HB9HIL:
> It seems that contest logging is working. But the display in the QSO table does not match this format:

This is okay. If it still works normally we can leave this for the moment. Working on the new feature. More important is that QSO logging is working properly with the new date format

kc9uhi:
PHP and SQL date formats are not directly compatible 1->1. 'M' (short month name) in PHP correlates to 'b' in SQL. Many more mappings are also now fixed.

AndreasK79 (approved):

AndreasK79:
I have clicked around, looking for trouble, so far not found any :)

@kc9uhi great work. Dealing with these date formats have been a real pain!

PR #2826, opened 2026-01-18
change presets to UTC, to match logbook entries

kc9uhi:
Date preset buttons used local timezone, logbook entries are UTC. Fix preset button functions to use UTC to clear mis-match.


phl0 (approved):
We probably didn't notice as we are off UTC by just one hour here :)
Tested the code and didn't spot a problem.

HB9HIL (approved):

PR #2828, opened 2026-01-19
Handle satellites without LoTW name correctly

phl0:
This is a branch to collect patches to handle satellites which are not supported by LoTW (yet) correctly. Currently there are several issues with satellites with empty name column. This affected adding, editing those satellites as well as adding/removing TLE. Also logbook view is showing bogus data.

Includes patches from https://github.com/wavelog/wavelog/pull/2821 and https://github.com/wavelog/wavelog/pull/2822.

Fixes:

- Removed functions re QSO count whose references were removed in https://github.com/wavelog/wavelog/pull/1751
- Clears TLEs on update only for satellites with name / norad id
- Sets SAT_NAME to null for all QSOs where SAT_NAME is '' to distinguish from sats without LoTW name (this prevents bogus DB queries with left joins on SAT_NAME
- Sets SAT_NAME to null upon logging/editing if SAT_NAME is empty
- Checks for dupe names upon adding a SAT

### DIFF
--- a/application/views/view_log/qso.php
+++ b/application/views/view_log/qso.php
@@ -302,7 +302,7 @@
                         <td><?= __("Satellite Name"); ?></td>
                         <td><a href="https://db.satnogs.org/search/?q=<?php echo $row->COL_SAT_NAME; ?>" target="_blank">
                         <?php if ($row->sat_displayname != null) {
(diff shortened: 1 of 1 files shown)


phl0:
Added two minor UI fixes. Merge time!

PR #2837, opened 2026-01-22
[DXCC ID] All checking now goes via the class

AndreasK79:
Refactored code to go via the DXCC class for any DXCC ID.
Better with one place.

Removed functions from logbook_model. Put them in the calltester controller, since I use that for testing purposes.

This is a bit tough to test, but I'll try to list up what is affected:
- QSO entry - dxcc id here
- Adif import - dxcc id (unless use dxcc from adif is checked)
- DX Calendar - dxcc id
- dxcluster/call - dxcc id - not sure where this is used
- dxclustercache/invalidateForCallsign - dxcc id


application/libraries/DxclusterCache.php
```diff
@@ -76,8 +80,8 @@ public function invalidateForCallsign($callsign) {
 		$this->deleteFile($this->getWorkedCallKey($logbook_key, $callsign));
 
 		// Look up DXCC and continent from callsign
-		$this->CI->load->model('logbook_model');
-		$dxcc_info = $this->CI->logbook_model->dxcc_lookup($callsign, date('Y-m-d'));
+		$dxccobj = new Dxcc($date);
```
int2001: `$date` isn't previously defined

AndreasK79: And that is why we check and test. Fixed now, thanks.

PR #2838, opened 2026-01-22
Show 0km instead of blank (same as QSO view)

phl0:
QSO view show 0km as distance if grid is same. LBA and logbook view show an empty string. Adjusted LBA and logbook view to be the same as QSO view.


phl0:
Also fixes JS failures if grid of source and destination is the same and Qrbcalc view is called:

image

PR #2845, opened 2026-01-24
Radical Cleanup of create_qso()

HB9HIL:
In preparation for the new contest module I need a bit of a clean up and post inputs shouldn't be in a model :-)

### DIFF
--- a/application/controllers/Contestcalendar.php
+++ b/application/controllers/Contestcalendar.php
@@ -139,7 +139,6 @@ class Contestcalendar extends CI_Controller {
 			curl_setopt($ch, CURLOPT_RETURNTRANSFER, true);
 			curl_setopt($ch, CURLOPT_CONNECTTIMEOUT, 10);
 			$rssRawData = curl_exec($ch);
-			curl_close($ch);
 
 			if ($rssRawData === FALSE) {
 				$msg = "Something went wrong with fetching the Contest Data";
@@ -148,8 +147,6 @@ class Contestcalendar extends CI_Controller {
 			}
 
 			$this->cache->save('RssRawContestCal', $rssRawData, (60 * 60 * 12)); // 12 hours cache time
-
-			curl_close($ch);
 		}
 
 		return $rssRawData;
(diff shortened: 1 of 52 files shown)


HB9HIL:
@int2001 For your websocket return nothing should have changed. See commit f740b188c81602132d569e6853038c80c118ed90

application/controllers/Qso.php
```diff
@@ -192,21 +196,88 @@ public function index() {
...
 	/*
 	 * This is used for contest-logging and the ajax-call
+	 * Returns JSON
 	 */
 	public function saveqso() {
 		$this->load->model('logbook_model');
-		$this->logbook_model->create_qso();
+
+		$qso_data = [
+			'manual' => $this->input->post('manual', TRUE),
```
int2001:
```suggestion
			'manual' => $this->input->get('manual', TRUE),
```

application/controllers/Qso.php
```diff
@@ -181,7 +181,11 @@ public function index() {
 			// Add QSO
 			// $this->logbook_model->add();
 			//change to create_qso function as add and create_qso duplicate functionality
-			$adif = $this->logbook_model->create_qso();
+			$saveresult = json_decode($this->saveqso(), true);
+
+			// Clear POST data to prevent re-submission on page reload
```
int2001, line 186: we're saving via ajax. sure it's needed?

HB9HIL: Yes because our ajax is not as it should be a different endpoint. The endpoint is the same and I was able to reproduce the issue that QSOs were save multiple times on page reload.

HB9HIL: For the new contesting I mainly need this PR. in another step we can change the ajax in qso.js from "ajax myself with post" to ajax qso/saveqso with POST data.

HB9HIL:
Needs more testing. Seems like there are some warnings on band. redraft

HB9HIL:
@int2001 I added a check to make sure a band exists. Please recheck when you have time.

Issue #2847, opened 2026-01-24
Database table collation mismatch


kc9uhi:
### I have checked the following

- [x] I am running the latest version of Wavelog
- [x] I have cleared my browser cache
- [x] This is a bug, not a setup question or feature request

### Bug Description

Tables added to DB during install follow the database server defaults, rather than value set in database.php (char_set and dbcollat).  In MySQL 8, the default is utf8mb4_0900_ai_ci. Wavelog commonly uses utf8mb4_general_ci, although some table creation statements (dxcc_master, themes) result in other collations (utf8mb3).

This becomes a problem when using DBTools to check gridsquares that do not match DXCC. MySQL throws a collation mismatch error.  Changing vuccgrids.gridsquare to utf8mb4_general_ci resolves the issue.

### Steps to Reproduce

Advanced Logbook -> Database Tools
click 'Check Gridsquares'

### Expected Behavior

Gridsquare check results returns successful query, not Internal Server Error

### Wavelog Version

2.2.2

### PHP Version

8.1.2

### MySQL/MariaDB Version

8.0.44

### Server OS

Ubuntu 22.04.5

### Browser

Chrome

### Device Type

Desktop

kc9uhi:
Fixed by pull #2848

PR #2848, opened 2026-01-24
fix db collation, set default

kc9uhi:
Fixes issue #2847

Previous database table creation statements are rather sloppy, don't follow the database.php settings, defaulting to database server settings. Multiple different collations in use throughout the wavelog DB, resulting in comparison errors.

Migration 084 addresses this, but more tables have been created without following the database.php setting. This migration normalizes remaining unfixed tables, and sets the DB default to database.php setting, hopefully preventing future collation issues.

### DIFF
--- a/application/migrations/270_fix_dbcollate_again.php
+++ b/application/migrations/270_fix_dbcollate_again.php
@@ -14,14 +14,22 @@ class Migration_fix_dbcollate_again extends CI_Migration
 			'vuccgrids'
 		);
 		foreach ($tables as $table) {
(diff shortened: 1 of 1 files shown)


HB9HIL:
Hello @kc9uhi 

Looks good. Test on a default fresh installation. However, as you might already noticed the world is full of strange installs. Modified installations by niphty users, Migrated Cloudlog Installation to Wavelog and so on 

Therefore we like to make things more failsafe. Please consider using the dbtry pattern for the queries we already used for example here:

https://github.com/wavelog/wavelog/blob/9ad49e2893f89b62672f9d033531eebf7fa5cf7e/application/migrations/267_add_lotw_cert_serial.php#L18-L24

AndreasK79:
@kc9uhi nice find. My dev db is not picky at all, so I don't pick up on these issues.

PR #2860, opened 2026-01-26
Skip callbook lookups if credentials are empty/not set

phl0:
Callbook lookups should only be done if credentials are indeed set. 

Test: Perform lookup for a call without credentials for callbooks being defined in config.php. Testscenario:

- config: $config['callbook'] = ['qrz', 'hamqth', 'qrzcq', 'qrzru'];
- config: creds for qrzru not set (i.e. config item not defined)
- Lookup VJ5A

Causes 

image

while querying qrzru services.

### DIFF
--- a/application/libraries/Callbook.php
+++ b/application/libraries/Callbook.php
@@ -53,8 +53,10 @@ class Callbook {
 		if (isset($callbook['error']) && $callbook['error'] != '') {
 			if (is_array($source_callbooks)) {
 				foreach ($source_callbooks as $source) {
-					$callbook['error_'.$source] = $callbook_errors['error_'.$source];
-					$callbook['error_'.$source.'_name'] = $callbook_errors['error_'.$source.'_name'];
+					if (isset($callbook_errors['error_'.$source])) {
+						$callbook['error_'.$source] = $callbook_errors['error_'.$source];
+						$callbook['error_'.$source.'_name'] = $callbook_errors['error_'.$source.'_name'];
+					}
 				}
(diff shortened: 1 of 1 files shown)


phl0:
We added some more code to show the error messages returned from the lookup services to give the user a hint what may be wrong instead of just hiding the information. Testing by callsigns which are not known by any of the configured services. E.g. VJ5A:

image

application/libraries/Callbook.php
```diff
@@ -47,38 +50,30 @@ public function getCallbookData($callsign) {
...
 		}
 
+		if (isset($callbook['error']) && $callbook['error'] != '') {
+			if (is_array($source_callbooks)) {
+				foreach ($source_callbooks as $source) {
+					$callbook['error_'.$source] = $callbook_errors['error_'.$source];
+					$callbook['error_'.$source.'_name'] = $callbook_errors['error_'.$source.'_name'];
+				}
+			}
+		}
```
int2001, line 62: Since we already had fuckups because of empty vars, i'd suggest the following here:

```suggestion
if (isset($callbook['error']) && $callbook['error'] != '') {
      if (is_array($source_callbooks)) {
          foreach ($source_callbooks as $source) {
              if (isset($callbook_errors['error_'.$source])) {
                  $callbook['error_'.$source] = $callbook_errors['error_'.$source];
                  $callbook['error_'.$source.'_name'] = $callbook_errors['error_'.$source.'_name'];
              }
          }
      }
  }
```

PR #2878, opened 2026-01-29
feat: broadcast callsign lookup results via WebSocket

TnxQSO-Admin:
This PR implements a global WebSocket broadcast whenever a callsign lookup occurs in the QSO window. It enables external hardware and software (rotators, external displays, and integrators) to react instantly to the station being logged.

Key changes:
- Added 'broadcastLookupResult' to assets/js/cat.js.
- Implemented automatic WebSocket initialization if a local CAT-URL (127.0.0.1/localhost) is detected.
- Triggered broadcast from assets/js/sections/qso.js upon successful lookup, including full metadata (Azimuth, LoTW status, US County, Slot confirmation, etc.).

### DIFF
--- a/assets/js/cat.js
+++ b/assets/js/cat.js
@@ -1286,7 +1286,7 @@ $(document).ready(function() {
             $('#toggleCatTracking').removeClass('btn-success').addClass('btn-secondary');
             // Display offline status when no radio selected
             displayOfflineStatus('no_radio');
-        } else if (selectedRadioId == 'ws') {
+        } else if (selectedRadioId == 'ws' || (websocketEnabled && websocket !== null)) {
             websocketIntentionallyClosed = false; // Reset flag when opening WebSocket
             reconnectAttempts = 0; // Reset reconnect attempts
             hasTriedWsFallback = false; // Reset WSS failover state - try WSS first again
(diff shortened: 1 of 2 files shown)


int2001:
Nice idea. was already on my todo-list - but you were faster ;)

Question about it:
- How should the User decide to use the bearing for rotation? your current implementation always broadcasts the information to the Gate/Client, which means: he has to switch off/on rotation outside wavelog.
- just for interest: do you already have a client/bridge which forwards the information to rotctl?

tbh: i built the same thing as preparation for SAT/Autorotate as proof-of-concept in #2780 - but my implementation is also missing such a switch. i think we need something general here.

Furthermore i discussed exactly this feature with @HB9HIL in our core-team because it could be needed for the contest-module.

open for ideas.

TnxQSO-Admin:
Hi @int2001! Thanks for the positive feedback. It is encouraging to hear that this aligns with the core team's vision for future contest modules and SAT features.

Regarding your questions, here is my take on the architecture based on my implementation:

**1. How should the User decide to use the bearing for rotation?**
I believe in a strong **"Separation of Concerns"**, Wavelog acts as the data provider, broadcasting the *context* (Current DX entity, Bearing, Grid), while the **Client** (Middleware/Gateway) handles the *behavior*. 

If we put the "Rotate ON/OFF" switch inside Wavelog, we clutter the UI. Instead, the downstream client should decide how to act on the data:
* **Scenario A (Contest):** The client might be configured to "Auto-rotate on lookup".
* **Scenario B (Casual DXing):** The client receives the bearing but waits for user confirmation before sending a command to the hardware.

**2. Do you already have a client/bridge which forwards the information?**
Yes! My team has developed **Wave-Flex Integrator**, a bridge tool specifically for FlexRadio users. We have developed a module called **"QSO Assistant"** that consumes this exact WebSocket broadcast.

Instead of blindly rotating the antenna immediately (which might be dangerous depending on the setup), the QSO Assistant populates a floating dashboard with the data. 
* It calculates **Short Path** and **Long Path** bearings.
* It presents two buttons: `ROTATE SP` and `ROTATE LP` showing the target degrees.
* The user makes the final decision to turn the beam by clicking the corresponding button.

QSO-Manager

It also uses the metadata (Grid, Callsign) to optionally display the operator's profile image and map on a secondary screen, and allows injecting the callsign directly onto the FlexRadio panadapter as a spot.

QSO-Manager II

**Regarding Rotctld/Hamlib:**
While Wave-Flex Integrator currently uses MQTT for rotator control (common for Remote QTH interfaces), writing a small Python or Node bridge that listens to this WebSocket and forwards commands to `rotctld` (port 4533) would be trivial. By keeping the Wavelog side generic, just broadcasting the current lookup data, we enable all these use cases, Rotators, External Matrix Displays, LED Maps, etc., without Wavelog needing to know the specifics of the hardware.

I hope this clarifies the design choice. I'm happy to adjust the PR if you prefer a different approach, but "Broadcast Always, Decide Locally" seems to offer the most flexibility for the community.

int2001:
Hmm. sounds fair and legit.
i think it's okay for now. But i am still thinking about enhancing the websocket-payload to some boolean-flag "AutoRotate".
this shouldn't break your logic.

however. PR looks fine and okay to me. Maybe @HB9HIL could also have a short look into it and approve it, before i put it into the dev-branch.

TnxQSO-Admin:
Great news! Thank you both @int2001 and @HB9HIL for the review and the approvals.

Regarding the `AutoRotate` flag: I think that is an excellent idea for a future enhancement. It would provide a clean way to pass user preferences from the Wavelog UI down to the clients/bridges without breaking the existing logic.

I'm excited to see this in the dev branch. I’ll keep an eye out for any "screams," but I suspect the FlexRadio and automation community will be very happy with this new capability! 🚀

int2001:
@TnxQSO-Admin please have a look at the auto-connect-feature for Websocket. Not good... https://github.com/wavelog/wavelog/commit/a937288042bbc441888cd94ac67045a2cafc5837#diff-ffb3d761aa9f432be37e4a62576af63d2f4d6edfedaa83d65081b3b19536c5fdR1315

PR #2879, opened 2026-01-30
Disable Auto-Websocket

int2001:
I am sure that the “Automatic connection to Websocket” feature #2878 was implemented with the best of intentions.
However, not everyone uses Websocket clients. There are many people who do not have a Websocket client.

And the feature keeps trying to connect to the Websocket, even if “None” is selected as an option.

Give people the option to choose whether they want to use “Websocket” or not.

### DIFF
--- a/assets/js/cat.js
+++ b/assets/js/cat.js
@@ -1312,20 +1312,6 @@ $(document).ready(function() {
             // Start standard polling
             CATInterval = setInterval(updateFromCAT, CAT_CONFIG.POLL_INTERVAL);
 
-            // --- PR ADDITION: Auto-enable WebSocket for local CAT URLs ---
-            // Fetch radio details to check if we should also start WebSocket
-            $.getJSON(base_url + 'index.php/radio/json/' + selectedRadioId, function(data) {
-                if (data.cat_url) {
-                    const url = data.cat_url.toLowerCase();
-                    if (url.includes('127.0.0.1') || url.includes('localhost')) {
-                        console.log("CAT: Local CAT URL detected (" + data.cat_url + "). Initializing WebSocket...");
-                        websocketIntentionallyClosed = false;
-                        initializeWebSocketConnection();
-                    }
-                }
-            });
-            // -------------------------------------------------------------
-
             if ((window.CAT_COMPACT_MODE === 'ultra-compact' || window.CAT_COMPACT_MODE === 'icon-only') && typeof window.isCatTrackingEnabled !== 'undefined' && !window.isCatTrackingEnabled) {
                 displayOfflineStatus('cat_disabled');
             }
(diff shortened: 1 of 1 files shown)


HB9HIL (approved):

TnxQSO-Admin:
Hi @int2001! I completely understand your concern. The auto-connect logic was indeed too aggressive by running even when "None" (ID 0) was selected, causing unnecessary console noise. Thanks for catching that!

However, reverting strictly to `selectedRadioId == 'ws'` creates a major hurdle for integrated radios like **Wave-Flex Integrator (WFI)**. 

WFI registers itself via the API (obtaining a numeric database ID), but it simultaneously relies on the WebSocket connection to receive lookup metadata for local station automation (like rotators and panadapter displays).

If we lock WebSockets to only the `'ws'` radio type, users are forced to choose:
1. Use the **API Radio** for frequency/mode sync, but lose all WebSocket metadata features.
2. Use the **Websocket Radio**, but lose the API-based frequency sync from the integrator.

**Proposal:**
Could we allow WebSocket initialization for API-based radios if they have a local CAT-URL configured? To avoid the spam you found, we could move the check inside the logic that only runs when a radio is actually selected (`selectedRadioId != '0'`).

I'm happy to contribute a PR that enables this "hybrid" behavior without the side effects!

int2001:
The local CAT-URL is the default for every CAT-Use. So if one uses one of the external tools (there are a lot out there which are only using the api/Radio POST roundtrip instead of the socket), he would be nagged by ws-tries. 

Tbh: localhost was hardcoded a long time ago. I only implemented it to be configurable on demand of one OM. So 99,9% use localhost.

Localhost doesn't mean there's websocket!

I could live with one try to establish a ws-connection if the user has chosen a radio, but not like it is at the moment.

If that's a compromise for you, go ahead and adjust it like that.

Deal?

TnxQSO-Admin:
> The local CAT-URL is the default for every CAT-Use. So if one uses one of the external tools (there are a lot out there which are only using the api/Radio POST roundtrip instead of the socket), he would be nagged by ws-tries.
> 
> Tbh: localhost was hardcoded a long time ago. I only implemented it to be configurable on demand of one OM. So 99,9% use localhost.
> 
> Localhost doesn't mean there's websocket!
> 
> I could live with one try to establish a ws-connection if the user has chosen a radio, but not like it is at the moment.
> 
> If that's a compromise for you, go ahead and adjust it like that.
> 
> Deal?

Hi @int2001!

After analyzing the WaveLogGate architecture, I realized that I was trying to solve the problem from the wrong angle.

Instead of patching Wavelog to force WebSocket connections for API-based radios (the specific logic you are currently reverting), I will update Wave-Flex Integrator to act as a proper WebSocket Server, emulating the WaveLogGate protocol on port 54322.

This allows users to simply select **"Live - WebSocket"** in the radio dropdown. This is a much cleaner solution as it provides:
1. Real-time frequency updates (no polling lag).
2. An active WebSocket channel for the metadata broadcasts (lookup results) we just implemented, without needing any specialized "hybrid" logic in the Wavelog core.
3. Full compatibility with Wavelog's standard workflow.

So, I withdraw my proposal about the auto-detect logic for API radios. Please proceed with your cleanup/revert of that specific part. WFI will adapt to the standard "Live" implementation instead.

Thanks for the patience and for pointing me in the right direction!

PR #2880, opened 2026-01-31
make sure $call is always uppercase to make regex pattern work

HB9HIL:
If the $call is lowercase the whole dxcc class does not work properly due to the case sensitive regex. Instead fixing all kind of regex we just can make sure the $call is always uppercase


int2001 (approved):
seems legit ;)

PR #2881, opened 2026-01-31
Tweaked dxcc class

AndreasK79:
Speeding up the DXCC lookup class

### DIFF
--- a/application/models/Dxcc.php
+++ b/application/models/Dxcc.php
@@ -1,10 +1,19 @@
 <?php
 
+use Wavelog\Dxcc\Dxcc_id;
+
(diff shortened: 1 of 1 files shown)


application/models/Dxcc.php
```diff
@@ -1,11 +1,20 @@
 <?php
 
+use Wavelog\Dxcc\Dxcc_id;
+
+require_once APPPATH . '../src/Dxcc/Dxcc_id.php';
```
HB9HIL, line 5: we have to rename src/Dxcc/Dxcc.php to src/Dxcc/Dxcc_id.php

HB9HIL (changes requested):

PR #2884, opened 2026-01-31
Improvements and fixes for import functions

HB9HIL:
Two main changes

- only load the stations model if not already loaded
- if in the record a dxcc is set we query the database which contains everything we need. so we can use this data to complete records if they got no cqz or cont


HB9HIL:
I think this is good to go fore the moment. There are much more things which can improved but once at a time :-)

AndreasK79 (approved):

AndreasK79:
Works when I tested yesterday.

PR #2885, opened 2026-02-01
implement config option to directly log into a club

HB9HIL:
There are use cases where this is the better option. Will be documented in the Wiki after merge. Does not affect current behaviour.


int2001:
Is the PR complete?
i am missing something which can "set" that parameter (login-form?)

Edit: Oh, it's a config-item.
How can i test it? What is it exactly doing?

int2001 (approved):

PR #2887, opened 2026-02-01
Prevent livelogging into POST-QSO

int2001:
Scenario:

you're POST-Logging a QSO from yesterday
at the same time you've DXCluster open and you see a wanted station.
you click on it and work the station. Yeah! Now you save the form.

problem: it will be saved POST-QSO with the wrong date/time.

this PR fixes the behaviour. Click on a spot now ALWAYS makes sure that it is prepared in live-logging window.

technically:
- POST-QSO won't answer any longer on a "ping" ("Is there a open QSO-Window")
- POST-QSO won't listen for Calls which are propagated by the cluster

### DIFF
--- a/assets/js/sections/qso.js
+++ b/assets/js/sections/qso.js
@@ -738,12 +738,14 @@ function save_fav() {
 }
 
 
-var bc_bandmap = new BroadcastChannel('qso_window');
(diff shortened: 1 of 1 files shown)


HB9HIL (approved):
confirm fixes this issue

PR #2593, opened 2025-11-24
[Backend] Redis cache for wavelog

szporwolik:
# Redis Caching for DXCluster

## Problem

DXCluster is currently not the quickest — per-user caching significantly increases load times.  
Both options enable file-based caching for DX Cluster:

```php
$config['enable_dxcluster_file_cache_band'] = true;
$config['enable_dxcluster_file_cache_worked'] = true;
```

However, this comes at a cost:  
- Thousands of cache files may be created.  
- Performance and scalability on larger instances remain unknown.

---

## Solution

This PR introduces **Redis caching** as an alternative to file-based caching.  
The goal is to evaluate performance improvements on larger DXCluster instances.

---

## How to Enable Redis Caching

1. **Update `config.php`**
2. **Create or update `redis.php`**
3. Set both DXCluster cache options to `true`:
   ```php
   $config['enable_dxcluster_file_cache_band'] = true;
   $config['enable_dxcluster_file_cache_worked'] = true;
   ```
4. Set the cache adapter to use Redis:
   ```php
   $config['cache_adapter'] = true;
   ```
5. Provide proper Redis server configuration in **`Redis.php`**:
   ```php
   $config['redis'] = [
       'host' => '127.0.0.1',
       'port' => 6379,
       'password' => '',
       'timeout' => 1.5,
   ];
   ```

---

## Dev Notes

- This is a **draft PR**, intended for initial testing and discussion.  
- Focus areas:
  - Performance and stability on large-scale instances  
  - Potential impact on system load and response times  
  - Overall usefulness of Redis caching for the project  

---

## Next Steps

- Gather performance data from test instances  
- Compare Redis-based caching vs file-based caching  
- **Decide whether to merge Redis support into the dev branch**

PR #2888, opened 2026-02-01
Support multiple caches

HB9HIL:
# Caching Support for different cache systems
This PR replaces #2593 which brings the final support for various different caching engines. As the previous PR only targeted redis it was not suitable for the majority of people and did not match with the existing codebase. But caching is nice and brings performance where needed. 

### `file`
The `file` based caching is the default as was already used in various places in Wavelog. Nothing changed here. 

>[!IMPORTANT]
> There might some places where Wavelog forces the file adapter instead using the configured one. This happens when caching for large files like images or comparable data is needed. So even when you configure another adapter here, Wavelog might still use file caching in some places and respects the configured cache path for that.

### `redis`
Redis or it's fork [Valkey](https://github.com/valkey-io/valkey) are the more common cache engines. Wavelog now supports redis as key-value cache. To make redis work you need:
- Obviously a redis or valkey server running
- the PHP extension for redis (`php-redis`; works with valkey aswell)
- you need to set the `$config['cache_adapter']` to `redis`
- you need to copy `redis.sample.php` to `redis.php` and configure your settings as needed

### `memcached`
One of the older ones but still fast and battle tested. Wavelog supports you to configure multiple memcached server with their configured weight for load distribution. To make memcached work you need:
- Obviously a memcached server running
- the PHP extension for memcached (`php-memcached`)
- you need to set the `$config['cache_adapter']` to `memcached`
- you need to copy `memcached.sample.php` to `memcached.php` and configure your settings servers as needed

Since Wavelog relies on Codeigniter 3, which is a pretty old framework there were still support for `apc`. `apc` doesn't work with PHP7/8 anymore and even doesn't exist anymore. But `apcu` still exists and is still reliable and fast. To make `apcu` work you need:
- the PHP extenstion for apcu (`php-apcu`)
- you need to set the `$config['cache_adapter']` to `apcu`
The rest works out of the box. Since this is native PHP caching (with opcache and apcu is just somthink like the user-data storage layer) there is no CLI tool to get the stored values. To still test it you can use this little[ test tutorial](https://gist.github.com/HB9HIL/0cb771d41e6672be1f929f9a4d1b2fd0).

~~### `wincache`
To be honest. I don't have a Windows machine and I never used Wavelog on Windows. So I have no idea if this still works and I don't want to test it. If someone is willing to test, be my guest. This was/is still part of the Codeigniter 3 Core and I didn't remove it. So technically there is support for wincache, but yeah.. This is out of scope for me :laughing:~~ 
UPDATE: Since Wavelog is not meant to run on Windows Server (even it's possible) we drop the support for wincache. other options also work under windows so this is causing more issues then it's worth it.

## Last Words
Always remember. You are hosting a web application. You need to figure out yourself how to configure Redis/Valkey, Memcached or APCu (or even this mystery Wincache. The only cache adapter which is redulary tested by us is the `file` adapter as this is the default. This does not meant that you can't tell us if you find any issues :wink:

### DIFF
--- a/application/models/Debug_model.php
+++ b/application/models/Debug_model.php
@@ -266,28 +266,6 @@ class Debug_model extends CI_Model
 			'key_prefix' => $cache_key_prefix
 		]);
 
-        if ($cache_adapter === 'file') {
-            $cache_path = $this->config->item('cache_path') ?: 'application/cache';
-            $cache_path = realpath(APPPATH . '../') . '/' . $cache_path;
-
-            if (!is_dir($cache_path)) {
-                return false;
-            }
-
-            $files = new RecursiveIteratorIterator(
-                new RecursiveDirectoryIterator($cache_path, RecursiveDirectoryIterator::SKIP_DOTS),
-                RecursiveIteratorIterator::CHILD_FIRST
-            );
-
-            foreach ($files as $file) {
-                if ($file->isFile() && !in_array($file->getFilename(), ['index.html', '.htaccess'])) {
-                    @unlink($file->getRealPath());
-                }
-            }
-
-            return true;
-        }
-
         if (method_exists($this->cache, 'clean')) {
             return $this->cache->clean();
         }


.gitignore
```diff
@@ -1,7 +1,11 @@
 /application/config/database.php
 /application/config/config.php
+/application/config/redis.php
+/application/config/memcached.php
 /application/config/*/config.php
 /application/config/*/database.php
+/application/config/*/redis.php
```
int2001: Was a "bug" before. the `*`-pattern in gitignore doesn't catch subdirs. `**` is needed. perhaps fix the other occurences as well?

```suggestion
/application/config/**/redis.php
```

int2001:
Started to test. this looks good so far. but need more tests

AndreasK79:
I can only report that debug view says that it works on xampp. So if that is good enough, go ahead and merge.

PR #2891, opened 2026-02-02
Install php-redis-extension

int2001:
Since CI supports REDIS as Cache for sessions (coming more),
we should add at least the php-redis extension to the dockerfile.

So one is able to use his existing redis-installation for Wavelog.

Later we can add valkey itself to the compose-example and configs, when using docker.


HB9HIL:
I added the extensions for memcached, apcu and wincache aswell. They are small and and since we support it we should pack them aswell.

HB9HIL (approved):
builds sucessfully

PR #2896, opened 2026-02-02
Dxclustercache fix since we made various caching drivers available

HB9HIL:
Depends on #2895


int2001 (approved):
Tested:
Cluster with REDIS enabled.
with `$config['enable_dxcluster_file_cache_worked'] = true;`
and `$config['enable_dxcluster_file_cache_worked'] = false;`

Garbage Collection works like intended. All key/values are shown as they should.

PR #2897, opened 2026-02-02
Satellite Deprecation fixes

HB9HIL:
This arrays throw some deprecation warnings. Nothing fancy. I fixed it by setting the arrays to an empty string if null. Don't know if the expected behaviour is still given, why @phl0 should check this one day. It's not that urgent. In fact this super nested nested nested array looks strange :laughing:


int2001 (approved):
works. more safety with the fallback to `''`

PR #2900, opened 2026-02-03
Optimize index-usage

int2001:
We already have an index at Logbook-Table which is a combined one of station_id and col_time_on.

but:
a lot of queries which use `DATE(COL_TIME_ON)` on the left side of the where. this leads into the fact, that the potential of index isn't used completly.

this PR refactors all occurences of `DATE(COL_TIME_ON)` within code to use the index.


HB9HIL:
nice! I stumble across that aswell but was busy with other parts. Will test later

HB9HIL (approved):
tested working

PR #2911, opened 2026-02-05
join sat-table in PHP instead of DB. Huge performance booster

int2001:


HB9HIL (approved):
Works nice and gives more performance :-)

AndreasK79 (approved):

PR #2915, opened 2026-02-06
Security fix / Dashbaord

int2001:
radio_display_component wasn't properly protected.
added auth in constructor for whole controller.


HB9HIL (approved):
good find

Issue #2905, opened 2026-02-04
The map markers and the table do not match.


tallcode:
I've observed a data discrepancy between the map visualization and the data table. While paginating through the logbook, the map markers become misaligned with the table rows. This issue is triggered when a WHERE clause is added to the SQL query; it appears to break the offset logic, causing the map to display data from a different range than the filtered grid.

_Originally posted by @phl0 in https://github.com/wavelog/wavelog/issues/2465#issuecomment-3460959319_

int2001:
confirmed. there's something wrong

int2001:
Resolved in #2917

PR #2917, opened 2026-02-06
Fix pagination/map

int2001:
This addresses the issue raised in #2905 

When having QSOs without grid/location-information, the pagination at map ran out of sync to the list.
This one fixes that behaviour.

Background:
SQL is now same, QSOs without location/grid-information are dropped after selecting the QSOs for the map.


phl0 (approved):
This works for me.

PR #2920, opened 2026-02-06
Fix empty mode dropdown if mode from cat is disabled in usermodes

### LINKED ISSUE #2918: Live/man logging: Mode empty if mode set by wlgate (CAT) and mode inactive
### I have checked the following

- [x] I am running the latest version of Wavelog
- [x] I have cleared my browser cache
- [x] This is a bug, not a setup question or feature request

### Bug Description

If CAT control is active via wlgate/flrig and radio is set to USB or LSB and these specific modes are set as inactive in wavelog, the respective QSO field stays empty when live or manual logging. 

background: user wants to log everything as SSB and not differentiate between USB/LSB.

### Steps to Reproduce

log with wlgate as radio and set mode on radio which is inactive in wavelog

### Expected Behavior

 if possible: map to SSB if radio says USB/LSB and these are inactive.

### Wavelog Version

2.2.2

### PHP Version

8.2.29

### MySQL/MariaDB Version

10.11.14-MariaDB-0+deb12u2

### Server OS

Apache/2.4.65 (Debian)

### Browser

Chrome

### Device Type

Desktop

### Screenshots / Browser Console

Image

### Additional Context

_No response_

HB9HIL:
fix #2918


HB9HIL:
this needs a refactoring

PR #2942, opened 2026-02-13
Add SAT to timeline results for satellite awards

phl0:


phl0:
SQL needs rework.

AndreasK79:
DXCC + VUCC works with the new query.

phl0:
Added satellite column to all other awards as well.

AndreasK79 (approved):

PR #2943, opened 2026-02-13
Merge migrations

HB9HIL:
We often talked about it and now we do it. All migrations from 1 to 169 are now merge into the install.sql of the installer. In addition  to that the migration 170 became just an placeholder to keep compatibility for the migration from cloudlog to wavelog. This PR is the first of multiple PR's which improve the installer of Wavelog. This part in particular speeds up the database creation and migration tasks.

I tested a lot but you may should try aswell :-)

### DIFF
--- a/application/migrations/170_welcome_to_wavelog.php
+++ b/application/migrations/170_welcome_to_wavelog.php
@@ -5,15 +5,17 @@ defined('BASEPATH') OR exit('No direct script access allowed');
 /*
 *   Just an empty placeholder. To keep compatibility for the migration from Cloudlog to Wavelog we need an empty migration here as placeholder.
 *   Find the tutorial here: https://github.com/wavelog/Wavelog/wiki/How-to-migrate-Cloudlog-to-Wavelog
+*
+*   During development in February 2026 the migration Tutorial above became deprecated since the code diverged way too much and with this also the database.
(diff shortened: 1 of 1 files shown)


int2001:
Tested at least docker: works so far.

AndreasK79:
Installed and tested on Windows and Xampp. No errors detected, but impossible to test every combination of things.

HB9HIL:
Thinking about possible cases:

1. Existing installations are not **directly** affected by this change. It only means they can not down migrate below mig version 170 and tbh, they shouldn't do that :-D haha
2. New installations have no problem, this PR was tested in docker, plain linux and windoze. So there shouldn't be any issue
3. This is the important case: After merge and release there are two "types of databases": 

- Existing databases which upgraded from an older version to this release
- New installations

The important thing here is if we write new migrations with any db changes they have to work with both types of databases. This is not a problem as long as both database version are in the same state. How I tested this?

I installed a Wavelog instance with the current dev branch and made a dump of the database without any changes
Then I installed a Wavelog instance with this pr branch and made a dump

then I compared both sql dumps. As long as they diff is clean (except date dynamic values like created_at, etc) everything is fine.

AndreasK79:
@HB9HIL that comparison was exactly what I was thinking, and really the only way to be sure :)

PR #2949, opened 2026-02-14
Cache buster fixes

HB9HIL:
Some fixes and additions for #2945

### DIFF
--- a/application/views/components/dxwaterfall.php
+++ b/application/views/components/dxwaterfall.php
@@ -68,7 +68,7 @@
 	</script>
 
 	<!-- DX Waterfall Component - CSS -->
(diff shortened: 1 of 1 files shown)


HB9HIL:
Would be nice if @AndreasK79 can test `dev` branch in his XAMPP environment with and without this PR if the cache_buster works without the filepath verification (backslash handling for windoze servers)


EDIT: Works without the special handling. So no need to include it yet but I added a comment in case issues pop up

AndreasK79:
@HB9HIL for the record, I've seen no trouble with the other PR in the dev branch on my Windows dev machine.

int2001 (approved):

PR #2952, opened 2026-02-15
[Advanced Logbook] Fetch all modes for edit

AndreasK79:
As DK1BOU pointed out, edit doesn't contain all modes in the Advanced Logbook. Instead, only the active ones was fetched. I changed this to fetch all the available modes instead.

The regular QSO edit fetches all the modes, regardless if they are active or not. With this PR, the batch edit does the same.


HB9HIL (approved):
reasonable change

int2001 (approved):
let's keep in mind, that Modes (at Usersettings) ONLY has effect on logging,
not on single-edit, not on bulk-edit.

with that knowledge it's fine for me

PR #2953, opened 2026-02-15
[DXCluster] Choose which column to filter

AndreasK79:
Dropdown for choosing which column to filter:
image

In response to #2946


int2001 (approved):
Tested it. works as designed.

one can consider to save it along favourites - but imho no showstopper / nothing urgent

PR #2954, opened 2026-02-15
add list_clubmembers api endpoint

kc9uhi:
Adds list_clubmembers API endpoint

Accepts API key of a 'club officer' (permission level 9), returns array containing users with permissions for associated club station.

Developed for use with the [k7rst event scheduler](https://github.com/wizardofaz/k7rst_event_scheduler)

### DIFF
--- a/application/controllers/Api.php
+++ b/application/controllers/Api.php
@@ -1314,19 +1314,28 @@ class API extends CI_Controller {
 	* API key needs to be of a club officer (permission level 9)
 	* returns array of club member details
 	*/
-	function list_clubmembers($key = '') {
+	function list_clubmembers() {
 		header('Content-type: application/json');
 
 		$this->load->model('api_model');
-		if ($this->api_model->access($key) == "No Key Found" || $this->api_model->access($key) == "Key Disabled") {
+
+		// Decode JSON and store
+		$obj = json_decode(file_get_contents("php://input"), true);
+		if ($obj === NULL) {
+		    http_response_code(400);
+			echo json_encode(['status' => 'failed', 'reason' => "wrong JSON"]);
+			return;
+		}
+
+		if ($this->api_model->access($obj['key']) == "No Key Found" || $this->api_model->access($obj['key']) == "Key Disabled") {
 			http_response_code(401);
 			echo json_encode(['status' => 'error', 'message' => 'Auth Error, invalid key']);
 			return;
 		}
 
 		$this->load->model('club_model');
-		$userid = $this->api_model->key_userid($key);
-		$created_by = $this->api_model->key_created_by($key);
+		$userid = $this->api_model->key_userid($obj['key']);
+		$created_by = $this->api_model->key_created_by($obj['key']);
 		$club_perm = $this->club_model->get_permission_noui($userid,$created_by);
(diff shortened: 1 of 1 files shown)


application/controllers/Api.php
```diff
@@ -1309,4 +1309,51 @@ private function sanitize_cat_url($url) {
...
+		$userid = $this->api_model->key_userid($key);
+		$created_by = $this->api_model->key_created_by($key);
+		$club_perm = $this->club_model->get_permission_noui($userid,$created_by);
+		if (($userid == $created_by) || (($club_perm ?? 0) != 9)) { // not club officer
+			http_response_code(401);
+			echo json_encode(['status' => 'error', 'message' => 'Auth Error, not enough permissions for this operation']);
+			return;
+		}
+
+		$this->load->model('club_model');
```
HB9HIL:
```suggestion

```
already loaded in line 1327

application/controllers/Api.php
```diff
@@ -1309,4 +1309,49 @@ private function sanitize_cat_url($url) {
...
+			echo json_encode(['status' => 'error', 'message' => 'Auth Error, not enough permissions for this operation']);
+			return;
+		}
+
+		$memberlist = $this->club_model->get_club_members($userid);
+		if (!empty($memberlist)) {
+			foreach($memberlist as $member) {
+				$members[] = [
+					'callsign' => $member->user_callsign,
+					'first_name' => $member->user_firstname,
```
int2001: in some circumstances all data besides the callsign may be treated as sensitive data regarding to GDPR.
i'd appreciate if we could work with the minimal dataset (call and level only ?) here.

What do you think?

Yes i know, this is a strange issue - but it can be a showstopper for huge clubs.

Let's turn the question:
Whats your usecase for all that data?

kc9uhi: Externally I'm checking if a given callsign is a member of the clubstation. Info returned by the API endpoint is the same as displayed on the Clubstation membership page, both requiring a user with at least 'Club Officer' privileges. I included the full data set to match what is already displayed on the UI.

int2001: yes. got that.
but: would you mind to reduce the api-output at least to only the absolute nescessary things (imho Call and p_level)?

a config-switch for "show all attributes" could also help here.

background: some huge clubs must suppress those information because of GDPR. if we'd have a switch for it, we could also extend the (optional) suppresion to the UI. i would implement the evaluation for the switch after your PR has been merged.

what do you think?

kc9uhi: I can reduce API output, that's not a problem. 

Reducing the info on the UI (at least for clubstation permissions) could introduce a new problem of selecting the proper user account to assign permissions to. Currently, user accounts only require a unique email address, callsign can be reused. Not including at least callsign and email creates ambiguity between user accounts.

int2001: hmm. valid point - but the user_name is also unique.
so: user_name + user_callsign + level?

HB9HIL (approved):

int2001:
@kc9uhi tnx for contrib and the compromise :)

Are you going to write a little documentation to the wiki afterwards, pse?

--> https://github.com/wavelog/wavelog/wiki/API

PR #2956, opened 2026-02-15
Dxcluster improvement

HB9HIL:
Improved the `get_batch_spot_statuses()` method which processes the dxcluster spots. The method is not 2-4 times faster then before


AndreasK79:
No errors detected. Spots loads faster, and all the checks I've done, seems correct regarding worked/confirmed.

PR #2957, opened 2026-02-15
Add option to link newly created station location to active logbook

kc9uhi:
Usage scenario: create station location via API, link automatically to active logbook

'link_active_logbook' added to station location data array. Values that cause empty() to return FALSE will trigger this routine.


kc9uhi:
The save_location() and calling import_locations_parse() functions are only called by automated import routes -- the station_create API and stationsetup/list_locations import button.

When a user is manually creating/editing a station location, they're already coming from/returning to the Station Setup page, where logbook association is already handled quite well. An option to link to active logbook at station location creation time could be handy, but at the same time redundant. These actions are also handled by entirely different functions in models/Stations.php

Documentation for this option would be best located in the wiki, correct?

int2001:
yes, wiki is good and okay.

int2001 (approved):

PR #2959, opened 2026-02-16
Automatically enrich qso on import or added via api

salyh:
Add logic and config item $config['always_enrich_qso'] (default: false) to automatically enrich qso upon import or added via api with data from configured callbook(s)

### DIFF
--- a/application/config/config.sample.php
+++ b/application/config/config.sample.php
@@ -29,6 +29,10 @@ $config['callbook'] = 'hamqth';
 
 $config['datadir'] = null; // default to install directory
 
+// If true every added qso will be automatically enriched with data from configured callbook(s)
+// Applies for adif imports as well as added qso via api
+$config['always_enrich_qso'] = false;
+
 /*
 |--------------------------------------------------------------------------
 | Logbook Options
--- a/application/models/Logbook_model.php
+++ b/application/models/Logbook_model.php
@@ -5535,6 +5535,13 @@ class Logbook_model extends CI_Model {
 				'COL_MORSE_KEY_TYPE' => (!empty($record['morse_key_type'])) ? $record['morse_key_type'] : '',
 			);
 
+			if($this->config->item('always_enrich_qso')) {
+				$this->load->model('logbookadvanced_model');
+				$callbook = $this->loadCallBook($record['call'], $this->config->item('use_fullname'));
+				$enrichedQso = $this->logbookadvanced_model->enrichQsoWithCallbookInfo($record, $callbook, true, $station_profile->station_gridsquare);
+				$data = [...$data, ...$enrichedQso];
+			}
+
 			// Collect field information from the station profile table thats required for the QSO.
 			if ($station_id != "0") {
 				if (!(array_key_exists($station_id, $this->station_result))) {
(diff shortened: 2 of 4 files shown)


HB9HIL:
Hello @salyh 

Thank you very much for your contribution and the effort you put into this PR. However this topic was already discussed several time within the team and also in the community. 

For example here: https://github.com/wavelog/wavelog/discussions/2823

We decided to not implement a feature like that because this would mean a lot of load for the server on bulk imports. If you log your QSOs directly in the WebUI the fields get populated automatically. If you work WSJT-X or something else you still can use the LogbookAdvanced "Update from Callbook" Feature. 

Therefore I gonna close this PR as rejected. I'm sorry that you put that effort in but you still can use this in your own instance. 

Vy 73 de HB9HIL

salyh:
I understand but obviously there seems do be a demand for this especially if the same discussion and contrubtions arise again and again. Therefoe i added a static config item which defaults to 'false'. This should give a user or admin full control depending on the server capacity (In my case I run a wavelog instance only for myself and will never hit any load issues).

Another idea would be to allow this kind of functionality for certain users or stations only (to limit load) or do it async via a cronjob. I happy to help adding such a more finegrained control.

For anyone who needs or want this functionality for the time beeing please look here: https://github.com/salyh/wavelog/wiki

TU
Hendrik (DO2SLY)

int2001:
Without repeating the argumentation again:

The API is also for bulk. And Bulk-Lookup against a Rest-API without a daemon isn't good. You'll run - even with one user - into serious issues, because php will process the QSOs synchronous. That's the reason it's built into Logbookadvanced. User can decide to abort (if it takes to long) and so on.

if you persist on "automatic lookup" for "automatic QSOs" it could be a better idea to enrich the QSOs before putting them into Wavelog. tbh: if i do phone or CW QSOs, i kindly ask for the name, QRZ is only a "safety net"

tdavis6:
If this was a PR on the WLGate end of things, would that work better? For example, when WSJT-X (for the sake of this example) sends the QSO to WLGate, and a setting is enabled within WLGate itself, it does a query with the callbook API, then sends the callbook information, along with the QSO information, to WaveLog? Thank you for your time with dealing with this!

int2001:
No, it should be done at the source imho. So a PR to WSJT-X for example. The main usecase for the API is for collecting QSOs. Like the ADIF-Import. It imports what has been made, including the data which is in it.

if a tool - like wsjt-x - is missing the informations you need, the missing informations should be added there.
a really good example is FLDigi: You type in the Call, do your RTTY/PSK31-QSO, type in all informations and press "log". if that's not enough you can press on a QRZ-Key at FLDigi, so QSO will be enriched before it is written to API or disk.

and the Callbook-API at Wavelog isn't meant as substitute for a payed QRZ-Subscription. Therefore we recommend to rate-limit it very tight.

So, if you plan to enrich QSOs (automatically) at its source, the source has to connect to the callbook(s) available.
The Bulk-"Update from Callbook" is already a compromise.

to be crystal clear: this will never be part of WavelogGate. One can discuss about something like "Personalized QRZ-Keys per User" plus check "only one QSO at a time" --> enable API-Lookup. but tbh: many many pitfalls

tdavis6:
Sounds good! Thank you for your time with this.

PR #2964, opened 2026-02-17
Dxcc stuff

HB9HIL:
### Refactor: Remove unnecessary $date from Dxcc class constructor

The `Dxcc` class in `src/Dxcc/Dxcc.php` accepts a `$date` parameter in the constructor, which adds an unnecessary `WHERE` clause to the SQL query when loading prefixes and exceptions. This creates several issues:

- The `dxcc_lookup()` method also accepts a date, meaning two different dates could be passed — one during class construction and another during lookup — which breaks the logic.
- Throughout the codebase, the date is used inconsistently (sometimes `null`, sometimes an actual date), but always intended to be today's date.
- The only place where the date genuinely differs due to user input is in `LogbookController::json()`.
- The additional `WHERE` clause is redundant, since start and end times are already checked within `dxcc_lookup()`.
- The date dependency in the constructor makes caching and cache invalidation unnecessarily complex.

#### Solution

Remove `$date` from `read_data()` and always load all prefixes and exceptions into the cache. Start/end validation already happens in `dxcc_lookup()`, and for `wpx()` the date is irrelevant.

### DIFF
--- a/application/controllers/Update.php
+++ b/application/controllers/Update.php
@@ -272,6 +272,8 @@ class Update extends CI_Controller {
 			$this->db->query($sql);
 			$this->db->trans_complete();
 
+			$this->_invalidate_dxcc_cache();
+
 			$this->update_status(__("DONE"));
 
 			echo 'success';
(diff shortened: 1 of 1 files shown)


AndreasK79:
Nice find. Good to simplify and remove not needed stuff!

PR #2965, opened 2026-02-17
Prevent downloading LoTW-Reports from users who have no QSO

int2001:
Situation/Issue:

Huge instance with thousands of Users. Some come, enter all of their credentials (including LoTW), and go without ever making/logging a QSO.

leading into Wavelog to Download LoTW-Confirms from the beginning on every time the sync runs. Blocking other Users from ever being synced.

The patch collects mitigations for it.
Furthermore it introduces parallel downloads for LoTW. max 5 at once.


phl0 (approved):
Tested this code on my dev and prod env (both single users) and it is working fine.

Issue #2970, opened 2026-02-18
When Custom Maptiles Server Live QSO Location auto lookup API still Use Default Settings


wohenbuguai:
### I have checked the following

- [x] I am running the latest version of Wavelog
- [x] I have cleared my browser cache
- [x] This is a bug, not a setup question or feature request

### Bug Description

When Custom Maptiles Server, Live QSO Location auto lookup API still Use Default Settings;
In China, OSM maps cannot be used normally. After I switched the tile map to AutoNavi, I found that the address search grid of the Live QSO module still uses OSM's API. Due to Time Out, the page is blocked and unresponsive. I hope a customized API can be provided or the API can be switched based on the tile server address.

<img width="1218" height="971" alt="Image" src="https://github.com/user-attachments/assets/b4621723-b0d1-4a8f-8d5d-6b7796445c84" />

<img width="2552" height="1118" alt="Image" src="https://github.com/user-attachments/assets/04ef9191-16c1-4f69-9ede-f21b2e4c6333" />

### Steps to Reproduce

Global Options > Maptiles Server > [Change Maptiles Server URL]
Username > Account > Wavelog Preference > Location auto lookup. "Yes“
QSO > Live QSO > Location > [Gridsquare should be automatically but not]

<img width="1125" height="931" alt="Image" src="https://github.com/user-attachments/assets/44ad7364-850f-42c5-8754-f60486346a45" />

### Expected Behavior

 hope a customized API can be provided or the API can be switched based on the Maptile server address.

### Wavelog Version

2.3.0

### PHP Version

8.2

### MySQL/MariaDB Version

8.0.36

### Server OS

Ubuntu 22

### Browser

Chrome,Edge

### Device Type

None

HB9HIL:
Hello @wohenbuguai 

Thanks for this Bug-Report. Indeed there are some issues in this feature which I fixd in the mentioned PR. However this may will fix your freezing browser (I added a timeout after 5 seconds), this will not solve your particular issue.

The QTH lookup has basically nothing to do with the map provider as the API which is used to lookup the coordinates is always the openstreetmap api. These are two different features. **So if `nominatim.openstreetmap.org` is not available in China the I fear you can't use this feature and I don't think there will be a fast solution for this.**  

The mentioned PR #2972  fixes the freezing brower and the grid handling. 

What takes me wonder is that the API call was successfully on your side (returns 200 GET)

[...]

PR #2972, opened 2026-02-18
Fixed QTH Lookup

HB9HIL:
There were several issues with the current QTH lookup in the QSO Logging. 

- There was no timeout the the ajax was explizit defined as async: false which cause the browser to freeze if openstreetmap is unavailable
- The insert function to insert the Locator in the Grid field was broken

I refactored the feature and added a timeout which should help for #2970. In addition I moved the js function `LatLng2Loc()` to common.js which is loaded everywhere and it makes more sense to have such an generic function globally available (lived in the radio helper)


int2001 (approved):
didn't know that we've such a feature.
Works so far, but i'd like to have @phl0 also have a look on it, since i am not 100% sure if it may conflict with VUCC/Grids.

From what i've tested/seen: no conflicts

phl0 (approved):
Tested and works. Don't currently see an issue with VUCC_GRIDS here. The lookup would always return lat/lon which gets converted to a (single) grid.

PR #2974, opened 2026-02-18
Hint for wrongly entered single grid into VUCC_GRIDS

phl0:
Entering a single maidenhead grid into the VUCC_GRIDS field leads to the QSO view failing to load:

image

We should catch this.

Also we show this as a hint to the user in QSO view by coloring the grid:

image


phl0:
Tnx review. I might add a mouse-over hint for explanation.

phl0:
Added a text hint to give a clue to the user.

image

Issue #2120, opened 2025-07-08
Req: 2-Factor-Authentication


Rakete2510:
**By submitting feature requests, you have the opportunity to contribute your ideas and inspiration to the Wavelog Dev-Team. Please be aware that not every feature request will be addressed or answered.**

**What Feature is Missing?**
A 2-Factor-Authentication Feature is missing  like a TOTP (Timed-Based-One-Time Password)

**Why is This Feature Important?**
As an additional security level

**Is This Feature Personal or Beneficial to Others as Well?**
Its Beneficial for other aswell.

HB9HIL:
Hello @Rakete2510 

This was already requested several times. I plan to create a PoC for that one but please keep in mind that this hasn't the hightes priority for us. But we don't reject it :)

Rakete2510:
Hi @HB9HIL 

I have already started with php myself and am currently trying to integrate it into the fork. maybe i can support you with that :)

HB9HIL:
Well we are always open to PRs (to the `dev` branch) :)

int2001:
jumping into this discussion.
Adding 2FA wouldn't mean "only add 2FA".
it means, that the API-Access has to be completly refactored. As long as one is able to modify/upload/read QSOs and things via API the 2FA would be senseless.

e.g.: Clublog has 2FA. The 2FA is absolutely nonsense there, because you're able to bypass it via the API. 
And tbh: We're not friends of snakeoil-security

kbabioch:
> jumping into this discussion. Adding 2FA wouldn't mean "only add 2FA". it means, that the API-Access has to be completly refactored. As long as one is able to modify/upload/read QSOs and things via API the 2FA would be senseless.
> 
> e.g.: Clublog has 2FA. The 2FA is absolutely nonsense there, because you're able to bypass it via the API. And tbh: We're not friends of snakeoil-security

I don't think this is true.

- First of all you can implement two factor authentication (2FA) for the regular user login separate from the API access. It is still a security benefit, since not everyone is using API access, etc.
- Second of all: Typically 2FA is **not** implemented for API access per se. There are tokens that are used for API access that can be monitored (last used, etc.), revoked independently and limited to specific tasks / API endpoints. As far as I can tell, this is already implemented for wavelog (see [here](https://github.com/wavelog/wavelog/wiki/API)).

[...]

Issue #789, opened 2024-08-22
OAuth login/register support!


ChristianSwift:
Request to add support for oauth2.0 login and registration (automatic registration)!  
For example, many site cluster systems are using CASDoor as user management system. If oauth authentication function can be provided, it can greatly increase the attractiveness of wavelog to administrators of existing websites and encourage them to use this log system without having to spend time on synchronizing the user table of the database.  
I hope you can consider my suggestion, thank you and 73!

user00265:
If this is difficult, there is a middle ground: external headers. This is supported by `authentik` as [Header Authentication](https://docs.goauthentik.io/docs/add-secure-apps/providers/proxy/header_authentication) and `Authelia` as [Trusted Header SSO](https://www.authelia.com/integration/trusted-header-sso/introduction/)... possibly others.

I am providing this in hope that it gives options of something to provide some form of SSO and then focus on making a OAuth-compliant application.

HB9HIL:
@user00265 SSO Support is not supported with #2978 
You might want to take a look into the documentation for it once it's merged. We tested Keycloak and Authentik. If you are familiar with Authelia we could add some documentation for it aswell

user00265:
I use Authelia and I'll be glad to take it for a spin! I'll update once I have done some testing.

HB9HIL:
@user00265 the documentation so far
https://github.com/wavelog/wavelog_docs/pull/4

user00265:
@HB9HIL I can confirm this is working with Authelia. Documentation to follow.

HB9HIL:
This is solved

PR #2978, opened 2026-02-19
Third Party Authentication (SSO/OIDC/Header Authentication)

HadleySo:
- Enable existing and new users to authenticate with JWT via trusted proxy header
- Enables OIDC login
- No clubstation direct login

Fixes 
- #1672 
- #789 
- #2120

### DIFF
--- a/application/controllers/Header_auth.php
+++ b/application/controllers/Header_auth.php
@@ -202,8 +202,8 @@ class Header_auth extends CI_Controller {
                 return null;
             }
 
-            $alg = $header['alg'] ?? '';
-            if (!in_array($alg, ['RS256', 'RS384', 'RS512', 'ES256', 'ES384'], true)) {
+            $alg = $header['alg'] ?? 'none';
+            if ($alg == "none") {
                 log_message('error', 'SSO Authentication: Algorithm "' . $alg . '" is not allowed.');
                 return null;
             }


HadleySo:
Call Notes:

- [x] User model flag
- [x] SSO users disable password login
- [x] Callsign from IdP
- [x] Use JWT for user data
- [x] UI: Option to hide username password from login
- [x] UI: Hide username and password from user menu

HB9HIL:
Since this is an ongoing development in teamwork with @HadleySo I drafted this PR

HB9HIL:
I think we can mark this PR as "Ready for Review" 

It needs good testing but I think this is good at the moment

kc9uhi:
I'm new to SSO and proxy auth, and learning quickly.

If I understand correctly, this requires an external proxy ahead of the wavelog server to perform user authentication, which effectively restricts *all* access to wavelog to sso authenticated users. To allow anonymous access to the 'visitor' logbook/oqrs pages, is the only method to 'allow-list' the appropriate paths in the external proxy to allow unauthenticated users? (such as /visitor,/index.php/oqrs,/assets,/js)

HadleySo:
@kc9uhi It would be the opposite, the authentication proxy would only be in front of the use login endpoint. All other endpoints would be public, with Wavelog authentication of course. 

There will be docs with example nginx and oauth2-proxy configs for Keycloak

HadleySo:
Here is an example nginx config, only `/index.php/header_auth/login` hits the auth proxy first.

<details>

<summary>Expand example nginx</summary>



```nginx
    server {

            

        location /oauth2/ {
            proxy_pass http://oauth2_proxy;
            proxy_set_header Host                    $host;
            proxy_set_header X-Real-IP               $remote_addr;
            proxy_set_header X-Auth-Request-Redirect $request_uri;

            proxy_buffer_size          16k;
            proxy_buffers              8 16k;
            proxy_busy_buffers_size    24k;
        }
        location = /oauth2/auth {
            proxy_pass       http://oauth2_proxy;
            proxy_set_header Host             $host;
            proxy_set_header X-Real-IP        $remote_addr;
            proxy_set_header X-Forwarded-Uri  $request_uri;
            # nginx auth_request includes headers but not body
            proxy_set_header Content-Length   "";
            proxy_pass_request_body           off;

        }
        location /index.php/header_auth/login {
            auth_request /oauth2/auth;

            auth_request_set $token $upstream_http_x_auth_request_access_token;
            proxy_set_header X-Auth-Request-Access-Token $token;
            error_page 401 =302 /oauth2/sign_in;

            proxy_pass http://app_backend;
            proxy_set_header Host $host;
            proxy_set_header X-Real-IP $remote_addr;
            proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
            proxy_set_header X-Forwarded-Proto $scheme;
        }
        location / {
            proxy_pass http://app_backend;
            proxy_set_header Host $host;
            proxy_set_header X-Real-IP $remote_addr;
            proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
            proxy_set_header X-Forwarded-Proto $scheme;
        }
    }
```

</details>

kc9uhi:
Thanks! This is making a lot more sense

int2001:
Checked this code as seen from a theoretical use (lack of SSO / OAuth-Provider for testing here). Some thoughts, besides the fact that SSO via OAuth is always very individual:

- User can change his Username / Mail or even Password. If Username changes, there'd be a "ghost" user in the system. The Ghost User wouldn't be accessible, because the claim from OAuth would create a new User...
- User and admin may trigger "forgot" or "reset_password" with no effect (OAuth is doing the job)
- It should be strongly recommended to set `auth_allow_direct_login = false` when using OAuth.

Regarding Individual OAuth:
the claims are hardcoded to the wavelog-fields. If someone already uses an IDP it would be already configured for other applications it is using. But this implementation insists on the hardcoded fields. Only configurable field is the one used for callsign. In reallife the IDPs are very individual. i'd suggest to make them configurable as well. i've also seen combinations of preferred_username, upn and sub as username in the wild...

[edit]One more thing which came to my mind: OAuth-Sessions are usually "sticky". Which means: The OAuth provider sets internally the session and stick it to the User. If this is configured in OAuth the "Logout" wouldn't be a real-logout (because User is still signed in at OAuth). Once he heads back to wavelog he'd automagically signed in. Consider a calling an explicit OAuth-Logout-URL in case of "Logout" at Wavelog should do that. Otherwise people with multiple identities are locked out / could only change identity if they use a fresh browser / incognito-mode. Logout at Wavelog for OAuth-Users has no effect there. At least worth a documentation.

tl'dr: The "Ghost-Issue" should be fixed at least. Furthermore: Think about more flexibility/configuration when it comes to claim-mapping.

HB9HIL:
I don't really agree with all points @int2001. While SSO can be individual an the most cases, it don't necessarily has to. It's not the goal of this PR to catch every possible use case. The goal is rathermore to implement a basic support for Open ID Connect (not Oauth, which is a different thing). 

> User can change his Username / Mail or even Password. If Username changes, there'd be a "ghost" user in the system. The Ghost User wouldn't be accessible, because the claim from OAuth would create a new User...

Thats true. Did is indeed still missing. I prepared the database flag but there should be some protection in the user::edit

> User and admin may trigger "forgot" or "reset_password" with no effect (OAuth is doing the job)

Not necessarily. There is nothing wrong with it - if direct login is enabled - to login directly into he user account which was created on the IDP data. Therefore also the password can be changed. But we need to catch these cases in the UI (same as above)

> It should be strongly recommended to set auth_allow_direct_login = false when using OAuth.

Nah I don't think so. It's important to provide this option but both is okay. 

> Regarding Individual OAuth:
the claims are hardcoded to the wavelog-fields. If someone already uses an IDP it would be already configured for other applications it is using. But this implementation insists on the hardcoded fields. Only configurable field is the one used for callsign. In reallife the IDPs are very individual. i'd suggest to make them configurable as well. i've also seen combinations of preferred_username, upn and sub as username in the wild...

Thats true. We could give the possibility to override those fields. I also could be discussed if we move the config stuff out of the config.php into a oidc.php

> [edit]One more thing which came to my mind: OAuth-Sessions are usually "sticky". Which means: The OAuth provider sets internally the session and stick it to the User. If this is configured in OAuth the "Logout" wouldn't be a real-logout (because User is still signed in at OAuth). Once he heads back to wavelog he'd automagically signed in. Consider a calling an explicit OAuth-Logout-URL in case of "Logout" at Wavelog should do that. Otherwise people with multiple identities are locked out / could only change identity if they use a fresh browser / incognito-mode. Logout at Wavelog for OAuth-Users has no effect there. At least worth a documentation.

We already discussed this and it's implemented like that. With the config option `auth_url_logout` you **can** define a logout URL which you get redirected to after the Wavelog session was killed. However since we are talking about OIDC and not Oauth, it's important to hit the Wavelog URL first to kill the Wavelog session first since this one is independent of the IdP session. In fact if an admin don't want to also logout out of the IdP session, which is absolutely okay the config option could be empty (`auth_url_logout` = `''`) which only kills the Wavelog session then. Both is legit.

HadleySo:
Most IdPs provide methods to construct and map custom JWT claim names - Keycloak, Authentik, Entra. Callsign will have to be a custom mapper anyways since its non-standard. But yes full claim name overrides is good.

I agree with disabling user modify username and password, but not email or name. Since user info is not updated on login, there is no way to propagate new user info from the IdP.

The example config has an example of the logout flow Wavelog session -> OAuth2-Proxy -> IdP front-channel.

HB9HIL:
Some questions popped up regarding JWT Token validation, unique user identifier and userdata updates from IdP's 
we need to refactor some stuff so we draft it again :smiley:

HB9HIL:
There is still one feature missing: Setting a clubstation via IdP. But since this is a bigger PR, the rest can be reviewed anyway.

application/controllers/User.php
```diff
@@ -177,6 +177,18 @@ function add() {
...
+		if ($data['auth_header_enable']) {
+			// expecting sso.php in the config folder
+			$this->config->load('sso', true, true);
+		}
+		$data['auth_header_allow_direct_login']  = $this->config->item('auth_header_allow_direct_login', 'sso') ?? true;
+		$data['auth_header_hide_password_field'] = $this->config->item('auth_header_hide_password_field', 'sso') ?? false;
+		$data['auth_header_locked_data_badge'] = $this->config->item('auth_header_locked_data_badge', 'sso') ?: 'IdP';
+		$data['auth_header_locked_data_tip'] = $this->config->item('auth_header_locked_data_tip', 'sso') ?: __("Can't be changed. Manage this through your Identity Provider.");
+		$data['sso_claim_config'] = $this->config->item('auth_headers_claim_config', 'sso') ?: [];
+
```
HB9HIL, line 187: Do we need to read the config values when adding a user manually? We could hardcode them instead couldn't we?

HB9HIL: tbh: did not check/test that

HadleySo: Actually yes a few should be hardcoded, I'll push in a moment

HadleySo: Resolved in [5f68295](https://github.com/wavelog/wavelog/pull/2978/commits/5f68295c9105d3cda94abb3b13768e0cd4861825)

application/models/User_model.php
```diff
@@ -746,6 +753,26 @@ function authenticate($username, $password) {
...
+	// FUNCTION: update specific user fields from SSO claims (bypass privilege check, used during login flow)
+	function update_sso_claims(int $user_id, array $fields): void {
+		// Cannot modify the following
+		$blocked = ['user_type', 'user_password', 'clubstation', 'external_account', 'login_attempts', 'created_at', 'modified_at', 'last_modified', 'last_seen', 'reset_password_date', 'reset_password_code'];
+		$fields = array_diff_key($fields, array_flip($blocked));
+
+		$this->db->where('user_id', $user_id);
+		$this->db->update('users', $fields);
+	}
+
```
HadleySo, line 818: Consider move to allowlist

HadleySo: [5eadc2a](https://github.com/wavelog/wavelog/pull/2978/commits/5eadc2a8a705470df4602ddec143a5ee33995207)

HB9HIL:
Really good job and nice teamwork @HadleySo

HadleySo:
Thanks for the help and support @HB9HIL!

application/controllers/Header_auth.php
```diff
@@ -0,0 +1,367 @@
...
+                return null;
+            }
+
+            if (isset($claims['nbf']) && $claims['nbf'] > time()) {
+                log_message('error', 'SSO Authentication: JWT Token is not valid yet.');
+                return null;
+            }
+
+            $alg = $header['alg'] ?? '';
+            if (!in_array($alg, ['RS256', 'RS384', 'RS512', 'ES256', 'ES384'], true)) {
```
kc9uhi: If the JWT signature isn't getting verified (empty($jwksUri)==true), is there any reason to check the signature type at this point?

Use case: Authentik built-in proxy utilizes HS256 signature

HadleySo: That's a holdover from before the firebase/JWT library was added, to check that OAuth2-Proxy had verified some signature.

We'll change it to accepting anything but none signature since there is JWKS now

application/controllers/Header_auth.php
```diff
@@ -0,0 +1,367 @@
...
+        if (empty($mapped['user_email']) || empty($mapped['user_callsign'])) {
+            log_message('error', 'SSO Authentication: Missing email or callsign claim in access token.');
+            $this->_sso_error();
+        }
+        if (empty($mapped['user_name'])) {
+            log_message('error', 'SSO Authentication: Missing username claim in access token.');
+            $this->_sso_error();
+        }
+
+        // $club_id = $this->config->item('auth_header_club_id', 'sso') ?: ''; // TODO: Add support to add a user to a clubstation
```
kc9uhi, line 267: May I suggest supporting a delimited list of clubstations when this does get implemented?

HadleySo: I was thinking mapping OIDC/claim groups to `club_id` so its more compatible. Easier to map whatever custom IAM to Keycloak or Authentik groups, compared to constructing a list into a custom claim.

Was there a particular use case @kc9uhi

kc9uhi: As long as there's support for more than one group/clubstation, it'll work and I won't have to customize too much. Perhaps a group prefix (/wl-[0-9]+/) to filter?

Use: multi-op multi-callsign special event such as CACTUS ( #2696 ). Multiple users, multiple event callsigns. Would be great if users were associated with clubstations when created from SSO. Current process involves back-door database brute force, as adding users to 8 clubstations is quite tedious via UI.

int2001 (approved):
Tried to dry-review code carefully.
This looks very solid now. Grats to all contributors.
It adds really nice flexibility in Enterprise-like environments.

As seen from code-perspective i can't spot any problems.
Real-Life will tell if it's safe and all edge-cases are covered.

Important for the "normal Users": If disabled it doesn't break anything.

But, some personal thoughts:
it adds complexity to the admin who wants to use-it/set it up. Proper knowledge of the OIDC-Provider is needed. 

i'd appreciate if we could add some kind of disclaimer in the wiki: Yes, we support Wavelog itself, but we're not able to help you setting up your OIDC-Provider.

application/controllers/Header_auth.php
HadleySo: #3103

Issue #2326, opened 2025-09-17
Support for CW Clubs


spREGR:
Request to add support for CW clubs during logging and also logging the membership number. 

During Live QSO/ Post QSO, once we input the callsign, today Wavelog displays Name, Photo etc from QRZ/HamQTH. It will be great to know if the OP is a member of a CW club and display that info. 

https://rbn.telegraphy.de/ uses RBN to display the callsigns and club names - an integration into Wavelog will be quite helpful amongst CW operators. Popular clubs are
1. LICW
2. CWOPS
3. SKCC
4. FOC
etc

73
VU2TUM, Puneit

int2001:
Tried this one?
Should fit to your needs. Simply create a list of Calls belonging to the club and add the URL of that List to Wavelog.

https://github.com/wavelog/wavelog/wiki/Hams-Of-Note

HB9HIL:
> Tried this one? Should fit to your needs. Simply create a list of Calls belonging to the club and add the URL of that List to Wavelog.
> 
> https://github.com/wavelog/wavelog/wiki/Hams-Of-Note

I fear this does not work properly. We have to work on that..

int2001:
It was meant to create an own list according to the wiki entry... however

cwhelchel:
I'd second this as well. Specifically via adding multiple user configurable call sign note lists. Here's a screen shot of how I display multiple "callsign notes" it in my logger Hunterlog (all are formatted like PoLo's hams of note list)

<img width="378" height="262" alt="Image" src="https://github.com/user-attachments/assets/06a53c33-14bb-4000-9e42-a53dd7d239e4" />

int2001:
let's talk about scaleability in this context.
if we implement such a feature, where every user could upload his **own** lists, we've to deal with the following things:
- Redundant Data in Database, because people are using the same lists.
- Housekeeping. People will upload their lists and forgot about their accounts. So some mechansim to delete/remove those lists would be needed if the user didn't use Wavelog for nn days/months/whatever
- priorities. If an OP is listed in the (existing) instance-wide list AND as well within the Userlist: What should be shown, without wasting space in log-window?

in case someone says: But i am alone on my instance, no need for housekeeping / checking redundant data, the question is: What stops you from using the already existing centralized list? Even the format is the same. and it's a modern implementation with the URL, so you may maintain the lists outside Wavelog (it's a logger, no CMS).

just first thoughts/issues which need to be solved before a possible implementation.

[...]

PR #2980, opened 2026-02-19
Show all HoN-Entries if there are multiple ones in src-file

int2001:
If there are multiple entries in HoN-File (instance) for one call, only one entry is shown.
This one extends the display to show all entries

Addresses #2326 

e.g.:
image


phl0 (approved):
Works for me.

PR #2981, opened 2026-02-19
Query callbook (hamqth) for DARC DOK

phl0:
This addresses https://github.com/wavelog/wavelog/discussions/2977.

We removed the Lookup->dok function which needed a separate HTTP request to lookup the DARC DOK. Instead we moved the function to the logbook model and added a function to fetch the DOK from callbook (only hamtqh support dok currently).

### DIFF
--- a/assets/js/sections/qso.js
+++ b/assets/js/sections/qso.js
@@ -1486,7 +1486,7 @@ $("#callsign").on("focusout", function () {
 						eqsl_member: result.eqsl_member,
 						qsl_manager: result.qsl_manager,
 						slot_confirmed: result.dxcc_confirmed_on_band_mode,
-						darc_dok: result.darc_dok
+						darc_dok: result.callsign_darc_dok
 					};
 					window.broadcastLookupResult(broadcastData);
 				}


assets/js/sections/qso.js
```diff
@@ -1485,7 +1485,8 @@ $("#callsign").on("focusout", function () {
 						lotw_days: result.lotw_days,
 						eqsl_member: result.eqsl_member,
 						qsl_manager: result.qsl_manager,
-						slot_confirmed: result.dxcc_confirmed_on_band_mode
+						slot_confirmed: result.dxcc_confirmed_on_band_mode,
+						darc_dok: result.darc_dok
```
int2001: isn't it `"callsign_darc_dok"` instead of `darc_dok` which is returned by json?
Have a look at L181 of Logbook.php-Controller

int2001 (changes requested):

assets/js/sections/qso.js
phl0: Indeed. I renamed it but forgot to change here. Fixed.

PR #2984, opened 2026-02-20
Fix Callbook Session persistence

HB9HIL:
It never worked properly. We stored the callbook session keys in the session data but due to the fact that the `User_model::update_session()` method overwrites the session data and the callbook session keys weren't part of that we obtained a fresh session key every single time.. Also on update from callbook in the LBA.. 

This is not only highly inefficient, but also not good for Wavelog's reputation in terms of load at third-party services. 

I refactored the session key stuff and now use the freshly refactored cache logic. Why caching and not session storage? Well the session keys have a lifetime. And why building a manual timebased session key invalidation if the cache system has this already built in. 

This reduced not only the load on third party services but also makes the callbook query around 40-50% faster!

Tested qrz.com and hamqth. Can't test qrzcq (no premium) and qrzru (no account) but the logic is still the same but now reusing session keys as it should be.

### DIFF
--- a/application/libraries/Hamqth.php
+++ b/application/libraries/Hamqth.php
@@ -127,6 +127,7 @@ class Hamqth {
 
 			}
 		} finally {
+			$data['source'] = $this->sourcename();
 			return $data;
 		}
 	}
--- a/application/libraries/Qrz.php
+++ b/application/libraries/Qrz.php
@@ -182,6 +182,7 @@ class Qrz {
 				$data['cqzone'] = '';
 			}
 		} finally {
+			$data['source'] = $this->sourcename();
 			return $data;
 		}
 	}
--- a/application/libraries/Qrzcq.php
+++ b/application/libraries/Qrzcq.php
@@ -143,6 +143,7 @@ class Qrzcq {
 
 			}
 		} finally {
+			$data['source'] = $this->sourcename();
 			return $data;
 		}
 	}
--- a/application/libraries/Qrzru.php
+++ b/application/libraries/Qrzru.php
@@ -114,6 +114,7 @@ class Qrzru {
 				$data['cqz'] = '';
 			}
 		} finally {
+			$data['source'] = $this->sourcename();
 			return $data;
 		}
 	}


application/libraries/Callbook.php
```diff
@@ -131,29 +171,29 @@ function qrzcq($callsign) {
...
 			}
 
-			$callbook = $this->ci->qrzcq->search($callsign, $this->ci->session->userdata('qrzcq_session_key'));
+			$callbook = $this->ci->qrzcq->search($callsign, $this->ci->cache->get($this->qrzcq_session_cachekey));
 
 			if ($callbook['error'] ?? '' == 'Invalid session key') {
 				$qrzcq_session_key = $this->ci->qrzcq->session($username, $password);
-				$this->ci->session->set_userdata('qrzcq_session_key', $qrzcq_session_key);
-				$callbook = $this->ci->qrzcq->search($callsign, $this->ci->session->userdata('qrzcq_session_key'));
+				$this->ci->cache->save($this->qrzcq_session_cachekey, $qrzcq_session_key, self::QRZCQ_SESSION_DURATION);
```
int2001: shouldn't it be `$qrzcq_session_key[1]` instead of whole array?
compare to L177 of the PR

int2001 (changes requested):
not sure - but the handling for initial looks different to the retry for qrzcq

application/libraries/Callbook.php
HB9HIL: Yep you are right! thx

application/libraries/Callbook.php
```diff
@@ -215,23 +255,23 @@ function qrzru($callsign) {
 			$username = $this->ci->config->item('qrzru_username');
 			$password = $this->ci->config->item('qrzru_password');
 
-			if (!$this->ci->session->userdata('qrzru_session_key')) {
+			if (!$this->ci->cache->get($this->qrzru_session_cachekey)) {
```
int2001, line 285:
```suggestion
            if (!$this->ci->cache->get($this->qrzru_session_cachekey)) {
                    $result = $this->ci->qrzru->session($username, $password);
                    if ($result) {
                            $this->ci->cache->save($this->qrzru_session_cachekey, $result, self::QRZRU_SESSION_DURATION);
                        }
                }
```

One more minor thing, which may cause trouble for the qrzu-users:

If ->session() fails (returns null/false), the bad result gets cached for 55 minutes — every lookup in that window will silently fail until the TTL expires. Since the cache outlasts a browser session, this failure mode is stickier than the old session-based approach.

Not sure if suggestion is directly applicable from github. it's more a hint ;)

HB9HIL:
@phl0 tnx fer review. I will clean things up a bit and add better error handling due to the new logic

phl0:
Also tested against hamqth with same result (~ double speed)

HB9HIL:
@phl0 and @int2001 

I guess now you test again. @int2001 I added a simple validation without overcomplicating things. If we want to refactor the whole logic in the future I'd suggest a dedicated CallbookExceptionClass which could handle invalida keys, timeouts, and so on for all callbook providers. But for now I think this should be okay

int2001 (approved):
great!

phl0 (approved):
All (still) working fine

PR #2985, opened 2026-02-20
Use HamQTH-DOK lookup as well for "Update from Callbook"

int2001:
#2981 introduced DOK-Lookup from HamQTH.
While it works wonderful during logging, the issue-owner states: "Doesn't work for Logbookadvanced or Edit". See #2977 

This patch adds the functionality to "Edit" and Logbookadanced, too.


phl0 (approved):
w0rx

PR #2992, opened 2026-02-20
Use APCu caching

HB9HIL:
We will use APCu as default cache for **new** users, Docker already supports it, it works out-of-the-box (if `php-apcu` is installed) and it improves user experience on slow systems.

In addition to that I improved the cache debugging in the debug view. It catches now more cases and also if both adapters fail (dummy adapter case).

general information: file caching fails entirely and codeigniter fallsback to "dummy" adapter, the user can still use Wavelog normally. It just will be slow and the admin see's warnings and information in debug view.

### DIFF
--- a/application/controllers/Debug.php
+++ b/application/controllers/Debug.php
@@ -110,7 +110,7 @@ class Debug extends CI_Controller
 		$data['cache_available_adapters'] = $cache_info['adapters'];
 		$data['cache_path'] = $cache_info['config']['cache_path'] ?: 'application/cache';
 		$data['cache_adapter'] = strtolower($cache_info['config']['cache_adapter'] ?? 'file');
-		$data['cache_backup'] = strtolower($cache_info['config']['cache_backup']) ?? 'file';
+		$data['cache_backup'] = strtolower($cache_info['config']['cache_backup'] ?? 'file');
 		$data['cache_key_prefix'] = $cache_info['config']['cache_key_prefix'] ?: __("(empty)");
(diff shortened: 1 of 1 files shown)


application/controllers/Debug.php
```diff
@@ -109,10 +109,10 @@ public function index() {
 		$cache_info = $this->Debug_model->get_cache_info();
 		$data['cache_available_adapters'] = $cache_info['adapters'];
 		$data['cache_path'] = $cache_info['config']['cache_path'] ?: 'application/cache';
-		$data['cache_adapter'] = ucfirst($cache_info['config']['cache_adapter'] ?? 'file');
-		$data['cache_backup'] = ucfirst($cache_info['config']['cache_backup'] ?? 'file');
+		$data['cache_adapter'] = strtolower($cache_info['config']['cache_adapter'] ?? 'file');
+		$data['cache_backup'] = strtolower($cache_info['config']['cache_backup']) ?? 'file';
```
int2001:
```suggestion
		$data['cache_backup'] = strtolower($cache_info['config']['cache_backup'] ?? 'file');
```
possibly only a typo...

int2001 (changes requested):
works. good one.
pse check typo. imho the fallback to file should happen **within** strtolower, not outside

application/controllers/Debug.php
HB9HIL: This doesn't matter and has nod effect. 'file' is already hardcoded lowerstring 😄 but no problem, I'll add that

PR #2993, opened 2026-02-21
Fix display of 'HRDLog realtime upload' in Station location list

kc9uhi:
Display the three possible states of "HRDLog realtime upload" (Disabled, No, Yes) in Station location list (/stationsetup/list_locations).

Previously displayed two values (No, Yes) with improper display of "Yes" for "Disabled".

Station Setup -> Station Locations, Show a location list button.  "HRDLog realtime upload" column now displays proper value.


AndreasK79:
@kc9uhi thanks for the fix!

PR #2997, opened 2026-02-21
set random cookies to prevent collisions

### LINKED ISSUE #2996: Prevent issues for new installations when switching between instances
@mdiepart claimed in #2994  (converted to #2995 ) that after he visited demo.wavelog.org and he installed wavelog himself he faced the error banner "You are not allowed to do that!". This happened because I did not add a custom cookie prefix for the demo (which is already fixed)

in addition this can happen also to other users on fresh installs which are switching between installations (which are not properly configured). We can catch that for the future by generating some sort of random prefixes for session and cookie. Needs to be added in the installer.

HB9HIL:
fixes #2996

### DIFF
--- a/install/config/config.php
+++ b/install/config/config.php
@@ -520,7 +520,7 @@ $config['encryption_key'] = '%encryptionkey%';
 |
 */
 $config['sess_driver'] = 'files';
-$config['sess_cookie_name'] = 'ci_wavelog';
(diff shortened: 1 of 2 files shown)


HB9HIL:
I'm a bit dump today.. this is not the issue (cookies are not cross domain at wavelog)

PR #2998, opened 2026-02-21
ADIF import: update existing QSOs by APP_WAVELOG_QSOID for QSL status synchronization

lycanananas:
This change adds an optional ADIF import mode that updates existing QSOs (instead of creating duplicates) when APP_WAVELOG_QSOID is present and matches a QSO in the same station profile.

This is primarily meant for synchronization with external QSL tools, so users can refresh QSL-related state and sent/confirmation method in Wavelog after external processing.

**Safety Consideration / Follow-up**
This feature should be used with care.
At the moment, an update may overwrite all mapped fields for a matched QSO, not only QSL-related data.

It would be worth considering one of these approaches:

* **Safe update mode (recommended):** update only QSL-related fields. With additional checks except QSOID.
* **Dangerous/full update mode:** keep current behavior, but mark it explicitly as advanced/high-risk and require clear user confirmation.

### DIFF
--- a/application/views/adif/import.php
+++ b/application/views/adif/import.php
@@ -145,24 +145,24 @@
 										</div>
 										<div class="small form-text text-muted"><?= __("Select if QSOs shall be imported, even if they already exist.") ?></div>
 									</div>
-										<div class="col-md-6">
-											<div class="form-check-inline">
-												<input class="form-check-input" type="checkbox" name="dxccAdif" value="1" id="dxccAdif">
-													<label class="form-check-label" for="dxccAdif"><?= __("Use DXCC information from ADIF") ?></label>
-											</div>
-											<div class="small form-text text-muted"><?= __("If not selected, Wavelog will attempt to determine DXCC information automatically.") ?></div>
+									<div class="col-md-6">
+										<div class="form-check-inline">
+											<input class="form-check-input" type="checkbox" name="dxccAdif" value="1" id="dxccAdif">
+											<label class="form-check-label" for="dxccAdif"><?= __("Use DXCC information from ADIF") ?></label>
 										</div>
+										<div class="small form-text text-muted"><?= __("If not selected, Wavelog will attempt to determine DXCC information automatically.") ?></div>
+									</div>
 								</div>
 
-                                    <div class="mb-3 row">
-                                        <div class="col-md-6">
-                                            <div class="form-check-inline">
-                                                <input class="form-check-input" type="checkbox" name="updateQso" value="1" id="updateQso">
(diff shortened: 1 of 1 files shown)


HB9HIL:
Hello @lycanananas 

I have to be honest with you. This is my opinion and we didn't discussed this in the team yet. But I want to say my opinion here:

I don't like this approach. We got an API for the last contact which you already using in your qslmaster (i checked the code). Besides the fact that you hardcoded `0` which pulls then all QSOs everytime (which is inefficient) why not a PR which extends this API point with the qsoid? Would be much much easier wo work over the API in json format then using ADIF for this. Next thing is that the ADIF import/export is used by everyone. Why should we export a non-adif field with the Wavelog QSO-ID at every adif export when this is only necessary on third party tools which talk over an API?

Please rethink this. Extending the `get_last_contacts()` API with the QSO-ID for outbound and the `qso()` API for inbound is a much better approach then dealing with ADIF in this case. 

Just my 5ct here. I'm open for discussions. But in the end: There is an ADIF standard which we want stick to.

HB9HIL:
PS: and there is definitly no need for any UI elements. Don't get me wrong. I like your tool but we have this feature in Wavelog. If you want to connect to Wavelog please use the API.

lycanananas:
ADIF explicitly supports application-defined fields:
https://www.adif.org/316/ADIF_316.htm#Application_defined_Fields

Currently, the Wavelog API (and Wavelog itself) does not support mass updates of QSOs. If full CRUD operations were available for QSOs, it would be possible to update specific records from other app.

At the moment, the API only allows downloading existing QSOs or adding new ones. There is no way to update or edit existing QSOs via the API or any other method except manual modification in the UI.

lycanananas:
Imagine integration with a QSL print service. You send them an ADIF file with your QSOs, and they send you back an ADIF file with information about which QSOs were printed and which were verified. In Poland you need to include local bureau number in via field.

Right now there’s no way to automatically update existing QSOs in Wavelog based on that returned file. Without update support in the API, this kind of workflow has to be done manually, which defeats the purpose of automation.

HB9HIL:
I know that this is currently not possible. But why bothering the User with this by uploading and downloading ADIF files? Extending the outbound qso api with the qsoid is super easy. The inbound qso api can be modified to update an qso if an qsoid is provided (and after some checks obviously). Everything in API, Everything in JSON, So easy

lycanananas:
Okey. API way looks also fine for me. It is full CRUD approach. That I suggested.

But API needs also export QSOs with ID so I can be easily updated.

HB9HIL:
@lycanananas I will prepare something for you okay?

HB9HIL:
had a chat with @lycanananas 
he will create a new PR when time with a logic to match qso's by time/date/mode/band/call, etc

PR #2999, opened 2026-02-21
[Advanced Logbook] Added duration as a column

AndreasK79:
In response to #2983.

image

Added column and filtering


phl0 (approved):
Works. However I wonder how these times can end up in the logbook while using the WSJT-X improved interface to log to Wavelog directly...

Image

Guess I have some fixes to do in my logbook ....

PR #3001, opened 2026-02-22
Add option to disable Map at QSO-Window

int2001:
Some Users don't want to see the Map at the QSO-Window (Even when working mobile).
So this one adds a option at Settings to "hide" the Map at QSO-Window


HB9HIL (approved):
works

PR #3005, opened 2026-02-23
Add Gridsquare locator function

eddiewebb:
This file contains a function that converts WGS84 coordinates into the corresponding Maidenhead Locator during callsign lookup on logbook entry.


HB9HIL:
Hello @eddiewebb 

is it necessary to have bot/ai references in the commit history?

grafik

application/helpers/maidenhead_helper.php
HB9HIL, line 1: Please do not add this as new helper file. We got a QRA Library which can be used to integrate this function

application/libraries/Qra.php

application/helpers/maidenhead_helper.php
```diff
@@ -0,0 +1,47 @@
...
+  $latitude += 90;
+
+
+  $letterA = ord('A');
+  $numberZero = ord('0');
+
+  $locator = chr($letterA + intval($longitude / 20));
+  $locator .= chr($letterA + intval($latitude / 10));
+  $locator .= chr($numberZero + intval(($longitude % 20) / 2));
+  $locator .= chr($numberZero + intval($latitude % 10));
```
HB9HIL: using modulo on floats is not that nice in a semantic manner. It works here but `fmod()` would be nicer

eddiewebb:
Will use the QRA library for Qrz.ocm lookups which wont return gridsquare without subscription.

Hamqth returns it on free tier, but someting in their response is not playing nice with the library, and returning grid as both Grid and qth (typically city I would expect)

I'll re-add the logic I had in Logbook controller but with call to QRA library where missing but lat/lon present.

PR #3016, opened 2026-02-25
Simplify lookup from previous QSOs

int2001:
The (7yr) old code used 11 (Eleven) queries to fetch information from previous QSOs.
Very inefficient plus the fact, that some functions had weird var-names.

this one consolicates those queries to only 1 query, which is more readable and more efficient.

Logic-Change(!) - discussable:
Old Function searched ALL QSOs for missing attributes,
New one only the last 10

At the end it was a tradeoff decision.

Testing:
- Go to Log-window.
- Enter a call you logged before
- Information from call before should be displayed via /json function

### DIFF
--- a/application/controllers/Logbook.php
+++ b/application/controllers/Logbook.php
@@ -166,18 +166,21 @@ class Logbook extends CI_Controller {
 		// Get user's lookup priority preference
 		$lookup_priority = $this->get_lookup_priority();
 
-		$return['callsign_name'] 		= $this->nval($this->logbook_model->call_name($callsign), $callbook['name'] ?? '', $lookup_priority);
-		$return['callsign_qra'] 		= $this->nval($this->logbook_model->call_qra($callsign), $callbook['gridsquare'] ?? '', $lookup_priority);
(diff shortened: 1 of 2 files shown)


AndreasK79:
@int2001 tested working here. Nice work.
Why only 10 QSOs? Since the optimization is great here, why not extend it more?

AndreasK79 (approved):

int2001:
> ...Why only 10 QSOs? Since the optimization is great here, why not extend it more?

The 10 are already a safety net.
They are "needed" for the following edgecase:

QSO 1 - 2025-01-01 LA8AJA, email: NULL
QSO 2 - 2025-01-03 LA8AJA, email: mailme@operator.ham
QSO 3 - 2025-01-04 LA8AJA, email: NULL
QSO 4 - 2025-01-05 LA8AJA, email: NULL

without the 10, email would be NULL (if not fetched from callbook). With this 10, it looks back 10 QSOs to collect missing fields. Very unlikely that valid data can't be found 10 QSOs back. (Remember: EdgeCase)

phl0:
Tested and works without issues so far on my end.

PR #3021, opened 2026-02-28
Dxcc award optimized

AndreasK79:
Major speedup for the DXCC award.
Using the same logic as I used in the CQ/ITU zone award.

Testing with a giant logbook:
Before:
16.55s 100 queries

After:
3.7s 31 queries


int2001:
Tested: Several issue with Map and with SAT-QSOs found.

phl0:
The map and QSO details modals are not in sync with regards to band "all". Table lists all bands excl. SAT. Map also seems to show map excl. SAT. But the QSO details modal seems to list QSOs regardless of band (i.e. incl. SAT):

image

ZD8 is red because only worked on SAT but QSO details modal lists (SAT) QSOs.

AndreasK79:
Thanks for testing. Need to fix some of the SAT logic then. I'll let you know when you can test again :)

phl0:
image

Discovered while showing "worked" for "All" bands.

PR #3025, opened 2026-03-01
Wae optimized

AndreasK79:
WAE optimization. Test and break.


phl0:
There is an issue with the non-ADIF standardized entities not being counted (correctly). Example TA1:

dev
image

this branch
image

phl0 (approved):
Works as expected now.

phl0 (approved):

PR #3026, opened 2026-03-01
[CQ Award] Fix for sat workeD/confirmed/not worked

AndreasK79:
This should fix the filtering for SATS


phl0:
Tested and fix for SAT QSOs confirmed. worked / confirmed / not worked now working correctly.

PR #3040, opened 2026-03-05
Take References from QSO-Table

int2001:
With this patch (discussable) the logic for retrieving specific MY-Informations slightly changes. 

Before patch: Everything was taken from profile, even if QSO-Table has different information.

With Patch:

From QSO table:
  - MY_IOTA, MY_SOTA_REF, MY_WWFF_REF, MY_POTA_REF, MY_SIG, MY_SIG_INFO

 Still from station_profile join:
  - STATION_CALLSIGN → station_callsign
  - MY_CITY → station_city
  - MY_COUNTRY → station_country
  - MY_DXCC → station_dxcc
  - MY_GRIDSQUARE / MY_VUCC_GRIDS → station_gridsquare
  - MY_CQ_ZONE → station_cq
  - MY_ITU_ZONE → station_itu
  - MY_STATE → state
  - MY_CNTY → station_cnty

----
Testing:
- Log a couple of QSOs
- Change - e.g. - IOTA in your station_location
- Add more QSOs

- Have a look at them at QSO-Table.
- Have a look at them at detail-view.
- Export them.

compare dev to this.

### DIFF
--- a/application/controllers/Labels.php
+++ b/application/controllers/Labels.php
@@ -297,12 +297,12 @@ class Labels extends CI_Controller {
 				'sat_band_rx' => ($qso->COL_BAND_RX ?? ''),
 				'qsl_recvd' => $qso->COL_QSL_RCVD,
 				'mycall' => $qso->COL_STATION_CALLSIGN,
-				'sig' => $qso->station_sig ?? '',
-				'sig_info' => $qso->station_sig_info ?? '',
-				'sota' => $qso->station_sota ?? '',
-				'iota' => $qso->station_iota ?? '',
-				'pota' => $qso->station_pota ?? '',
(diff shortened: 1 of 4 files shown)


kc9uhi:
MY_*OTA are all location specific references, just like ITU, CQ, STATE, etc. Why separate them from the station_profile info? This would break the current logic of station_profile holding the authoritative location information for associated  QSOs

int2001:
> MY_*OTA are all location specific references, just like ITU, CQ, STATE, etc. Why separate them from the station_profile info? This would break the current logic of station_profile holding the authoritative location information for associated QSOs

valid points. we're also discussing it at the moment. there's no "white smoke", yet.
that's the reason for drafting it ;)

PR #3049, opened 2026-03-09
Rename existing SAT mappings for AMSAT SAT status upload

phl0:
SAT names for amsat.org/status have been renamed. We need to take care about this on upload.

- [x] Rename existing mappings
- [x] Add SATs which do not have a mapping yet but require one now


phl0:
Added remaining active SATs and completed the mapping function. So ready for review here.

PR #3069, opened 2026-03-15
Check returned array properly

phl0:
During working on https://github.com/wavelog/wavelog/pull/3068 we found an issue with the return value of `list_logbook_relationships`. In https://github.com/wavelog/wavelog/commit/f6ddbe1a69bfbadc25233dc1ed81dd6c7a57c377 the return value was changed so that an array is returned also in case of failure. But the check for the returned value in the API controller was not adjusted accordingly.


phl0:
We probably have to check (many) more checks for return value. Drafting it for review of the other function calls.

phl0:
> if you want to be 100% sure, replace `==` with `===` (Rare-Edge-Cases)

Applied that. Pls re-review :)

PR #3078, opened 2026-03-15
Switch to percent on QSL-View

int2001:
Slightly overengeneered... 
but on demand ;)

no problem if PR will be declined

### DIFF
--- a/application/views/statistics/qsltable.php
+++ b/application/views/statistics/qsltable.php
@@ -1,8 +1,12 @@
 <div class="container">
 <br />
 <div class="card">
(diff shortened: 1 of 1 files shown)


AndreasK79 (approved):

phl0:
Oh, why not both? 😇✌️

int2001:
> Oh, why not both? 😇✌️

Switch it ;)

phl0:
> > Oh, why not both? 😇✌️
> 
> Switch it ;)

Should have tried before commenting :)

PR #3083, opened 2026-03-16
Cache Buster Log Flood

HB9HIL:
With the implementation of the cache buster in #2945 there were some changes how files get loaded. A little log flaw appeared in the datatables language loader. This was catched with an exception for the error log directly in the cache buster. However this did not solve the issue completely and in some cases in produces a lot of error logs. 

This solves this now by refactoring the datatables language url logic and removing the exception in the cache buster.


int2001 (approved):
works

PR #3096, opened 2026-03-18
Show OpenSSL version in debug view

phl0:


int2001 (approved):
Consider adding `?? __("No SSL found");

PR #3098, opened 2026-03-19
[Advanced Logbook] Added frequency to batch edit

AndreasK79:


application/models/Logbookadvanced_model.php
```diff
@@ -1186,6 +1187,22 @@ function saveEditedQsos($ids, $column, $value, $value2, $value3, $value4) {
 			$frequencyBandRx = $bandrx == '' ? null : $this->frequency->defaultFrequencies[$bandrx]['CW'];
 
 			$query = $this->db->query($sql, array($value, $value2, $frequencyBand, $frequencyBandRx, json_decode($ids, true), $this->session->userdata('user_id')));
+		}  else if ($column == 'COL_FREQUENCY') {
+
+			if ($value == '') return;
+
+			$value2 == '' ? '' : $value2;
```
int2001: whoot?

AndreasK79: It was copied from band, then I changed the name of the first variable, and I see that it doesn't make sense at all :) Removed that line now.

int2001 (changes requested):
if RX Freq is blank, it fails to update with 500

AndreasK79:
@int2001 very strange, as that doesn't happen in my dev setup...

int2001 (approved):
Works here now. since we've different db-engines i'd appreciate if you can test at least also, andreas

AndreasK79:
Still works here. Thanks for the fix.

PR #3100, opened 2026-03-20
Show indication if filter is set

int2001:
LBA-Filter-Indicator.

Before patch:
No visual feedback if a filter was set (regardless of applied).

After Patch:
- Filter-Button becomes yellow, if there's a difference to the default Values.
- Yellow removed if all values back to normal

Discussable (but complex / much more effort):
make the button yellow which isn't default...

Therefore draft at 1st


phl0 (approved):
Tested and works fine for me.

AndreasK79 (approved):

PR #3101, opened 2026-03-20
Try to handle Bytecount vs. Charcount at ADIF

int2001:
Hybrid Approach for ADIF ingestion.

Problemstatement:
1. Some ADIFs are counting the Characters (as it is in ADIF-Spec)
2. Some ADIFs are counting the Bytes
3. Some ADIFs contain UTF-8 Characters (NOT alllowed according to ADIF-Spec, if strictly interpreted)

The Problem starts with UTF8/Unicode chars, since they consume 2 Bytes for a (special)-character.

The WL-Parser is actually able to handle the following scenarios:
- Simply count (ASCII-)chars, compare to given length and parse the Field (fully ADIF-compliant)
- Count "Characters" for fields with utf8-stuff. e.g.: `<NAME:5>Jorgé`, compare with length and parse it (not ADIF-compliant)

This one Breaks in the following example:
 - Count "Characters" for fields with utf8-stuff. e.g.: `<NAME:6>Jorgé`, compare with length and parse it (not ADIF-compliant). See the `6`? 6 Bytes are given, because the `é` is special-char

This patch is trying to mitigate that. How?
At first it tries to check the lengths as before. if there's a difference it falls back to Byte-Counting. if that also fails it skips the field.

This needs to be tested well

### DIFF
--- a/application/libraries/Adif_parser.php
+++ b/application/libraries/Adif_parser.php
@@ -227,7 +227,7 @@ class ADIF_Parser
 			}
 
 			$a += $consumed_chars - 1; // adjust for loop increment
-			$return[mb_strtolower($tag_name, "UTF-8")] = $value;
+			\$return[mb_strtolower(\$tag_name, "UTF-8")] = rtrim(\$value);
 		}
 
 		// skip comments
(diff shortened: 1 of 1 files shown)


AndreasK79 (approved):

phl0 (approved):
W0rx!

PR #3104, opened 2026-03-21
[Advanced Logbook] Added action for merging QSOs, with dialog to select columns

AndreasK79:
Go to the advanced logbook -> select 2 qsos -> click merge qsos under action menu.
Select and merge.


phl0 (changes requested):
Basically works. However there is a bug that selecting primary or secondary QSO leaves both radio buttons checked.
I created a PR against the patch branch to fix this:

https://github.com/AndreasK79/wavelog/pull/63

It also includes showing lines with different values between primary and secondary QSO in bold to aid finding the differences.

phl0 (approved):
Fine with me now :)

PR #3107, opened 2026-03-21
Fix CSV-Download at LBA

int2001:
A recent change in handling of the whole table at LBA caused that "CSV"-Export wasn't working.
This one fixes the issue


phl0 (approved):
Fixes the issue.

AndreasK79 (approved):

AndreasK79:
Makes it work again. Also pushed a small fix for data stripping from columns.

PR #3118, opened 2026-03-23
Lba quicksearch fix

### LINKED ISSUE #3110: Problems with Quickfilters in Advanced Logbook
### I have checked the following

- [x] I am running the latest version of Wavelog
- [x] I have cleared my browser cache
- [x] This is a bug, not a setup question or feature request

### Bug Description

I open Advanced Logbook and select one of the locations. Then, I click on the "Search" button and the QSO list appears as expected.
Next, I select one of the QSOs (checkbox next to it), go into "Quickfilters" and then click "Search Date" in order to see only the QSOs made on the same date as the selected QSO.

The result is that no QSOs are found.
One strange thing I've noticed is that in the Location dropdown list the location still seems to be selected and is highlighted, but the checkbox next to it is unchecked!
This also happens when more than one location initially selected: they all remain highlighted and the header of the dropdown list says they are active, but the checkboxes are all unchecked and the search returns no results.

### Steps to Reproduce

1. go to Logbook > Advanced
2. select a Station Location
3. click on Search
4. select/mark one QSO (checkbox)
5. go to Quickfilters > Search Date

### Expected Behavior

The QSOs should show and the Location list should behave normally.

### Wavelog Version

2.3.1

### PHP Version

8.3.6

### MySQL/MariaDB Version

MariaDB 10.11.14

### Server OS

Ubuntu 24.04.4 LTS

### Browser

Chrome

### Device Type

Desktop

### Screenshots / Browser Console

Image

Image

Image

### Additional Context

_No response_

AndreasK79:
Fixes #3110

### DIFF
--- a/assets/js/sections/logbookadvanced.js
+++ b/assets/js/sections/logbookadvanced.js
@@ -670,7 +670,7 @@ $(document).ready(function () {
 	$('#dxcc').multiselect({
 		// template is needed for bs5 support
 		templates: {
-		  button: '<button type="button" class="multiselect dropdown-toggle btn btn-sm btn-secondary me-2 w-auto" data-bs-toggle="dropdown" aria-expanded="false"><span class="multiselect-selected-text"></span></button>',
+			button: '<button type="button" class="multiselect dropdown-toggle btn btn-sm btn-secondary me-2 w-auto" data-bs-toggle="dropdown" aria-expanded="false"><span class="multiselect-selected-text"></span></button>',
 		},
 		enableFiltering: true,
 		enableFullValueFiltering: false,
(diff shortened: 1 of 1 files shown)


phl0 (changes requested):
There still seems to be some bug here. If I do as described in https://github.com/wavelog/wavelog/issues/3110 and select only one location then check QSO and then do the date quick search I end up with:

Image

It says "2 selected" but drop down shows only one checked location. Search results show QSOs of more than just the initially selected location.

phl0:
Update: It seems that the current active station location is always also searched when using the date quick filter. But that location is not marked with the checkbox in the drop down list in the results. In the screenshot above the first location is the active one. Results also show QSOs matching the date filter from that location despite it not being selected.

phl0 (approved):
This works (correctly) for me now

PR #3125, opened 2026-03-24
[Winkey] Fixing dialog that will not close because of error

AndreasK79:
The "Macros saved" dialog could not be closed after editing macros.
Getmacros was not in the scope.

I did some fixes recently, and it broke other stuff, so a fix is needed.


phl0 (approved):
Fix confirmed. Brings winkeyer device back to life on pressing the buttons with the mouse.

PR #3137, opened 2026-03-31
Cqzone wrap

AndreasK79:
Solves #3127. Draws an extra set of cq zones to the left and to the right.

While looking into this, I discovered that Z1 and Z19 didn't align. I adjusted that, and found that Big Diomede was placed in Z1, instead of Z19. Adjusted borders for Z1 and Z19.

Added the same logic for the ITU zones too. Had to adjust Z26 to align with Z1 for Big/Little Diomede islands.


phl0 (approved):
works

PR #3144, opened 2026-04-01
Timeplotter adjustments

AndreasK79:
Added mode and heatmap to timeplotter.
Changed to only use current dxccs in the dropdown.
The graphs are now stacked, with QSOs per band.


phl0 (approved):
Nice stuff!

PR #3151, opened 2026-04-02
New UI for JCC award

cebarobot:
Hi, everyone~

This PR introduces a new UI for the JCC award, providing a more intuitive way to visualize the status and progress of QSOs across more than 800 cities.

In the new UI, each city is represented by a slot. Different style of slot means confirmed / worked / not worked. Clicking on confirmed / worked city could open the QSO list. Hovering on slot could show the full city number and name. All cities are grouped by prefecture.

Comparing to traditional city × band table, the new UI does not contain band info. Because JCC award does not consider band in score, this could be acceptable. In addition, band, mode and other endorsements could be checked with the filter above. The upcoming AJA award will still use city × band table, which is also useful for JCC. 

Also, this new UI is inspired by rdaward.ru.

image

### DIFF
--- a/application/helpers/awards_helper.php
+++ b/application/helpers/awards_helper.php
@@ -108,20 +108,18 @@ if (!function_exists('awards_render_jcc_grid_slot')) {
 	function awards_render_jcc_grid_slot($slot, $postdata) {
 		$classes = array(
 			'award-grid-slot',
-			'badge',
+			'btn',
 			'border',
 			'd-inline-flex',
 			'align-items-center',
 			'justify-content-center',
-			'fw-semibold',
-			'text-decoration-none',
 		);
 		if (($slot['status'] ?? '-') === 'C') {
-			$classes[] = 'text-bg-success';
+			$classes[] = 'btn-success';
 		} elseif (($slot['status'] ?? '-') === 'W') {
-			$classes[] = 'text-bg-danger';
+			$classes[] = 'btn-danger';
 		} else {
-			$classes[] = 'text-bg-light';
+			$classes[] = 'btn-light';
 		}
 		if (!empty($slot['deleted'])) {
 			$classes[] = 'award-grid-slot-deleted';
--- a/application/helpers/awards_helper.php
+++ b/application/helpers/awards_helper.php
@@ -96,3 +96,75 @@ if (!function_exists('awards_render_jcc_cell')) {
 		return '<div class="' . $class_name . '"><a href="' . html_escape($href) . '">' . $status . '</a></div>';
 	}
 }
+
+if (!function_exists('awards_render_jcc_grid_slot')) {
+	/**
+	 * Renders a slot for the grouped JCC demo grid.
+	 *
+	 * @param array $slot The slot metadata
+	 * @param array $postdata The postdata containing filter options
+	 * @return string The HTML string for the slot
+	 */
+	function awards_render_jcc_grid_slot($slot, $postdata) {
+		$classes = array(
+			'award-grid-slot',
+			'badge',
+			'border',
+			'd-inline-flex',
+			'align-items-center',
+			'justify-content-center',
+			'fw-semibold',
+			'text-decoration-none',
+		);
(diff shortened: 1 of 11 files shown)


HB9HIL:
Hello @cebarobot 

Why do you create a new model instead extending the existing JCC model?

int2001:
putting this one onHold.

many things to consider:
- optics. Doesn't follow our normal UI-Design. CSS inside view, instead of using the CSS which is there
- Huge JSON-Files delivered every time instead of using tables in database (e.g. primary_subdivisions)
- SQL: Why using "escape" instead of state-of-the-art prepared-statements? we invested huge effort to refactor everything to secure prepared statements
- ...

just things i noticed while having a quik look on this PR.

to be clear: i like the approach. and maybe JCC needs a different view/optic than before. but there are some things which tell me: It's mostly AI instead of understanding the codebase.

int2001:
I'll have a deeper look tomorrow.

In short:
- UI/Common Design: look at the other awards. Clean tables, no "pills", straight design. CSS in assets instead of defining them in single views.
- escape: we use prepared statements with bindings wherever possible instead of PHP/CI-Magic
- secondary_subdivisions: yes, go ahead. Good one.

int2001:
This wasn't an attack. We highly appreciate any kind of contribution.
But it needs to fit into the design as seen from UI AND Code.

- Prepared-Statements: Yes, sorry. my fault. they're prepared. good!
- Pills: understand with the mass of entities - perhaps it's possible to replace those pills with something which goes a bit more into the table-direction? perhaps a table with cols for the subdivisions instead of the pills?
- QSL-Filter: This is completly different from what has been there before. How do i filter on Entities i've worked without any confirmation? Looks like you dropped that filter completely.
- Prop_mode - yes was missing in old JCC as well. Please have a look at the other awards. we've some special logic for SAT there. i am missing "All HF (w/o SAT)" at Bands. would be a good chance to implement it. i as myself worked a lot of JA-Stations via SAT. At the moment the only way to see what has been wked on HF but not SAT is scrolling through every HF-Band one by one.
- Performance: Consider refactoring the nested SQL in Jcc_award_model L276ff - yes was also before, but may be worth a look because it may cause issues with large logs

i try to help you where it's possible. Going to dig and comment at the code, once you tell me: "Yes, ready for first deep review".

int2001:
ah, one more:
the subdivisions: one should also be able to log them. at the moment user can only log the primary ones.
so one reason more for putting them into database.

application/helpers/awards_helper.php
```diff
@@ -96,3 +96,73 @@ function awards_render_jcc_cell($entity, $band, $status, $postdata) {
...
+			'd-inline-flex',
+			'align-items-center',
+			'justify-content-center',
+		);
+		if (($slot['status'] ?? '-') === 'C') {
+			$classes[] = 'btn-success';
+		} elseif (($slot['status'] ?? '-') === 'W') {
+			$classes[] = 'btn-danger';
+		} else {
+			$classes[] = 'btn-light';
```
AndreasK79, line 122:
```suggestion (as this looks better in dark mode)
			$classes[] = 'btn-secondary';
```

application/views/awards/jcc/index.php
```diff
@@ -1,72 +1,76 @@
...
+        }
+    }
+
+    .award-grid-slot {
+        width: 3rem;
+        height: 2rem;
+    }
+
+    .award-grid-slot-deleted {
+        background-image: repeating-linear-gradient(135deg, rgba(0, 0, 0, 0.18) 0, rgba(0, 0, 0, 0.18) 2px, transparent 2px, transparent 6px);
```
AndreasK79, line 33:
```suggestion (to make this a bit more visible)
        background-image: repeating-linear-gradient(135deg, rgba(0, 0, 0, 0.5) 0, rgba(0, 0, 0, 0.5) 3px, transparent 3px, transparent 6px);
```

AndreasK79:
@cebarobot thanks for your work. I did look into the visuals here, and came with a few suggestions.

PR #3157, opened 2026-04-06
Add Sourcefilter to cluster

int2001:
Since DXClusterAPI may/can provide data from more than one cluster it is a good idea to chose which sources should be displayed in Bandlist.

Default is "All". List is automatically populated by the sources available on the DXClusterAPI configured.

image


phl0 (approved):
Works here.

PR #3158, opened 2026-04-07
Take station-locations qth instead of options

int2001:
The Map at DXCluster always showed the location of the instance instead of those from the user.

Reason was: Code tries to take it from user_options (WTF?) and falls back to location set in config.

This PR fixes it. Grid from station_location will be taken.


phl0 (approved):
works

PR #3162, opened 2026-04-08
Fix bandxuser handling

int2001:
Problemstatement:
- Create a new band as admin
- Try to log a QSO with it --> Works
- Login as a User which existed before and try to activate the band. no chance.

This one fixes it:
- left join instead of inner join for showing the bands
- upsert instead of insert when activating it

Side-Bugs fixed:
- Delete didn't remove the band (there was/is even a warning that it is removed completly - only the ref was deleted)
- Take care of security for award-columns

### DIFF
--- a/application/models/Bands.php
+++ b/application/models/Bands.php
@@ -99,7 +99,7 @@ class Bands extends CI_Model {
 
 	function get_all_bands_for_user() {
 		$sql = "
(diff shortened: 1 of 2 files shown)


phl0 (approved):
I can confirm that a newly created band can now be used and activated for different awards by a different/non-admin user.

PR #3164, opened 2026-04-13
Add DXCC-Lookup to 1st-login-wizard

int2001:
Add DXCC/ITU/CQ Lookup to 1st Login-Wizard.

Testing:
1. Create new User as admin
2. Login (incognitor or previously logged out) as that new User
3. Type in Call. If DXCC/ITU/CQ is deriveable all three fields are preselected/prefilled automatically

putted it into view and NOT extra-JS to avoid overhead. it neeeded a check if JQuery is already loaded.

### DIFF
--- a/application/views/user/modals/first_login_wizard.php
+++ b/application/views/user/modals/first_login_wizard.php
@@ -65,6 +65,7 @@
                                         <div class="row gx-2">
                                             <div class="col-md-6">
                                                 <select class="form-select" id="stationCQZoneInput" name="station_cqz" required>
+                                                    <option value=""></option>
                                                     <?php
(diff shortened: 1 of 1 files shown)


phl0:
Basically works. But changing/correcting callsigns does not reset zones. Please see https://github.com/int2001/wavelog/pull/27. This also adds an empty value so that users cannot simply go ahead with zones 1,1.

int2001:
> Basically works. But changing/correcting callsigns does not reset zones. Please see [int2001#27](https://github.com/int2001/wavelog/pull/27). This also adds an empty value so that users cannot simply go ahead with zones 1,1.

good finding!

phl0 (approved):

PR #3169, opened 2026-04-16
OQRS Delivery-Method

int2001:
Let User chose the offered delivery method for OQRS.

Old:
Requester can chose between direct and bureau, even if Callowner doesn't want to fullfill direct-OQRS.

New:
Callowner can chose what should be offered in settings:

image


phl0 (approved):
Works as designed.

PR #3180, opened 2026-04-19
Fix 13cm band designator

phl0:
As per https://wwrof.org/cabrillo/cabrillo-qso-data/ (which seems to be the official standard for cbr) the 13cm band should be exported as "2.3G" instead of "2.4G".


int2001 (approved):
TIL new things ;)

PR #3181, opened 2026-04-20
LBA: Batchedit for SIG / SIG_INFO + Display

int2001:
This one adds:
- Showing SIG_INFO in LBA as extra (optional) column
- Batchedit of SIG and SIG_INFO

**open**:
Leave SIG-Display as combined field in LBA or remove SIG_INFO, since both cols are shown now? @AndreasK79 

image


phl0 (approved):
works for me.

AndreasK79 (approved):

PR #3182, opened 2026-04-20
Fix empty SAT-Name treated as SAT-QSO

int2001:
Some Tools like the one mentioned at https://github.com/wavelog/wavelog/discussions/3177 send an empty string - catch it!

### DIFF
--- a/application/models/Cat.php
+++ b/application/models/Cat.php
@@ -10,7 +10,7 @@
 				$prop_mode = $result['prop_mode'];
 			// For backward compatibility, SatPC32 does not set propergation mode
 			} else if (isset($result['sat_name'])) {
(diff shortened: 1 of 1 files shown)


application/models/Cat.php
```diff
@@ -10,7 +10,11 @@ function update($result, $user_id, $operator) {
 				$prop_mode = $result['prop_mode'];
 			// For backward compatibility, SatPC32 does not set propergation mode
 			} else if (isset($result['sat_name'])) {
```
phl0:
```suggestion
			} else if (isset($result['sat_name']) && trim($result['sat_name']) != '') {
```

phl0: Maybe a little easier? :)

phl0 (approved):
This works fine

PR #3185, opened 2026-04-21
simplify radio tuning with database cat url (drop http/https hazzle)

HB9HIL:
The external cat url logic had a bug for other targets then our default `http://127.0.0.1:54321`. 
The line
```javascript
const tryHttps = !isHttps;
```
inverted the logic as it should be. Started the URL with https the browser first tried http and failed because of mixed content. 
We shouldn't play around here. If a user runs wavelog on http and enters an http cat url which isn't localhost we don't need to try https first. This just causes noise. 
I propose to drop the http/https (or maybe https/http/https?) logic and just use the url the user provided. Localhost will still work normally.

This PR comes with https://github.com/wavelog/WaveLogGate/pull/121 to work properly 

thx to @zone11 for the hint

### DIFF
--- a/assets/js/cat.js
+++ b/assets/js/cat.js
@@ -399,59 +399,20 @@ $(document).ready(function() {
         // Validate and normalize mode parameter
         const validModes = ['lsb', 'usb', 'cw', 'fm', 'am', 'rtty', 'pkt', 'dig', 'pktlsb', 'pktusb', 'pktfm'];
         const catMode = mode && validModes.includes(mode.toLowerCase()) ? mode.toLowerCase() : 'usb';
+        const requestUrl = catUrl + '/' + freqHz + '/' + catMode;
 
-        // Determine which protocol to try first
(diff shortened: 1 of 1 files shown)


int2001 (changes requested):
Had a first look into it.
in general a good idea to clean it up. But:

http is the "default"-handler for radios in Hardware-Interfaces. It was implemented because some OPs requested that feature. (Before it was http as well - but not configureable).

When using Safari (the reason we built in https, because Safari fails X-protocols) QSY doesn't work any longer, unless the op changes his radio-url to https.

Most of the people use WL/WLG "out of the Box".

So we either need a clever solution here (migration with update to https is not...) or we live with that noise. Was/Is there exactly for that reason.

Open for ideas. Usability has higher priority than "noise" which isn't noticed by a normal user.

HB9HIL:
https://github.com/HB9HIL/wavelog/pull/140 sounds legit and is merged. I'm happy with it

int2001 (approved):

PR #3199, opened 2026-05-01
Cache CAT-Data on 1st connect

int2001:
Problem:
When using WavelogGate (or any other websocket-CAT-provider):
- Navigating to `/qso` or `/bandmap/list` leads into connection to the Websocket (if chosen)
- Websocket sends welcome-message and 1st radio-info upon that connect
- Wavelog itself ignores the 1st radio-info, even when activating "CAT/WS" and Bandmap/QSO . 
- So User always has to turn VFO once, to force sync at bandmap or qso

This PR fixes that behaviour.
- The 1st info from websocket (if available) is cached and used (if not outdated)
- So User doesn't have to turn VFO if there's already fresh radiodata

Testinghints:
- Launch WavelogGate (or any other websocket-CAT-Tool) and make sure Gate has actual frequency/mode
- Navigate to `bandmap/list`, turn on CAT via websocket (or polling for testing the whole thing)
- Last QRG should be visible within WL

### DIFF
--- a/assets/js/cat.js
+++ b/assets/js/cat.js
@@ -221,6 +221,16 @@ $(document).ready(function() {
 
         // Handle radio status updates
         if (data.type === 'radio_status' && data.radio && ($(".radios option:selected").val() == 'ws')) {
+            // Calculate age from timestamp, defaulting to 0 (fresh) if timestamp is missing
+            if (data.timestamp) {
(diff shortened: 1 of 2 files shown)


HB9HIL (approved):
tested working

PR #3203, opened 2026-05-03
[Advanced Logbook] Duplicate query rewrite

AndreasK79:
It turns out that the existing query for finding duplicates in the Advanced Logbook contains a weakness.
It will not find a duplicate, if you have a third QSO with that station outside the duplicate window of 1800s. 
That is because of the usage of min/max for col_time_on.

Since we require MySQL8+, we can use windowing functions.


int2001 (approved):
looks good and works

Issue #3202, opened 2026-05-03
ADIF Export


AlbrechtDIETRICH:
### I have checked the following

- [x] I am running the latest version of Wavelog
- [x] I have cleared my browser cache
- [x] This is a bug, not a setup question or feature request

### Bug Description

The ADIF Export is not at all compliant with ADIF Definitions.

- WaveLog exports field not being in the ADIF spec but does not declare them in the ADIF Header to be **USEDEF** as required

- When exporting ADIF **NOT all Fields** used by WaveLog **are exporte**d. So the ADIF export is not suitable for backup (as described in the help): One will use an bunch of information when he plays back a WaveLog created ADIF after let's say a Docker container crash.

Here in use:
PRETTY_NAME="Ubuntu 24.04.4 LTS"
VERSION_ID="24.04"
VERSION="24.04.4 LTS (Noble Numbat)"
WaveLog="2.4.1" running in a Docker container

### Steps to Reproduce

Just export ADI and let it ckeck by an ADIF validator or consider the ADIF specs.

### Expected Behavior

Correct ADIF Export 100 % compliant with ADIF specs and ALL fields used by WaveLog.

### Wavelog Version

2.4.1

### PHP Version

8.4

### MySQL/MariaDB Version

MySQL

### Server OS

Kubuntu 24.04.4 LTS (Noble Numbat)

### Browser

Brave

### Device Type

Desktop

HB9HIL:
Hello @AlbrechtDIETRICH 

1. Please list all ADIF fields which are exported by Wavelog but not declared in ADIF Standard
2. Please list the fields which are actively used by Wavelog but not exported in ADIF.

HB9HIL:
Also please provide your `docker-compose.yml`

AlbrechtDIETRICH:
These are the fields WaveLog is using but they are not in the ADIF Specs:
_modified
BIOGRAPHY
CLUBLOG_QSO_DOWNLOAD_DATE
CLUBLOG_QSO_DOWNLOAD_STATUS
CNTY_ALT
DCL_QSL_RCVD
DCL_QSL_SENT
DCL_QSLRDATE
DCL_QSLSDATE
EQSL_AG
EQSL_STATUS
HEADING
ion_id
LOTW_STATUS
MORSE_KEY_INFO
MORSE_KEY_TYPE
MY_CNTY_ALT
MY_DARC_DOK
MY_POSTCODE_INTL
[...]

PR #3205, opened 2026-05-03
only export COUNTRY in ADIF if DXCC is not 0

HB9HIL:
the only issue I could find in #3202 is that we export `- None - (E.g. /Mm, /Am)` as the `COUNTRY` field when DXCC is `0`. This can be considered as bug and should be fixed. 

Before:
If COL_DXCC is `0` the ADIF Line for country was: `<COUNTRY:24>- None - (E.g. /Mm, /Am)`

After:
If COL_DXCC is `0` we omit `COUNTRY`
otherwise we export normally


phl0 (approved):
Thats a very valid find and fix works for me.

PR #3206, opened 2026-05-04
Skip to Clubstation - First Login Wizard

HadleySo:
In the wizard, allow users to skip directly to clubstations they have permissions to.

Add a button / dropdown to First Login Wizard to allow users to skip setting up a station location if they have a clubstation. If they have no clubstations, UI does not change.

Wizard will continue to show on future logins

<details>
  <summary>Show screenshots</summary>
firstLoginWizard_skip_single
firstLoginWizard_skip_multiple
</details>

### DIFF
--- a/application/views/user/modals/first_login_wizard.php
+++ b/application/views/user/modals/first_login_wizard.php
@@ -101,6 +101,43 @@
                     </div>
             </div>
             <div class="modal-footer">
+
+                <!-- Show more buttons if Clubstation enabled and user has rights -->
+                <?php if ($this->config->item('special_callsign') && !empty($this->session->userdata('available_clubstations'))) { ?>  
+                    <?php if (count($this->session->userdata('available_clubstations')) > 1) { ?> 
+
+                        <!-- If user has more than 1 clubstation, use dropdown -->
+                        <div class="dropdown">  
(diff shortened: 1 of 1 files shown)


application/views/user/modals/first_login_wizard.php
```diff
@@ -101,6 +101,43 @@
                     </div>
             </div>
             <div class="modal-footer">
+
+                <!-- Show more buttons if Clubstation enabled and user has rights -->
+                <?php if ($this->config->item('special_callsign') && !empty($this->session->userdata('available_clubstations'))) { ?>
```
int2001:
```suggestion
                <?php if ($this->config->item('special_callsign') && is_array($this->session->userdata('available_clubstations'))) { ?>
```

see https://github.com/wavelog/wavelog/blob/e32790ee3bd5dc6a6ffea73c1f298dae0c98280d/application/models/User_model.php#L634 and https://github.com/wavelog/wavelog/blob/e32790ee3bd5dc6a6ffea73c1f298dae0c98280d/application/models/User_model.php#L659 - available_clubstations could also be a string instead of an array.

int2001 (changes requested):
See Changerequest.
Rest works and help Users which are only using WL for clubcalls. Well done

application/views/user/modals/first_login_wizard.php
HadleySo: Thanks for the review and feedback!

int2001 (approved):

PR #3212, opened 2026-05-06
Add ADIF 3.1.3 prop modes

phl0:
Seems we missed to add the two prop modes introduced with ADIF 3.1.3. See https://github.com/wavelog/wavelog/discussions/3210


int2001 (approved):
Seems legit :)

PR #3222, opened 2026-05-08
Centralize ADIF propagation modes

phl0:
Working on https://github.com/wavelog/wavelog/pull/3212 we discovered that the ADIF prop modes are scattered across the whole code and in many places. This is the attempt to centralize that to make updating and maintaining easier.


application/views/view_log/partial/log_ajax.php
```diff
@@ -50,71 +51,7 @@ function echo_table_col($row, $name) {
...
-					echo '<td>' . _pgettext("Propagation Mode","Trans-equatorial") . '</td>';
-					break;
-				case 'TR':
-					echo '<td>' . _pgettext("Propagation Mode","Tropospheric ducting") . '</td>';
-					break;
-				default:
-					echo '<td>' . htmlspecialchars($row->COL_PROP_MODE ?? '') . '</td>';
-					break;
-				}
+				echo '<td>' . _pgettext("Propagation Mode", $adif_propmodes[$row->COL_PROP_MODE]) . '</td>';
```
int2001: what if someone uploads an unknown prop_mode?
```suggestion
				echo '<td>' . _pgettext("Propagation Mode", $adif_propmodes[$row->COL_PROP_MODE] ?? $row->COL_PROP_MODE) . '</td>';
```

int2001: same at application/views/view_log/qso.php L249

application/config/adif_propmodes.php
```diff
@@ -0,0 +1,29 @@
...
+   "INTERNET" => __("Internet-assisted"),
+   "ION" => __("Ionoscatter"),
+   "IRL" => __("IRLP"),
+   "LOS" => __("Line of Sight (includes transmission through obstacles such as walls)"),
+   "MS" => __("Meteor scatter"),
+   "RPT" => __("Terrestrial or atmospheric repeater or transponder"),
+   "RS" => __("Rain scatter"),
+   "SAT" => __("Satellite"),
+   "TEP" => __("Trans-equatorial"),
+   "TR" => __("Tropospheric ducting")
```
HB9HIL: The translations are already existing but were referenced by context. Without this context everything needs to be translated again which is not necessary. 


```suggestion
   "AS" => _pgettext("Propagation Mode", "Aircraft Scatter"),
   "AUR" => _pgettext("Propagation Mode", "Aurora"),
   "AUE" => _pgettext("Propagation Mode", "Aurora-E"),
   "BS" => _pgettext("Propagation Mode", "Back scatter"),
   "ECH" => _pgettext("Propagation Mode", "EchoLink"),
   "EME" => _pgettext("Propagation Mode", "Earth-Moon-Earth"),
   "ES" => _pgettext("Propagation Mode", "Sporadic E"),
   "FAI" => _pgettext("Propagation Mode", "Field Aligned Irregularities"),
   "F2" => _pgettext("Propagation Mode", "F2 Reflection"),
   "GWAVE" => _pgettext("Propagation Mode", "Ground Wave"),
   "INTERNET" => _pgettext("Propagation Mode", "Internet-assisted"),
   "ION" => _pgettext("Propagation Mode", "Ionoscatter"),
   "IRL" => _pgettext("Propagation Mode", "IRLP"),
   "LOS" => _pgettext("Propagation Mode", "Line of Sight (includes transmission through obstacles such as walls)"),
   "MS" => _pgettext("Propagation Mode", "Meteor scatter"),
   "RPT" => _pgettext("Propagation Mode", "Terrestrial or atmospheric repeater or transponder"),
   "RS" => _pgettext("Propagation Mode", "Rain scatter"),
   "SAT" => _pgettext("Propagation Mode", "Satellite"),
   "TEP" => _pgettext("Propagation Mode", "Trans-equatorial"),
   "TR" => _pgettext("Propagation Mode", "Tropospheric ducting")
```

HB9HIL (changes requested):

application/views/view_log/partial/log_ajax.php
phl0: > same at application/views/view_log/qso.php L249

Good idea. But need to fix empty prop modes as well. QRX

application/config/adif_propmodes.php
int2001: Aren't they needed at least somewhere (with correct context) in codebase, that gettext/weblate knows they're existing?

Only pulling them from database and passing them to pgettext won't work (in my mind), because the texts appear only during runtime, but not in code...

HB9HIL: This code here ist placed in a new config file called `application/config/adif_propmodes.php`. This was introduced by @phl0. This is enough for gettext as the array gets created when the file is called in the session. This array then maps the DB value to the correct translation.

phl0: Ok so all correct now

int2001 (approved):
checked only technical - can't spot an issue here from that perspective.
good one - brings more flexibility

HB9HIL (approved):

PR #3224, opened 2026-05-08
Fix missing DCL-Cols

int2001:
DCL-Cols where missing at Timeline, which produced a warning because of not-defined variables


phl0 (approved):
Fixes the bug.

phl0:
image

Distances has the same issue ...

phl0:
> Distances has the same issue ...

Fixed with: https://github.com/wavelog/wavelog/pull/3224/commits/dc285ae2643223f0aba64070cfb086ad2d349ac2

Issue #1439, opened 2025-01-02
Callbook should be a User setting


KidVizious:
I am running a Wavelog server and wanted to provide access to some of my friends. Unfortunately, since the callbook credentials are global, all of their lookups will also be using my Callbook credentials.

Callbook login creds should be moved into user settings so that each user can use their own accounts for callbook lookups.

HB9HIL:
Since this is only for the Callbook Lookup and has no affect to your user account,
why does it matter?....

KidVizious:
It does have an impact on my user account.  Assume I'm not a paid QRZ subscriber, then I'm limited in the number of lookups per day. If multiple people are using my login for lookup, that limit will be reached quite quickly. 

And even if I am a paid subscriber, it just seems to me that in general requiring a user to use personal credentials as a global server setting isn't best practice.

HB9HIL:
> It does have an impact on my user account. Assume I'm not a paid QRZ subscriber, then I'm limited in the number of lookups per day. If multiple people are using my login for lookup, that limit will be reached quite quickly.


If you don't have a paid subscription at QRZ.com this will also reduce the amount of data you get back from QRZ.com. For example you won't get gridsquares.  

> And even if I am a paid subscriber, it just seems to me that in general requiring a user to use personal credentials as a global server setting isn't best practice.

Well it's up to you to share your Wavelog Instance with other users. You also could set up hamqth or qrzcq as a callbook source. 

In my eyes it's much much easier and better practice to purchase the subscription and share the costs among the users (if the money is an issue) than to force every user to use a paid subscription individually if they want to full joy of qrz.com data...

int2001:
Full ACK to  @HB9HIL's argumentation.

We have discussed this feature request in our team and have come to the conclusion that it will **not be implemented** in wavelog.

Lets summarize it - You've a couple of options:
1.) Purchase a subscription.
2.) Switch over to another callbook-provider like hamqth.
3.) Stop sharing your own subscription and leave it blank. - nobody forces you to share your instance.

However. Many thanks for the idea. We're happy about every feedback. But not every wish can be fullfilled.

If you're extremly uncomfortable with this decision, you can still add the feature on your own to your instance. That's one of the benefits a FOSS-Software allows.

PR #3233, opened 2026-05-12
Add QRZCALL.EU as a callbook provider

Ron6519:
## Add QRZCALL.EU as a callbook provider

**[QRZCALL.EU](https://qrzcall.eu/)** is a QRZ-compatible amateur-radio callsign database. This PR adds it as a new `'qrzcall'` callbook source alongside the existing `qrz`, `qrzcq`, `hamqth`, and `qrzru` providers.

### What's new

| File | Change |
|---|---|
| `application/libraries/Qrzcall.php` | **new** — ~210 lines, mirrors `Qrz.php` (same method signatures: `session`, `set_session`, `search`, `sourcename`) |
| `application/libraries/Callbook.php` | adds `case 'qrzcall'` in `queryCallbook()`, `_qrzcall()` private method, `QRZCALL_SESSION_DURATION` constant, and the matching session-cache key |
| `application/config/config.sample.php` | adds `qrzcall_username` / `qrzcall_password` defaults + inline docs |
| `install/index.php` | adds **QRZCALL.EU** to the install-wizard callbook dropdown and extends the tooltip |
| `install/config/config.php` | adds `%qrzcall_username%` / `%qrzcall_password%` placeholders to the wizard's template config |
| `install/includes/core/core_class.php` | adds `'qrzcall'` to the `$callbooks` whitelist so credentials are substituted |

### Why

QRZCALL.EU is a growing alternative to QRZ.com, particularly popular in Europe. Adding it gives Wavelog users another option (or a fallback in the `['qrzcall', 'qrz', 'hamqth']` array form already supported by Wavelog).

### Behaviour

- **Auth**: callsign + password → JWT (cached 7h, just below the 8h token TTL). The bearer pattern is hidden behind the existing "session key" abstraction, so the Callbook orchestrator's retry-on-expiry and `/portable` reduced-mode fallback paths work unchanged.
- **Access tier**: requires a Data or Extra subscription on QRZCALL.EU. The XML endpoint returns 401 otherwise.
- **XML schema**: identical field names to QRZ.com (`call`, `fname`, `name`, `addr1`, `state`, `country`, `dxcc`, `grid`, `class`, `email`, `lotw`, `eqsl`, …) → no downstream changes to the QSO entry form, ADIF, stats, etc.

### How I tested

- Built a docker-compose stack (MariaDB + this branch) and ran the full install wizard against it
- Selected `QRZCALL.EU` in the wizard, entered live QRZCALL.EU Data-tier credentials, completed install
- From New QSO, looked up several callsigns — fields populate correctly (name, QTH, grid, DXCC, IOTA where set, etc.)
- Wiped `application/cache/*` between runs to force re-auth — confirmed JWT round-trip via `application/logs/`
- Smoke-tested the array form `$config['callbook'] = ['qrzcall', 'qrz', 'hamqth']` — confirmed fallthrough
- Verified `Callbook::_validate_sessionkey()` accepts QRZCALL JWTs (well over 10 chars)
- All five PHP files pass `php -l`

No new dependencies — pure PHP + cURL, like the existing providers.

### Open for discussion

1. **Session duration** — I chose 7h (one hour below the 8h JWT TTL). Existing providers use 4h–55min. Happy to align if there's a preferred policy.
2. **Config key naming** — used `qrzcall_username` to match the established `qrz_*` / `qrzcq_*` / `qrzru_*` snake-case-without-underscore-tld pattern. Easy to rename to e.g. `qrzcalleu_*` if preferred.
3. **Reduced-mode trigger** — the library treats any error containing `"not found"` as the trigger to retry the base callsign in reduced mode (consistent with what the upstream service emits). Existing providers use case-sensitive `"Not found"`. If you'd rather have a stable error contract, I can emit a sentinel error code instead.

### Future work (separate PRs, not in this one)

- Per-user callbook credentials, tracked in #1439 — happy to migrate this provider when the broader change lands
- QRZCALL.EU **logbook upload** as a separate ADIF-over-HTTP provider — QRZCALL.EU has an ADIF import endpoint we could target

### Maintainer / 73

PA4R — happy to address review comments, ship follow-ups, or migrate to a different auth scheme if the team prefers a different approach.


Ron6519:
### Test account for reviewers 👋

If you want to verify the lookup live without registering on QRZCALL.EU yourself, feel free to use this throw-away Data-tier account:

```
Callsign:  TEST0WVLG
Password:  WaveLogTest_2026!Demo
```

- Data subscription is active until **2099-01-01**, so it won't expire mid-review.
- The bound callsign profile (`TEST0WVLG`) is intentionally `HIDE`-flagged so it doesn't pollute public search results, but it's reachable by direct lookup.

#### Quick sanity check from a shell (no Wavelog needed)

```bash
# 1) Issue a JWT
TOKEN=$(curl -s -X POST https://api.qrzcall.eu/v1/auth/login.php \
  -H 'Content-Type: application/json' \
  -d '{"callsign":"TEST0WVLG","password":"WaveLogTest_2026!Demo"}' \
  | python3 -c "import sys, json; print(json.load(sys.stdin)['token'])")

# 2) Look up any real callsign
curl -s "https://api.qrzcall.eu/v1/pub/callsign_xml.php?callsign=PA4R" \
  -H "Authorization: Bearer $TOKEN"
```

#### Inside Wavelog

After applying this PR and running through the install wizard:

- **Step 3 — Configuration** → set callbook to **QRZCALL.EU**, username `TEST0WVLG`, password `WaveLogTest_2026!Demo`
- After install, **Logbook → New QSO** → type any callsign in the database (e.g. `PA4R`) → fields auto-populate

Happy to rotate or revoke this account on request. 73!

HB9HIL:
Please base pull requests always on `dev` branch. We don't merge directly to master

Ron6519:
### Updated: switched to Personal Access Tokens (commit `0300047`)

Since this PR hadn't been reviewed yet, I took the chance to modernize the auth pattern before merge. The previous commit had Wavelog log in with the user's QRZCALL.EU **callsign + password** and cache the resulting JWT — same shape as the QRZ.com provider. After thinking about it more, that pattern made me uncomfortable: a leaked Wavelog config file would expose the user's password (which they may reuse elsewhere), and rotating the password would break every logger at once.

The new commit switches to a single long-lived **Personal Access Token** the user generates in the QRZCALL.EU SPA (My Profile → Account → API Tokens). This matches the GitHub / Stripe / Cloudflare pattern: each Wavelog install gets its own revocable `pat_…` token, and the user's account password never touches the logger.

#### What's different in the diff

- `application/libraries/Qrzcall.php` — drops `session()` and `set_session()`. `search()` just sends `Authorization: Bearer pat_…` directly. **~30 lines shorter.**
- `application/libraries/Callbook.php` — `QRZCALL_SESSION_DURATION` constant and `qrzcall_session_cachekey` field removed. The `_qrzcall()` private method collapses from ~40 to ~20 lines (no cache, no retry-on-expiry).
- `application/config/config.sample.php` — `qrzcall_username` / `qrzcall_password` replaced by a single `qrzcall_token`.
- `install/index.php` — the wizard now shows a single "API Token" input when QRZCALL.EU is selected, with a hint link pointing the user to the token UI. Other providers continue to show the username/password row.
- `install/includes/core/core_class.php` — the substitution logic now distinguishes username/password-style providers (`qrz/hamqth/qrzcq/qrzru`) from token-style providers (`qrzcall`). Cleaner extension point for any future Bearer-only callbook.
- `install/config/config.php` — `%qrzcall_token%` placeholder replaces the username/password pair.

Net diff: **+168 / −156** across 6 files (same files as before; this commit is purely additive on top of the first).

#### How to verify locally

Fresh clone of this branch → apply patches → install Wavelog → in the install wizard, pick **QRZCALL.EU**, paste the test PAT below, finish install → New QSO → type any callsign.

Test PAT (TEST0WVLG, Data tier through 2099-01-01):

```
pat_b3gvr7n6wyzjvs9xk4jmpgxm3r3gtgdx
```

Or curl directly (single header, no JWT dance):

```bash
curl -s "https://api.qrzcall.eu/v1/pub/callsign_xml.php?callsign=PA4R" \
     -H "User-Agent: Mozilla/5.0" \
     -H "Authorization: Bearer pat_b3gvr7n6wyzjvs9xk4jmpgxm3r3gtgdx"
```

> *Heads-up:* the host now rejects the default `curl/X.Y` user-agent at the WAF (returns 415 from openresty). Wavelog's own `User-Agent: Wavelog/<version>` is fine; this is only relevant for manual `curl` testing.

#### Backward compatibility

The previous JWT flow (`POST /v1/auth/login.php` → JWT) still works on the server — this PR just moves Wavelog to the recommended pattern. Anyone who already integrated against the JWT endpoint is unaffected.

#### What didn't change

- Library output shape — `search()` still returns the same associative array, so the QSO entry form and everything downstream needs no changes.
- All other callbook providers — `Qrz.php`, `Hamqth.php`, `Qrzcq.php`, `Qrzru.php` and their dispatch in `Callbook.php` are untouched.
- The "callbook" config key and the `['callbook' => 'qrzcall']` selection mechanism.

Happy to squash the two commits into one before merge if you'd prefer a single-commit PR — just say the word.

73, PA4R

HB9HIL:
Hello @Ron6519 

I'm gonna review this by the end of the week 👍

AndreasK79:
@Ron6519 thanks for the work.
Tested working in QSO entry and the advanced logbook. 
Installer also works!

install/index.php
```diff
@@ -398,15 +398,18 @@ function convertToBytes(string $value): int {
...
+												<label for="global_call_lookup" class="form-label"><?= __("Optional: Global Callbook Lookup"); ?><i id="callbook_tooltip" data-bs-toggle="tooltip" data-bs-placement="top" class="fas fa-question-circle text-muted ms-2" data-bs-custom-class="custom-tooltip" data-bs-html="true" data-bs-title="<?= __("This configuration is optional. The callsign lookup will be available for all users of this installation. You can choose between QRZ.com, HamQTH, QRZCQ, QRZ.ru and QRZCALL.EU. While HamQTH also works without username and password, you will need credentials for the other providers. QRZ.com needs an XML subscription for the Call Locator; QRZCALL.EU needs a Data or Extra subscription on https://qrzcall.eu/. HamQTH does not always provide the locator information."); ?>"></i></label>
```
HB9HIL: Please don't use URL's in translations

```suggestion
												<label for="global_call_lookup" class="form-label"><?= sprintf(__("Optional: Global Callbook Lookup"); ?><i id="callbook_tooltip" data-bs-toggle="tooltip" data-bs-placement="top" class="fas fa-question-circle text-muted ms-2" data-bs-custom-class="custom-tooltip" data-bs-html="true" data-bs-title="<?= __("This configuration is optional. The callsign lookup will be available for all users of this installation. You can choose between QRZ.com, HamQTH, QRZCQ, QRZ.ru and QRZCALL.EU. While HamQTH also works without username and password, you will need credentials for the other providers. QRZ.com needs an XML subscription for the Call Locator; QRZCALL.EU needs a Data or Extra subscription on %s. HamQTH does not always provide the locator information."), "https://qrzcall.eu/"); ?>"></i></label>
```

HB9HIL (changes requested):
Looks really good. Only one thing in the translation in the installer UI. We don't want to use URL's in translations. Use `sprintf()` and `%s` instead.

HB9HIL (approved):

HB9HIL:
Good Job! Thx for Contribution

PR #3236, opened 2026-05-15
Remove duplicate mode/submode entries from adif_modes

kc9uhi:
Migration 041 created the initial entry of PSK63F in the adif_modes table with active=0.
Migration 094 created a duplicate entry of PSK63F in the adif_modes table, with active=1.
Migration 231 created unique index of the adif_modes table allowing only for one mode,submode,qrgmode,active combination. This prevented PSK63F from being deactivated from the Modes list as two identical mode,submode,qrgmode,active entries would exist.

This migration checks for two entries of PSK63F, one active=0 and one active=1, and removes the active=0 entry. The effective mode list is unchanged, but allows users to deactivate PSK63F from the list if desired.

install/assets/install.sql is updated to remove the duplicate PSK63F entry to not create this issue in new installations.


application/migrations/279_remove_duplicate_psk63f.php
```diff
@@ -0,0 +1,31 @@
+<?php
+defined('BASEPATH') or exit('No direct script access allowed');
+
+class Migration_remove_duplicate_psk63f extends CI_Migration
+{
+	public function up()
+	{
+		$db_result = $this->dbtry("SELECT submode, COUNT(*) AS total_rows, SUM(CASE WHEN active = 1 THEN 1 ELSE 0 END) AS active_rows, SUM(CASE WHEN active = 0 THEN 1 ELSE 0 END) AS inactive_rows FROM adif_modes WHERE mode='PSK' AND submode = 'PSK63F' AND qrgmode='DATA' GROUP BY submode HAVING COUNT(*) = 2 AND SUM(CASE WHEN active = 1 THEN 1 ELSE 0 END) = 1 AND SUM(CASE WHEN active = 0 THEN 1 ELSE 0 END) = 1;");
+		if($db_result->num_rows() > 0) {
+			if(($db_result->active_rows == 1) && ($db_result->inactive_rows == 1)) {
```
AndreasK79: $row = $db_result->row();
			if(($row->active_rows == 1) && ($row->inactive_rows == 1)) {

AndreasK79: @kc9uhi nice find. I tried to suggest a fix you could commit, but obviously didn't manage that :)
The migration fails, so you need a small modification for it to work.

HB9HIL:
@AndreasK79 we should check if we can include this migration in the next release mig

kc9uhi:
Further research shows that MySQL allows for multiple _NULL_ values in a unique index. This allows for the undesired duplication of mode,qrgmode entries with _NULL_ value submode. Current unique index constraint includes the \`active\` column, which also allows for duplicate mode,submode,qrgmode rows (given \`active\` is different) preventing activation/inactivation unless one is deleted.

Fixing this is handled in multiple steps:

1) Find duplicate mode,submode,qrgmode entries, removing the active=0 entry.
2) Add generated \`submode_strnull\` column to the table, which converts _NULL_ value to the string 'NULL'
3) Alter unique index to reference \`submode_strnull\` and remove reference to \`active\`. This enforces exactly one mode,submode,qrgmode row in the \`adif_modes\` table, regardless of active=/[0,1]/.

AndreasK79 (approved):

HB9HIL:
@kc9uhi Thanks for this one and good find. However it seems this needs some more discussions about legacy support (our README says we support MariaDB > 10.2, I currently struggle to find some information about indexing virtual columns).

We will take care of this but not in the next release (which is planned today 😄 )

sorry for the delay

kc9uhi:
Virtual column indexes in MariaDB was first supported in 10.2.1, an alpha release of the 10.2 series. First 'stable' release of 10.2 was 10.2.6. Perhaps the README should be updated to reflect the more specific stable release?

HB9HIL:
@kc9uhi we discussed in team and will bring this change with the next major release. Thx for your patience

int2001 (approved):
fixed and fine.

tbh: very special thing. need approve from @phl0 as well, since he has the pickiest DB of us

phl0:
> fixed and fine.
> 
> tbh: very special thing. need approve from @phl0 as well, since he has the pickiest DB of us

Tested and works on my system. PSK63F is removed as desired.

phl0 (approved):

PR #3237, opened 2026-05-15
[DXCC Award] Adds progress bards for bands/modes/continents

AndreasK79:
This PR adds progress bars for the DXCC award.

The card is collapsable.

I added ordering on the main dxcc table. In order for this to work, I needed to use the correct tag in the table header.
That means TD -> TH. Since the same JS was used for WAE, I include the fix for the view there too.


assets/js/sections/dxccmap.js
```diff
@@ -300,3 +300,57 @@ function onClick(e) {
...
+            Africa: +$('#Africa').prop('checked'),
+            Asia: +$('#Asia').prop('checked'),
+            Europe: +$('#Europe').prop('checked'),
+            NorthAmerica: +$('#NorthAmerica').prop('checked'),
+            SouthAmerica: +$('#SouthAmerica').prop('checked'),
+            Oceania: +$('#Oceania').prop('checked'),
+            Antarctica: +$('#Antarctica').prop('checked'),
+            sat: $("#sats").val(),
+            orbit: $("#orbits").val(),
+			dateFrom: $('#dateFrom').val(),
```
int2001: sat / orbit vs. sat**s**, orbit**s** ?
see:
https://github.com/AndreasK79/wavelog/blob/aee1a5d3804ef897ec05213a34f89a94c130f549/application/controllers/Awards.php#L242

assets/js/sections/dxccmap.js
```diff
@@ -300,3 +300,57 @@ function onClick(e) {
...
+        data: {
+            band: $('#band2').val(),
+            mode: $('#mode').val(),
+            worked: +$('#worked').prop('checked'),
+            confirmed: +$('#confirmed').prop('checked'),
+            notworked: +$('#notworked').prop('checked'),
+            qsl: +$('#qsl').prop('checked'),
+            lotw: +$('#lotw').prop('checked'),
+            qrz: +$('#qrz').prop('checked'),
+            eqsl: +$('#eqsl').prop('checked'),
```
int2001: clublog? gone?

application/models/Dxcc.php
```diff
@@ -196,6 +274,30 @@ function get_dxcc_array($dxccArray, $bands, $postdata, $location_list, $map = fa
...
+								$modeCounters['confirmed'][$mode][$dxccKey] = true;
+							}
+						}
+
+						// Track by mode and band (for detailed breakdown)
+						if (isset($dxcc->mode)) {
+							$mode = $dxcc->mode;
+							if (isset($modeBandCounters['confirmed'][$dxcc->col_band]) && !isset($modeBandCounters['confirmed'][$dxcc->col_band][$mode][$dxccKey])) {
+								$modeBandCounters['confirmed'][$dxcc->col_band][$mode][$dxccKey] = true;
+							}
```
int2001: sure about this? as far as i can see the mode-progress bars are populated by mode_progress() → getModeDxccTotals(). this calculation seems to be unsused.

could be that i am wrong here...

application/views/awards/dxcc/index.php
```diff
@@ -256,6 +259,162 @@
...
 
+	<div class="tab-pane fade" id="progress" role="tabpanel" aria-labelledby="progress-tab">
+    <?php
+    $i = 1;
+    if ($dxcc_array) {
+    	// Progress summary
+    	// Calculate total DXCC entities from the actual array (respects include deleted filter)
+    	$total_dxcc_entities = count($dxcc_array);
+
+    	echo '<div class="card mb-3">';
```
int2001: may be a stupid question:
why `echo` them explicitly? it's php...

int2001 (changes requested):
works, but a found a few strange things.

assets/js/sections/dxccmap.js
AndreasK79: Seems comment didn't transfer from vscode. The form has these in plural, so nothing to fix.

assets/js/sections/dxccmap.js
AndreasK79: You found an older bug. This was copied from load_dxcc_map, and it was missing there too :) Fixed now.

application/models/Dxcc.php
AndreasK79: You are right. I missed this while cleaning up. All code for mode progress was moved when it was rewritten for mode_progress.

application/views/awards/dxcc/index.php
AndreasK79: Well, no reason. Removed a lot of it.

phl0 (approved):

int2001 (approved):
works great now

phl0 (approved):
Works fine now in my picky db setup :-)

PR #3239, opened 2026-05-16
Reset propmode when coming from cluster

int2001:
When working SAT with SAT, SAT_MODE, PROP_MODE entered, Wavelog remembers those informations for the next QSO.
Which is okay and wanted.

But: When working mixed mode with coming from DXCluster, SAT & Co. are still in QSO-Form, even if cluster is HF-Only.

This patch fixes that behaviour. When coming from cluster prop, name and mode are emptied


HB9HIL (approved):
lgtm

PR #3243, opened 2026-05-18
Fix fatal parse error in installer callbook label

Ron6519:
## Problem

The install wizard is currently **completely broken** on `dev`. A `sprintf(`
ended up on the wrong `__()` in the **Optional: Global Callbook Lookup** label
(`install/index.php` line 401), producing invalid PHP. Opening the installer
shows only:

```
Parse error: syntax error, unexpected token ";", expecting ")" in install/index.php on line 401
```

Two things were wrong on that line:

- the **label** `__()` got wrapped in an unclosed `sprintf(` — but the label
  has no placeholder, so it must not be wrapped at all;
- the **tooltip** `__()` (which *does* contain `%s`) lost its `sprintf(`,
  leaving an orphan `, "https://qrzcall.eu/")` argument.

## Fix

Move the `sprintf(` back onto the tooltip `__()` and leave the label as a
plain `__()`:

```php
<?= __("Optional: Global Callbook Lookup"); ?> ... data-bs-title="<?= sprintf(__("… %s …"), "https://qrzcall.eu/"); ?>"
```

One-line change. `php -l install/index.php` is clean and the installer renders
again.


HB9HIL (approved):
upps 😆

PR #3246, opened 2026-05-20
add support for commented CBR files in CBR import

DB4SCW:
Some programs create commented CBR files (like fldigi in some circumstances).
With this PR, the parser now skips these lines completely instead of treating them like a header.


int2001 (approved):
sounds legit

phl0 (approved):

PR #3248, opened 2026-05-20
format dates on LotW dashboard according to user preferences

DB4SCW:
This PR is just a small optic change.

It changes the display format of dates in the LotW dashboard according to the user preferences instead of statically pulling from the config.


int2001 (approved):
works. good one

PR #3251, opened 2026-05-20
validate lotw url and check for http 200

HB9HIL:


application/controllers/Lotw.php
```diff
@@ -691,6 +692,17 @@ private function loadFromFile($filepath, $station_ids, $display_view = "TRUE") {
...
 	}
 
+	/**
+	 * Helper function to validate lotw url
+	 */
+	private function _validate_url(string $url) {
+		$scheme = strtolower(parse_url($url, PHP_URL_SCHEME) ?? '');
+		if ($scheme !== 'https') {
+			log_message('error', 'LoTW URL rejected – disallowed scheme: '.$scheme);
+			show_error('LoTW download URL is invalid (only http/https allowed).', 500);
```
int2001: ? we're checking for https. why mention http only?

int2001 (changes requested):
only little cosmetics.
rest is fine and works as "safety net"

application/controllers/Lotw.php
HB9HIL: was leftover

int2001 (approved):

phl0 (approved):
tested and works as expected

PR #3252, opened 2026-05-21
[Advanced Logbook] Added some help text for batch edit, and restyled help

AndreasK79:
image
image


application/views/logbookadvanced/edit.php
```diff
@@ -1,3 +1,7 @@
+<div class="alert alert-info" role="alert">
+	<i class="fas fa-exclamation-circle me-2"></i>
+	<?= __("Not all columns are available for batch editing. If you can't find the column you are looking for, use the regular QSO edit (click the callsign, then click the edit QSO button)."); ?></p>
```
int2001: that `</p>` at the end... where is the opening? can't spot it

int2001 (changes requested):
Minor one (see comment).

application/views/logbookadvanced/edit.php
AndreasK79: Thanks. Copy/pasted and missed that one.

int2001 (approved):

PR #3254, opened 2026-05-21
Amsat rover award

int2001:
Introduces the AMSAT-Rover-Award, described here: https://www.amsat.org/amsat-rover-award/

Since the process for application is highly manual, the award page supports the application by drafting the needed things as an email-template.

Screenshot:

image


phl0 (approved):
Great work.

PR #3258, opened 2026-05-23
Oqrs fix invalid slug

int2001:
when calling an invalid slug (e.g. wavelog.url/oqrs/doesntexist) there are PHP-warnings, because of not catched null-values.

this one fixes it.


AndreasK79 (approved):

AndreasK79:
@int2001 nice find, and looks a lot better without the error message.

PR #3261, opened 2026-05-25
[DXCC Award] Mode progress doesn't filter correctly if a band is selected (SAT or ALL works)

AndreasK79:
Small fix if you set band other than SAT or ALL.


phl0 (approved):
Fixes the issue on my system.

Issue #2422, opened 2025-10-19
Add rows for qslprint


Arielfoever:
**By submitting feature requests, you have the opportunity to contribute your ideas and inspiration to the Wavelog Dev-Team. Please be aware that not every feature request will be addressed or answered.**

**What Feature is Missing?**
Describe as precisely as possible what you believe is missing. Also, explain how we could achieve the inclusion of this feature.

Add rows in qsl print table showing

- QSL sent to this callsign (ignore band and mode)
- QSL rcvd from this callsign

**Why is This Feature Important?**
Not every feature holds equal importance. Clarify why you think your suggested feature is crucial and how it enhances Wavelog.

Sometimes people want to sent their cards to new receivers so there can be a statics for them.

**Is This Feature Personal or Beneficial to Others as Well?**
If the feature is solely for your benefit, and not beneficial to the broader Wavelog community, we may not prioritize or respond to your request. Ensure that your proposed feature contributes to the entire Wavelog user base.

AndreasK79:
Do you mean QSL rcvd on same band/mode/(sat), or ignore that?

Arielfoever:
> Do you mean QSL rcvd on same band/mode/(sat), or ignore that?

both or you can adjust it.

lycanananas:
Hello,

I issued already problem with QSL printing in Wavelog. After discussion with Wavelog team we agreed that external application is best option. 

Now you can easily vibecode one that fulfull our need. You can check my github for my application that generates QSL labels.

PR #3262, opened 2026-05-25
[QSL Queue] Add qsl information, fix delete bug

AndreasK79:
This pr does the following:

- If you cancel in the dialog for removing QSO from table, it would be removed regardless. Cancel now works
- Added pagination to table
- Changed so that dropdowns are on same line on page
- Fixed an issue with a full group by in the main query when entering page

Requested in #2422:
- Added information about previous QSL sent, regardless of band/mode
- Added information about previous QSL received, regardless of band/mode
Example (mouseover that says what is what):
image

### DIFF
--- a/application/controllers/Qslprint.php
+++ b/application/controllers/Qslprint.php
@@ -168,38 +168,38 @@ class QSLPrint extends CI_Controller {
 	}
 
 	public function delete_from_qsl_queue() {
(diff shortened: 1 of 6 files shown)


phl0 (approved):
Tested and works as expected. No queued requests got deleted on cancelling the popup.

int2001 (approved):

PR #3265, opened 2026-05-26
Fix prop and type

int2001:
- Propmode was missing in controller for statistics and distancerecords.
- Limit input on EDIT / MASSEDIT to the correct types. Otherwise User may - e.g. - Enter "100W" for TX-Power and it fails in backend.


phl0 (approved):
Tested. Works

AndreasK79 (approved):

PR #3271, opened 2026-05-28
[QSL Queue] Add option to show all rows + change labels in header

DB4SCW:
This PR changes 2 small things introduced by [[QSL Queue] Add qsl information, fix delete bug #3262](https://github.com/wavelog/wavelog/pull/3262) lately.

1) Add pagination option to show "all" rows.
This restores the functionality of the workflow introduced in #2306, as well as the opportunity to view and sort for larger QSL jobs then 100. Defining the pagination options this way also allows us to add additional different steps later on in addition to the default steps. I also added 25 rows as a default, which is overwritten by the stateSave feature once the user decides on his personal preference. In the current version, the view would start at a (very conservative imho) 10 entries.
2) Changes the label for the "Band or Frequency" select to not wrap

From:
grafik
To:
grafik

### DIFF
--- a/application/views/qslprint/index.php
+++ b/application/views/qslprint/index.php
@@ -1,3 +1,7 @@
+<script>
+	var lang_export_qslprint_pagination_all = '<?= __("All") ?>';
+</script>
+
 <div class="container">
 
 	<br>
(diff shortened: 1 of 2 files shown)


application/views/qslprint/index.php
```diff
@@ -1,3 +1,7 @@
+<script>
+	var lang_export_qslprint_pagination_all = '<?= __("All") ?>';
```
HB9HIL:
```suggestion
	var lang_export_qslprint_pagination_all = "<?= __("All") ?>";
```
we had issues in the past with using single quotes in translations. Use double quotes in JS :)

HB9HIL (changes requested):

application/views/qslprint/index.php
DB4SCW: good information. Will submit another PR for another place where single quotes are still in use.

AndreasK79 (approved):

HB9HIL (approved):

PR #3277, opened 2026-05-30
[EME initials] Unify page design

AndreasK79:
Before:
image

After:
image


int2001 (approved):

application/views/statistics/initials.php
```diff
@@ -1,39 +1,44 @@
-<div class="container">
+<div class="container oqrs px-3 px-lg-4 mt-3 mb-3">
```
int2001: minimal thing: i think cp issue: OQRS Class? in EME? isn't defined anywhere, but doesn't hurt. looks only a bit strange?

AndreasK79: Yep, you are right. Copy/paste... removed now.

PR #3279, opened 2026-05-31
[DXCluster] Have the page follow the same width as the top menu

AndreasK79:
Before:
image

After:
image


int2001 (approved):
looks more aligned now. even if it isn't using the full-width. but consistency first

PR #3284, opened 2026-06-04
[Sat skeds] Gave the page a better look

AndreasK79:
Before:
image

After:
image

### DIFF
--- a/application/views/satellite/skedtable.php
+++ b/application/views/satellite/skedtable.php
@@ -2,29 +2,33 @@
 if (!empty($overlaps)) {
 	echo '<div class="table-responsive"><table style="width:100%" class="table-sm table table-bordered table-hover table-striped table-condensed text-center">';
 	echo '<thead>
-	<tr>
-	<th>' . __("Satellite") . '</th>
-	<th>' . __("Date") . '</th>
-	<th>' . __("Sked AOS Time") . '</th>
-	<th>' . __("Sked LOS Time") . '</th>
-	<th>' . __("Duration") . '</th>
-	</tr>
-	</thead>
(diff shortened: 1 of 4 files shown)


HB9HIL (changes requested):
when entering a sked partner grid I'm missing these graphs

Image

without the sked partner grid I see the graphs

in the previous version the sked partner was hidden with a button "Add a sked partner". Could we consider it here aswell to make it more clean that this is optional

HB9HIL:
> when entering a sked partner grid I'm missing these graphs
> without the sked partner grid I see the graphs

ah I see. this is the same behaviour as before.. however hiding the sked partner and adding a plus sign would be appreciated 😁

application/views/satellite/skedtable.php
int2001, line 70: no custom_format here? at previous table you changed it

int2001 (changes requested):
in terms of fitting this page to the unified optics - well done. only minor one.
regarding that path when sched is enabled, i'd suggest another PR later...

AndreasK79:
> ah I see. this is the same behaviour as before.. however hiding the sked partner and adding a plus sign would be appreciated 😁

Yes, same behavior, depending if grid is filled or not. Do we really need to hide and add?

application/views/satellite/skedtable.php
AndreasK79: It is using $format, which is custom format + time, set in controller.

HB9HIL (approved):
fine from my side
let's wait for @int2001

int2001 (approved):
sorry. my requested change was wrong. everything works

PR #3288, opened 2026-06-04
new 404 override

HB9HIL:
I tried before in #2188 to redesign the 404 page.. I thought I need to do that at the 404 catch.. but well... Next time I should read documentation better 😆 

there is actually a override in the routes to set your own controller. No issues with gettext and theme. Works perfectly

just visit `https://log.your-wavelog.org/index.php/foobar` to test

Before:
Bildschirmfoto 2026-06-05 um 00 29 57

After:
grafik

### DIFF
--- a/application/config/routes.php
+++ b/application/config/routes.php
@@ -49,7 +49,7 @@
 */
 
 $route['default_controller'] = "dashboard";
-$route['404_override'] = '';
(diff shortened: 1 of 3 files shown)


int2001 (approved):
solid one

PR #3291, opened 2026-06-06
[Advanced Logbook] Adjusted some classes and adjusted checkbox in table

AndreasK79:
Removed sorting on first column, as it would interfere with the checkbox. Now the checkbox is centered.
The other classes are just extra so that one can adjust things in custom themes.

Before:
image

After:
image


HB9HIL (approved):
lgtm, tested working

PR #3299, opened 2026-06-08
Sessiondata

HB9HIL:
refactored the `update_session()` function in the user_model. This reduces the initial query count at login from around 60 down to 10. User options queries at login: 32, now 1. Memory improvement by storing the stuff in structured arrays

- better readability
- easier maintenance at the future
- login much faster and much less load on database

also maintained the user_options_model to match our today code style, removed unused code and added the main query to get all user options at once in a structured array

### DIFF
--- a/application/models/User_model.php
+++ b/application/models/User_model.php
@@ -632,7 +632,7 @@ class User_Model extends CI_Model {
 			'active_station_logbook' 		=> $u->active_station_logbook,
 			'user_language' 				=> isset($u->user_language) ? $u->user_language: 'english',
 			'isWinkeyEnabled' 				=> $u->winkey,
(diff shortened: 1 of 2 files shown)


HB9HIL:
PS: Logic still exactly the same

application/models/User_model.php
int2001, line 1134: old "bug", but noticed while reviewing this PR. 
in L635 you read/write `shown`, while `showed` is used here (and a few more times later)

int2001 (changes requested):
in general it works. but it's worth to consider cleaning up the `shown` vs `showed` missmatch as well

application/models/User_model.php
HB9HIL: Good catch

int2001 (approved):

PR #3303, opened 2026-06-10
Qsldesigner refactor

AndreasK79:
So this is a draft of discussion #3051. This is based on the works of @twicklund76, and the branch @HB9HIL did.

Hard requirements:
- PR #3315 

I have done the following:
* Added migration script
* Split out js
* Split out most of the css (can probably split out rest)
* Replaced hardcoded table name
* Added translatable strings to the designer view

What I have noticed that needs to be fixed:
- [x]  The qsl_postcard_templates table needs a user_id, following that, the queries needs to check for templates belonging to the userid
- [x] We need to be able to delete templates
- [x] no print if no address, but if you just want to print on the card without address, this should be allowed
- [x] Switched to toasts in the designer for feedback
- [x] Sec: The functions in model are missing the mandatory station_id filter.
- [x] The code for fetching addresses from callbook must be refactored to use existing libraries. 
- [x] Consider caching (short TTL) for the callbook-fetches.
- [x] Filter from QSL Queue (in connection to the button "Print Postcards for all QSOs") doesn't do anything, removed!
- [x] Give message about no templates found
- [x] Designer: CSS for the ruler adjusted
- [x] Designer: CSSS right click menu needs to be adjusted
- [x] Sec: Path-Traversal for Background-Image seems to be possible. BG-Upload needs to be bulletproof
- [x] at the very end the `log_message("Error",...);` needs to be removed or switched to debug-level
- [x] Add image background delete on template delete
- [x] Fix multiple QSOs on one card, it's not built for this, so logic needs to be added
- [x] Rewrote the form and backend to use template options from the "Print QSL Postcards" forms
- [x] Added colors to objects

Designer before:
image
Designer after:
image


AndreasK79:
@twicklund76 as you can see, I've done most of the adaption needed, but some changes are still needed.

I see you have hidden font, font size and bold in the view. Any reason for that? Future plans?
Also I noticed the filter from the button in QSL Queue doesn't really do anything. Future plans here too?

AndreasK79:
@twicklund76 I've figured out the font thingy. It is element properties of course. I am just figuring out how this works :)

HB9HIL (approved):
Awesome! I'd say we merge it to further test it in dev branch!

PR #3304, opened 2026-06-10
Docker Image Refactoring

HB9HIL:
> [!note]
> Relies on https://github.com/mlocati/docker-php-extension-installer/pull/1305 why this PR is still drafted yet

One big thing which was requested several times, was the ability to set custom UID/GID in the container and maybe other environment variables. This was not possible due the way our Dockerfile was designed. Everything was hardcoded. 

I refactored the Dockerfile and the whole docker logic. Best news first: For existing docker-compose nothing changes. This is fully backwarts compatible. 

1. I fixed some structural bug in the dockerfile which lead to commands do not 
2. User can optionally overwrite some values in the container (see documentation below)
3. the new entrypoint can be extended as needed

---
### `PUID`
UID the web server runs as. Files newly created by Wavelog are owned by this UID.

- **Default:** unset — the container runs as `www-data` (UID `33`) and behaves
  exactly as before.
- **When to set:** on bind-mount setups (e.g. TrueNAS, Unraid), set it together
  with `PGID` to your host user's UID/GID so the data aligns with your host.

### `PGID`
GID the web server runs as and that owns Wavelog's writable directories
(`uploads`, `userdata`, `backup`, `application/cache`, `application/config`, …).
The directories are group-writable, so this group has write access to the data.

- **Default:** `33` (the `www-data` group).
- **When to set:** alongside `PUID` to match your host user's group.

### PHP settings

These map directly to the corresponding `php.ini` directives.

| Variable                  | Default | Sets                  |
| ------------------------- | ------- | --------------------- |
| `PHP_MEMORY_LIMIT`        | `256M`  | `memory_limit`        |
| `PHP_UPLOAD_MAX_FILESIZE` | `64M`   | `upload_max_filesize` |
| `PHP_POST_MAX_SIZE`       | `64M`   | `post_max_size`       |
| `PHP_MAX_EXECUTION_TIME`  | `600`   | `max_execution_time`  |

### Example
```yaml
    services:
      wavelog:
        image: ghcr.io/wavelog/wavelog
        environment:
          PUID: 1000
          PGID: 1000
          PHP_UPLOAD_MAX_FILESIZE: 128M
        volumes:
          - ./data:/var/www/html/userdata
```
---


HB9HIL:
Successfull Docker Image Run: https://github.com/HB9HIL/wavelog/actions/runs/27329749852

Dockerfile
```diff
@@ -1,52 +1,48 @@
...
-post_max_size = 64M\n\
-max_execution_time = 600\n" > $PHP_INI_DIR/conf.d/wavelog.ini
-
-# Copy proper file to target image
-COPY ./ /var/www/html/
+        gd
+
+# server config
+RUN a2enmod rewrite \
+    && echo "* * * * * root curl --silent http://localhost/index.php/cron/run >/dev/null 2>&1" > /etc/cron.d/wavelog \
```
int2001, line 26: just idea: we can also run this one as www-data instead of root. what do you think?

int2001 (approved):

Dockerfile
HB9HIL: No need that the webserver owns the models and other code. Would weaken the security on normal installations. www-data just need to read. This applies only for non PUID/PGID overwrites

PR #3306, opened 2026-06-11
CORS for API

int2001:
so far only "radio"-API had proper CORS-Header, that other web-only-apps can use CAT.
All other APIs were only reachable from serverside or fatclient 3rd-party apps.

This one opens our APIs for web-applications (how it should be, when providing API).


HB9HIL (approved):
lgtm

PR #3311, opened 2026-06-11
Fix the summary of WAPC award

cebarobot:
This patch fixes some bugs in WAPC award.

1. The summary count is incorrect. The previous implementation count the distinct pair of DXCC and STATE, which causes double counting for DXCC other than 318.

This patch uses `case` to normalize the province in summary count. This will solve the issue.

2. In `get_wapc_array()`, a small issue for count.


HB9HIL:
I missclicked hahaha and accidentially requested a review from Copilot

AndreasK79 (approved):

AndreasK79:
@cebarobot thanks for the fixes. This looks good to me.

PR #3315, opened 2026-06-11
Paths Refactoring

HB9HIL:
This is a larger refactoring of the Paths Library, namely the userdata path logic

when it was implemented by @abarrau a while ago, improved by @int2001 and myself we didn't take into account that in future there might by other userdata uploaded aswell. This will propably now the case with #3303 as we want to store this data in the userdata directory, seperated by the users. 

To make this possible we need to refactor the logic and (which is a good idea anyway) reduce duplicate code. The get_Imagepath logic was duped in qsl and eqsl logic only seperated by a keyword, aka. data- `$type` and is with this PR now merged into one userdata paths logic. This new logic allows future expansion with other datatypes. 

As small bonus for the future: There is a new security helper function `valid_uid()` which securely validates a user_id and makes sure it's an integer larger 0. This prevents dealing with `?? '') != '' || foobar` and makes it much easier to securely validate a user_id (which is part of the filepath by the way)

> [!note]
> The paths lib is autoloaded so you don't need `$this->load->library('paths');`

---

### a small documentation

you want the web path for eqsl image $foobar, use
```php
<?php
// second parameter defaults to 'u' for url
$filepath = $this->paths->$this->paths->getUserdataPath('eqsl_card') . '/' . $foobar;
```
you want the serverside filepath for eqsl image $foobar, use
```php
<?php
// 'p' is path like serverside filepath
$filepath = $this->paths->$this->paths->getUserdataPath('eqsl_card', 'p') . '/' . $foobar;
```
you want the web path for qsl card image $foobar, use
```php
<?php
$filepath = $this->paths->$this->paths->getUserdataPath('qsl_card') . '/' . $foobar;
```
you need a path to store the receipe for a good german currywurst on the server, use
```php
<?php
$filepath = $this->paths->$this->paths->getUserdataPath('currywurst', 'p') . '/' . $foobar;
```
...
you get an idea

### DIFF
--- a/application/controllers/Eqsl.php
+++ b/application/controllers/Eqsl.php
@@ -26,7 +26,7 @@ class eqsl extends CI_Controller {
 
 		$this->load->model('eqsl_images');
 		$this->load->library('Genfunctions');
-		$folder_name = $this->eqsl_images->get_imagePath('p');
+		$folder_name = $this->paths->getUserdataPath('eqsl_card', 'p');
 		$data['storage_used'] = $this->genfunctions->sizeFormat($this->genfunctions->folderSize($folder_name));
 
 		// Pagination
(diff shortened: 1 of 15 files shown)


int2001 (approved):
as seen from code-perspective: Well done.
tbh: cannot really test everything, because i never used storage of QSL-Cards in my instance.

besides this well-done PR: Idea for next refactor: 
- make QSL-Cards-storage safe. e.g.: Deny huge files (perhaps auto-scale-down for giant files), check for real images, etc. etc.

PR #3318, opened 2026-06-13
Introduce UploadGuard for Qsl-Images and use it

int2001:
This one tries to harden User-Uploads.

Features:
- [x] Check if uploaded QSL-Image is really an image
- [x] Reject upload if diskspace doesn't allow it (at least 1 GB should stay free / tight, but there are lot of instances running on Cheap hosters)
- [x] Show possible Failures to user (as toast)
- [x] Use Uploadguard as well in other logics


HB9HIL (approved):
i like :)

PR #3319, opened 2026-06-13
Add ITUZ to private_lookup API (if reliable)

int2001:
`/api/private_lookup` lacks of ITU-Zone.

This one adds the ITUZ in exactly the same way as CQZ is derived


application/controllers/Api.php
```diff
@@ -1347,6 +1361,22 @@ function private_lookup() {
...
+		}
+		$sql = "SELECT GROUP_CONCAT(DISTINCT ituzone) AS zones, COUNT(DISTINCT ituzone) AS cnt
+			FROM dxcc_master WHERE countrycode = ? AND ituzone <> ''";
+		$row = $this->db->query($sql, [(int) $dxcc_id])->row();
+		if ($row !== null && (int) $row->cnt === 1 && (int) $row->zones > 0) {
+			return (int) $row->zones;
+		}
+		return null;
+	}
+
```
HB9HIL: it would be more concise to move SQL to a model

HB9HIL (changes requested):

application/controllers/Api.php
int2001: working on it - there's a more easy way

AndreasK79 (approved):

PR #3322, opened 2026-06-14
delete the whole userdata directory for the user if the user is deleted

HB9HIL:


int2001 (approved):
works, but consider logging if rmdir failed:

https://github.com/HB9HIL/wavelog/pull/148

PR #3345, opened 2026-06-18
Activators grid color

phl0:
For some reason we used a separate hard coded green color for grid actvators map. This changes the code to use the user configured color for confirmed QSOs to make these maps a little more flexible. Screenshot shows before and after:

image

### DIFF
--- a/application/controllers/Activators.php
+++ b/application/controllers/Activators.php
@@ -18,8 +18,12 @@ class Activators extends CI_Controller
     public function index()
     {
         $data['page_title'] = __("Gridsquare Activators");
+        $validHex = function($color, $default) {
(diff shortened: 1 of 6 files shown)


application/controllers/Activators.php
```diff
@@ -18,6 +18,9 @@ function __construct()
     public function index()
     {
         $data['page_title'] = __("Gridsquare Activators");
+        $map_custom_colors = json_decode($this->optionslib->get_map_custom());
```
int2001, line 24: fallback to default color(s)? like here? 

https://github.com/wavelog/wavelog/blob/1d358c23a855c2800cfc6f146b740bc58ad66677/application/controllers/Bandmap.php#L61

phl0: Good idea. Fixed.

int2001 (approved):
good one for consistency

PR #3349, opened 2026-06-18
Switch Operator

HB9HIL:
This requested from various operators during testing of the new contesting logger. This PR introduces a "switch operator" feature for the new contest logger.

How it works:
If a contesting session is open in a clubstation you can see a little badge with the operator callsign
grafik

If clicked it opens the control panel and shows the "Switch Operator" button
grafik

This opens a modal which allows either login as new user and directly jump into the same contesting session with the new user

OR if `club_direct` is enabled and it was a direct login simply rewrite the php session data to the new operator. 
grafik

---

this allows contest teams to fastswitch operator during a contesting session without going aall the way back (close session, logout, login, contestmanager, start session) which is really cumbersome.

### DIFF
--- a/assets/js/sections/operator.js
+++ b/assets/js/sections/operator.js
@@ -51,7 +51,7 @@ function saveOperator() {
 				window.location.reload();
 			},
 			error: function () {
-				showToast("error", __("Error saving operator callsign. Please try again."));
+				showToast("error", lang_operator_modal_save_error, 'bg-danger text-white', 5000);
 			}
 		});
 	} else {
--- a/assets/js/sections/operator.js
+++ b/assets/js/sections/operator.js
@@ -43,10 +43,16 @@ function saveOperator() {
 			data: {
 				operator_callsign: operatorCallsign,
 			},
+			// Reload only after the session write succeeded — reloading earlier races
+			// the POST and can render the previous operator.
+			success: function () {
+				closeOperatorDialog();
+				window.location.reload();
+			},
+			error: function () {
+				showToast("error", __("Error saving operator callsign. Please try again."));
+			}
 		});
-		closeOperatorDialog();
-		// console.log("saveOperator executed");
-		// console.log("operator:" + operatorCallsign);
 	} else {
 		operatorInput.addClass("is-invalid");
 	}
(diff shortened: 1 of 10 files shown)


assets/js/sections/operator.js
```diff
@@ -43,10 +44,16 @@ function saveOperator() {
...
 				operator_callsign: operatorCallsign,
 			},
+			// Reload only after the session write succeeded — reloading earlier races
+			// the POST and can render the previous operator.
+			success: function () {
+				closeOperatorDialog();
+				window.location.reload();
+			},
+			error: function () {
+				showToast("error", __("Error saving operator callsign. Please try again."));
```
int2001: sure about this? gettext in JS? is it new? asking for a friend

HB9HIL: upps you are right, this is bullshit

int2001:
Minor issue:

if QSOs are pending to sync, and the user is doing a OP-Change, there might be a racecond

HB9HIL:
> Minor issue:
> 
> if QSOs are pending to sync, and the user is doing a OP-Change, there might be a racecond

grafik

Already catched

int2001:
works.
a couple more annotations (you may fix them in a different PR, or i got it wrong):
- Where can OP see which OP is active in contest-engine?
- When switching over to a different op (call it OP-B), and calling dashboard / something else (different tab) everything happens as OP-B under Clubcall, which is intended. But: when user wants to go back to his own OP-B, it fails. imho no big problem (he could re-login), but a bug

HB9HIL:
> works. a couple more annotations (you may fix them in a different PR, or i got it wrong):
> 
>     * Where can OP see which OP is active in contest-engine?

Is only shown if it's a clubstation
grafik

> 
>     * When switching over to a different op (call it OP-B), and calling dashboard / something else (different tab) everything happens as OP-B under Clubcall, which is intended. But: when user wants to go back to his own OP-B, it fails. imho no big problem (he could re-login), but a bug

True. All the other tabs are then logged in as the other user. This is intended and there is no real solution for this.

int2001 (approved):
perhaps add a information to the "Switch Operator"-Modal who is the current OP.

i know, it's already there as a small badge, but when users are changing, or it's busy during contesting, it'd avoid searching for it.

PR #3354, opened 2026-06-19
Switches

int2001:
Replace boolean dropdowns in user-settings with switches (less space, more unified UI).

Before Patch:
image

After patch:
image

### DIFF
--- a/application/views/user/edit.php
+++ b/application/views/user/edit.php
@@ -410,76 +410,90 @@
 							<div class="card">
 								<div class="card-header"><?= __("QSO Logging Options"); ?></div>
 								<div class="card-body">
-									<div class="mb-3">
-										<label for="logendtimes"><?= __("Log End Times for QSOs Separately"); ?></label>
-										<?php if(!isset($user_qso_end_times)) { $user_qso_end_times='0'; }?>
+									<?php if(!isset($user_qso_end_times)) { $user_qso_end_times='0'; }?>
+									<div class="d-flex align-items-start gap-2 mb-3">
 										<input type="hidden" name="user_qso_end_times" value="0">
-										<div class="form-check form-switch">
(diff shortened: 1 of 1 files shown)


application/views/user/edit.php
```diff
@@ -410,76 +410,90 @@
...
+									<div class="d-flex align-items-start gap-2 mb-3">
 										<input type="hidden" name="user_qso_db_search_priority" value="N">
-										<div class="form-check form-switch">
+										<div class="form-check form-switch mt-1">
 											<input class="form-check-input" type="checkbox" role="switch" id="db_search_priority" name="user_qso_db_search_priority" value="Y" <?php if ($user_qso_db_search_priority == 'Y') { echo 'checked'; } ?>>
 										</div>
-										<small id="SelectDateFormatHelp" class="form-text text-muted"><?= __("When set to \"Yes\", callsign lookup will first use data from your previous QSOs before querying external services. Set to \"No\" to always use external lookup services instead."); ?></small>
+										<div>
+											<label class="d-block mb-0" for="db_search_priority"><?= __("Prioritize database search over external lookup"); ?></label>
+											<small id="SelectDateFormatHelp" class="form-text text-muted"><?= __("When set to \"Yes\", callsign lookup will first use data from your previous QSOs before querying external services. Set to \"No\" to always use external lookup services instead."); ?></small>
```
HB9HIL: we have to keep an eye on the double quotes how weblate will handle that. didn't test that for over an year and not with the newer versions of weblate

HB9HIL:
I like so far :)

application/views/user/edit.php
int2001: Fixed it

HB9HIL (approved):

PR #3355, opened 2026-06-19
Some enhancements

imlonghao:
1. make sure API endpoint with json responses have the correct `application/json` content-type
2. make sure `$stationid` is int to prevent SQL injection
3. prevent user run other's stored queries


int2001 (approved):
solid one

HB9HIL (approved):
lgtm

PR #3357, opened 2026-06-20
Support multiple station_ids at once for get_contacts_adif

int2001:
`get_contacts_adif` api has to be called for every single station_id.
if there are multiple station_ids (at one key/user) to be fetched, this has to be done over and over.

this patch fixes it, so station_id now takes an array as well as a single station_id (like before the patch).

Test:

new (station_id 1 and 2 at once)

`curl -X POST -d '{"key":"[WLKEY]", "station_id":[1,2], "fetchfromid":1578569, "output_format":"adif"}'  https://[WL_URL]/api/get_contacts_adif`

single (as before patch - only station_id 1):

`curl -X POST -d '{"key":"[WLKEY]", "station_id":1, "fetchfromid":1578569, "output_format":"adif"}'  https://[WL_URL]/api/get_contacts_adif`

### DIFF
--- a/application/controllers/Api.php
+++ b/application/controllers/Api.php
@@ -532,8 +532,24 @@ class API extends CI_Controller {
 
 		//extract relevant data to variables
 		$key = $obj['key'];
-		$station_id = $obj['station_id'];
 		$fetchfromid = $obj['fetchfromid'];
+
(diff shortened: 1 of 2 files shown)


HB9HIL (approved):
lgtm. cr done

PR #3358, opened 2026-06-20
SAT: Add OMM Support to be future-proof

int2001:
As everyone knows the days of classic TLE's (Two Line Elements // keps) are counted. For the interested user: See 3.1. in this [document](https://space.commerce.gov/wp-content/uploads/2026/01/TraCSS-Spec-004-v1.2_OMM.pdf)

However - the proposed the new Standard is OMM, which solves several problems.

This PR makes WL future-proof in the following way:
1. Legacy TLE - still there and primary source
2. Fill gaps with OMM from https://celestrak.org/NORAD/elements/gp.php?GROUP=amateur&FORMAT=JSON
3. Show if TLEs or OMM is taken at admin -> satellites
4. Process TLEs and OMMs in the same way at: sat-passes, sat-flightmap, qso

Testing:
1. Go to admin -> SATs -> fetch TLE's
2. Check for SATs which are only OMM, and check sat-passed, sat-flightmap, qso for them
3. Do the same (backwards-compat) with TLE sats
4. more tests: manually add TLE. (Adding OMM manually is currently not in scope)

### DIFF
--- a/application/controllers/Satellite.php
+++ b/application/controllers/Satellite.php
@@ -223,11 +223,10 @@ class Satellite extends CI_Controller {
 
 		// Synthesize TLE text at render time for the legacy JS propagator (OMM-stored sats).
 		if ($satellite_data && !empty($satellite_data->tle)) {
-			$raw = trim($satellite_data->tle);
-			if ($raw !== '' && ($raw[0] === '{' || $raw[0] === '[')) {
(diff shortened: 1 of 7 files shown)


phl0 (approved):
Works fine so gar. No glitches detected.

PR #3360, opened 2026-06-20
Mobile marker in QSL Postcard designer

HB9HIL:
the current `qso.portable` marker triggers on /P and /M

on my QSL cards I have two boxes, one for /P and one for /M


int2001 (approved):
absolutely

PR #3362, opened 2026-06-20
option for active logbook stations

kc9uhi:
This PR adds a new user option 'stations_active_log_only'. Restricts listing of station locations to only those linked to the active logbook.

QSO entry, station location quickswitch, ADIF import/export, and API endpoint station_info follow this option selection.


application/controllers/Logbookadvanced.php
```diff
@@ -60,7 +60,11 @@ function index() {
 		$pageData['iotaarray'] = $this->logbook_model->fetchIota();
 		$pageData['sats'] = $this->logbookadvanced_model->get_worked_sats();
 		$pageData['orbits'] = $this->bands->get_worked_orbits();
-		$pageData['station_profile'] = $this->stations->all_of_user();
+		if (!empty($this->session->userdata('user_stations_active_log_only'))) {
+			$data['station_profile'] = $this->logbooks_model->list_logbooks_linked($this->session->userdata('active_station_logbook'));
+		} else {
+			$data['station_profile'] = $this->stations->all_of_user();
```
int2001: why `$data` instead of `$pageData`?
whole view is loaded with `$pageData`

int2001: i really appreciate your contribution(s).
But: can we please commit to keep them as small as possible?

since we review by hand/manual it's really effort to do it

kc9uhi: human copy-paste failure across different controllers with same logic block.

some use $data for everything (like Qso), some split $data and $pageData.

kc9uhi: fixed with e62d7b3

application/controllers/Api.php
```diff
@@ -261,33 +261,44 @@ function create_station($key = '') {
 	function station_info($key = '') {
 		$this->load->model('api_model');
 		$this->load->model('stations');
+		$this->load->model('user_options_model');
```
HB9HIL: model is autoloaded in application/config/autoload.php
no manual load necessary
```suggestion
```

application/controllers/Api.php
```diff
@@ -261,33 +261,44 @@ function create_station($key = '') {
...
-				$result['station_cq']=$row->station_cq;
-				$result['station_itu']=$row->station_itu;
-				$result['station_state']=$row->state;
-				$result['station_country']=$row->station_country;
-				array_push($station_ids, $result);
+
+			$dkey_opt=$this->user_options_model->get_options('stations',array('option_name'=>'active_log_only','option_key'=>'boolean'), $userid)->result();
+			$user_stations_active_log_only = (count($dkey_opt)>0) ? $dkey_opt[0]->option_value : false;
+			if($user_stations_active_log_only) {
+				$this->load->model('logbooks_model');
```
HB9HIL: model is autoloaded in application/config/autoload.php
no manual load necessary
```suggestion
```

HB9HIL (changes requested):

application/controllers/Api.php
kc9uhi: these are loaded manually in many other locations, perhaps a future clean-up task

HB9HIL: yep

HB9HIL:
We need to discuss this for Logbook Advanced. In the past $users where used to it that LBA shows every station location. I'm fine with it since this option is default disabled (which is important in this case imho). But I want to keep an eye on that to avoid unexpected behaviours and support tickets  - @AndreasK79

HB9HIL:
PS: To be consistent you should add this behaviour also to the new contesting

application/views/adif/import.php
```diff
@@ -270,14 +250,15 @@
 									<div class="small form-text text-muted"><?= __("Select Station Location") ?></div>
 									<select name="station_profile" class="form-select mb-2 me-sm-2">
 										<option value="0"><?= __("All") ?></option>
```
HB9HIL, line 254: "All" means export from all station locations. A user could be confused if the option is enabled but only the currently active ones are exported. May you should add a little tooltip or something to the help text, so it's clear for the user that "All" means in this case only the locations linked with the current active logbook. Just an idea

HB9HIL:
> PS: To be consistent you should add this behaviour also to the new contesting

and SimpleFLE. Other places can be checked in another PR as they are more likely "sync"/"export" related and not really in the main user area. Would overload this PR. But SimpleFLE would be in scope here I guess aswell

kc9uhi:
I started in on expanding to all other locations as applicable. Main focus is user QSO entry and adif import/export. Will hold off on those for another expansion PR.

Will finish up SimpleFLE, QSO edit, and comments in adif export.

application/views/adif/import.php
kc9uhi: added info line, only appears if stations_active_log_only option is active
07a9d2f

HB9HIL:
good you found the bug in the session modal 😆 you can keep it 😁

phl0 (approved):
Tested successfully:
- QSO logging
- Logbook advanced
- API endpoint station_info
- ADIF export and import

PR #3368, opened 2026-06-21
[Advanced Logbook] Adjust query so you can dupe check across locations

AndreasK79:
The duplicate check was adjusted a little while ago, using a windowing function.
Problem was that it partitioned with station_id. That means that it only found duplicate QSOs within the station location, no matter if you selected several locations.

Thanks to @HB9HIL for finding this one.


HB9HIL (approved):
approved by finding some more dupes in my log 😆

PR #3369, opened 2026-06-22
Add display of pic to qrz.ru lib

phl0:
Implements https://github.com/wavelog/wavelog/discussions/3365 and adds callbook image to qrz.ru lib.


AndreasK79 (approved):

AndreasK79:
Can't test, but code looks good. Also seems the same as the sample that was provided.

PR #3370, opened 2026-06-23
Enable clubmembers to massedit/delete their QSOs

int2001:
Problem: A clubmember can edit/delete the QSOs he made. But only QSO by QSO, not in Logbook-advanced.
This one fixes it.

Changes:
- Backend of LBA checks if QSO was created by user (helper-function in clubhelper)
- Operator-Massedit is prohibited for Clubmember
- instead of returning the amount of edited/removed QSOs, the controller returns its ids. so JS/Frontend can only the QSOs where edit/rm suceeded. If user wants to edit/delete a QSO he didn't create:
image
image

Testing:
- Test (Mass-) Edit/Delete for admin, normal user, clubadmin in LBA
- Test (Mass-) Edit/Delete for clubmember. Create or use a NORMAL User which is granted as clubmember to a clubstation. Switch over. Don't test it via Impersonate. Relogin as clubmember

### DIFF
--- a/application/controllers/Logbookadvanced.php
+++ b/application/controllers/Logbookadvanced.php
@@ -641,7 +641,7 @@ class Logbookadvanced extends CI_Controller {
 	}
 
 	public function editDialog() {
-		if(!clubaccess_check(9)) return;
+		if(!clubaccess_check(3)) return;
 
 		$this->load->model('bands');
 		$this->load->model('modes');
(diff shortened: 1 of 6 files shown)


phl0 (approved):
Tested and works on my end. Log:

- Created a clubstation
- granted my own user club officer
- created two separate users with club member permissons
- logged a few QSO with each of these callsigns

Now:

- Club officer can edit/delete everything
- club member 1 can only edit/delete his own QSOs
- club member 2 can only edit/delete his own QSOs

So works as designed imho. Approved.

PR #3373, opened 2026-06-24
Additional implementation of 'stations_active_log_only' option

kc9uhi:
This is an extension of #3362

Adds support for the stations_active_log_only user option to QSO search functions, and station location listing (for clubmember/clubmember adif). Help text indicating limited listing is present when user option is enabled.

The clubmember/clubmember adif already have limited privileges in the stationsetup pages; this extends that behavior to hide station locations not associated with the active logbook (if option is enabled). (also found station location export button permission bug, that'll be a different PR)

Pages affected:
search
search/filter
search/lotw_unconfirmed
stationsetup/list_locations

### DIFF
--- a/application/models/Stationsetup_model.php
+++ b/application/models/Stationsetup_model.php
@@ -250,7 +250,17 @@ class Stationsetup_model extends CI_Model {
 		$result = $query->result();
 		$this->load->model('user_options_model');
 
-		foreach($result as $location) {
+		// restrict station locations listing to those linked to active logbook for clubmember/clubmember adif, if user option enabled
+		$stations_linked = '';
+		if (in_array(($this->session->userdata('cd_p_level') ?? 0), [3,6]) && !empty($this->session->userdata('user_stations_active_log_only'))) {
+			$stations_linked = $this->logbooks_model->list_logbook_relationships($this->session->userdata('active_station_logbook'));
+		}
+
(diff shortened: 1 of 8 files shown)


application/models/Stationsetup_model.php
```diff
@@ -250,7 +250,17 @@ function list_all_locations() {
 		$result = $query->result();
 		$this->load->model('user_options_model');
 
-		foreach($result as $location) {
+		// restrict station locations listing to those linked to active logbook for clubmember/clubmember adif, if user option enabled
+		$stations_linked = '';
+		if (in_array(($this->session->userdata('cd_p_level') ?? 0), [3,6]) && !empty($this->session->userdata('user_stations_active_log_only'))) {
```
HB9HIL: this would probably be better to check by a > or < comparison since maybe we add more roles in future

HB9HIL (changes requested):

application/models/Stationsetup_model.php
kc9uhi: The view that calls this function also explicitly checks for user level of 3 or 6, rather than > or <, that's why I followed the same pattern. 

Adjusted to use < comparison.

phl0 (approved):
Works fine for me.

HB9HIL (approved):

PR #3378, opened 2026-06-25
allow clubmember / clubmember adif to export station locations

kc9uhi:
Station location list has existing 'CSV' and 'Export All Locations' buttons displayed to all users. 'Export All Locations' is not allowed to be accessed by clubmember and clubmember adif users. This PR enables the export_locations method to all users.

Information in exported json is the same as displayed and current CSV export button.

Page affected
stationsetup/list_locations

### DIFF
--- a/application/controllers/Stationsetup.php
+++ b/application/controllers/Stationsetup.php
@@ -12,7 +12,7 @@ class Stationsetup extends CI_Controller {
 		$this->load->helper(array('form', 'url'));
 
 		$this->load->model('user_model');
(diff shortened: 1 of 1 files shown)


int2001:
Hmm. There's already an api for it.
What's the UseCase for exporting every stationlocation to a "normal" clubmember/ADIF?

kc9uhi:
The button was already there, but the stationsetup controller disallowed access. CSV export button is also present and functions.

Other option is to remove both buttons for 'normal' clubmember. Station export button to json could be useful for clubmember adif users as they also have ADIF export abilities and can use both to import to their own wavelog account.

PR #3385, opened 2026-06-27
[Design QSL] fix upload background image

abarrau:
Hi team, 
Upload background image return 404 error.
I propose this fix : correct URI (with index.php).


HB9HIL (approved):
welcome back @abarrau. It's been a while since your last contribution 😄 

this one looks absolutely reasonable

abarrau:
Thanks, not have time, but i use always wavelog for me and some OM in my club ;) ... 
see u

PR #3399, opened 2026-06-29
Feature: Add Frequency Templates for quick QSO logging (UI + client-side JS)

CrazyBasti:
### Summary
- Adds a small, optional frequency‑template system to the QSO logging UI to speed up live logging for operators who frequently use repeaters, commonly used simplex frequencies, or specific digital channels.
- Provides a lightweight client-side implementation (UI + JS) so reviewers can test the workflow without a required server API or DB migration.

### Motivation
- Many users repeatedly log QSOs on the same repeater/simplex frequencies. Manually entering TX/RX, band and mode each time is slow and error prone.
- A one‑click template apply improves workflow speed and reduces entry errors — particularly valuable in portable/digital operation, contests, or when switching between repeaters.

### What this PR contains
- UI additions in the QSO view:
  - Template dropdown, “Apply” button and a modal to create templates (inserted next to the Frequency fields).
  - UI strings use the project’s i18n helper where appropriate.
- New client asset:
  - `assets/js/qso_templates.js` — handles loading, saving, seeding defaults, and applying templates to existing form fields.
- Optional demo/test view:
  - `application/views/dev/qso_templates_demo.php` — sandbox to preview and exercise the feature (if included in this branch).

### Implementation details
- The JS file currently uses `localStorage` for persistence (quick, no-backend test mode). It:
  - Seeds example templates when empty,
  - Fills the template dropdown,
  - Applies template fields to the QSO form (TX -> `freq_calculated`/`frequency`, RX -> `frequency_rx`/`freq_rx`, and attempts to set `mode`, `band`, `rst_*`, `comment`).
- The view only includes a small HTML block and a single `<script src="<?= $this->paths->cache_buster('/assets/js/qso_templates.js'); ?>"></script>` line. All logic is in the JS asset (separation of concerns / cacheable asset).

### Why this structure
- Keeps the view minimal and readable.
- Enables browser caching and consistent cache-busting through the existing asset mechanism.
- Makes it simple to swap `localStorage` for server persistence later by replacing the load/save functions in one file.

### Suggested server-side follow-ups (not part of this PR)
- Add a `qso_frequency_templates` table and a CI controller/model (CRUD) to persist templates per station/user.
- Replace `loadTemplates()` / `saveTemplates()` in `qso_templates.js` with AJAX calls to the new API.
- Add RBAC checks so only authorized users can create/edit station‑scoped templates.

### Testing instructions for reviewers
[...]

PR #3402, opened 2026-06-30
Fav names

int2001:
Issue:
- Favs only allowed one Fav per Band / Sat and Mode.
- Favs were automatically named.

since @CrazyBasti was looking for an alternative in #3399, this one fixes both issues. One can now freely name the FAV and save multiple ones per Band/Mode

Please(!) be aware of one very important thing to keep the hamspirit alive:
- When doing QSOs via Brandmeister-DMR, set Propagation "Internet assisted" to the QSO / Fav
- When doing QSOs via Repeater, set Propagation "Repeater" to the QSO / Fav

telegram-cloud-photo-size-2-5339261604907915852-y

image

Testinghints:
- Chose different QRGs (TX/RX), different Propmodes, SATs, etc. and save them as FAV
- Recall them

### DIFF
--- a/application/controllers/User_options.php
+++ b/application/controllers/User_options.php
@@ -21,6 +21,7 @@ class User_Options extends CI_Controller {
 		} else {
 			$option_name=$obj['band'].'/'.$obj['mode'];
 		}
+		$option_name = mb_substr($option_name, 0, 45);
 		$this->user_options_model->set_option('Favourite',$option_name, $obj);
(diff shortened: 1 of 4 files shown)


CrazyBasti:
@int2001 Thanks for That Update. I try that on my own Server 👍🏼👍🏼

HB9HIL (approved):
i like :)

PR #3405, opened 2026-07-01
One function for printing / fix filename

int2001:
QSLPostCard:
- [x] Move generator / filehandling-functions to one function/
- [x] Limit length for name (to prevent 500s)
- [x] Enable User to explicitly download the PDF on Top to existing inline-view


AndreasK79 (approved):

AndreasK79:
Works for me.

PR #3409, opened 2026-07-01
fix page size limit on mysql 8+

### LINKED ISSUE #3408: Installation fails on MySQL 8.0.46 - ERROR 1118 "Row size too large"
### Discussed in https://github.com/wavelog/wavelog/discussions/3404

<sup>Originally posted by **iu7wce** June 30, 2026</sup>
Hello everyone,
I would like to report an installation issue that seems to affect recent MySQL 8 installations.

**Environment**

- Wavelog 2.5.1
- Wavelog 3.0.0 (same behaviour)
- PHP 8.0
- MySQL 8.0.46
- Shared hosting (Tophost / TopWeb Plus)
- InnoDB
- utf8mb4

**Problem**
During a fresh installation, the installer stops while creating the database.
_The error is:_
```
ERROR 1118 (42000):
Row size too large (> 8126).
Changing some columns to TEXT or BLOB may help.
In current row format, BLOB prefix of 0 bytes is stored inline.
```
The failing table is:
``TABLE_HRD_CONTACTS_V01`
**`Important observations**
This is not an upgrade.
This is a completely fresh installation.
Exactly the same SQL schema worked perfectly on my previous hosting provider (MySQL 5.x).
After migrating to MySQL 8.0.46, both Wavelog 2.5.1 and 3.0.0 fail during installation with the same error.
The hosting provider confirmed that:
innodb_strict_mode cannot be disabled.
Server configuration cannot be modified by customers.
**What I tried**

- ROW_FORMAT=DYNAMIC
- Converting many VARCHAR columns to TEXT
- Fresh database
- Fresh installation
- Different Wavelog versions (2.5.1 and 3.0.0)

Converting many VARCHAR fields to TEXT allows the table to be created, but later database migrations fail because some fields (for example COL_POTA_REF) must remain VARCHAR to create indexes.
Therefore, patching the SQL manually is not a viable solution.
****Question**
Has anyone successfully installed Wavelog on MySQL 8.0.46 with innodb_strict_mode enabled?
Is there an official fix planned for the TABLE_HRD_CONTACTS_V01 schema to make it compatible with newer MySQL installations?**

Best Regards
_73 IU7WCE Marco_</div>

HB9HIL:
fixes #3408 

TLDR:

Issue is the innodb_strict_mode which is enabled at some hosters with mysql 8 databases. In this case the db page size is limited to 8kB. the qso table is pretty wide (a lot of columns) and a lot of the varchar columns (there are a lot of them) are smaller than 64 chars. varchar columns <64 can't be store off page by the mysql database, but >64 can by using just a small pointer. By widen a lot of varchar columns to 64 chars we can get the whole page size for the table smaller than 8kB and the issue that Wavelog can't be installed with stric mode enabled is fixed. 

This migration will cause some load on big databases with a lot of QSOs. I tested it in various setups with at max 500k QSOs in the database and it works fine.


int2001 (approved):
seems good/reasonable

phl0 (approved):
Works for me. Tested:
- upgrading running installation
- completely new installation

PR #3412, opened 2026-07-02
Feature/custom map theme style（Customize for each user）

wohenbuguai:
To begin with, I discovered during use that the current map is tightly coupled with the existing theme, making it impossible to modify styles independently to achieve decoupling; furthermore, I wanted to use light-colored map tiles within a dark theme. This was my original motivation for developing this module.

Implementation method: This is achieved by using CSS rules with higher specificity than the existing styles. The overall architecture facilitates future extensibility and adheres to the principle of on-demand loading; to implement it, you simply need to define a new ID in the corresponding configuration file and create a CSS file with the matching name in the `assets/custom_map_style` directory.（PS：Regular testing confirms it works well across various map modules.）
workflow

Final result:

This is the first PR I’ve submitted, so I might be a bit inexperienced or overly wordy. Please feel free to point out any areas that need improvement—thank you very much.

### DIFF
--- a/application/config/autoload.php
+++ b/application/config/autoload.php
@@ -64,7 +64,7 @@ $autoload['libraries'] = array('database', 'session', 'curl', 'OptionsLib', 'Pat
 |	$autoload['helper'] = array('url', 'file');
 */
 
-$autoload['helper'] = array('url', 'security', 'language', 'club');
+$autoload['helper'] = array('url', 'security', 'language', 'club', 'custom_map_style');
 
 
 /*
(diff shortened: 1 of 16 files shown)


AndreasK79 (approved):

AndreasK79:
Hi, just tested this PR. I have no trouble here with it, and I can switch map themes without any problems. I can understand that some people do not want dark maps with dark themes, so this solves that problem.

Good work!

int2001 (approved):
Well done.

PR #3414, opened 2026-07-02
[User settings] adding a search for the page

AndreasK79:
This PR adds a search in the user settings page. Works with all languages.
image

- [x] Add a search input in settings
- [x] Refactor to use gettext/po instead of mirroring/cloning the strings


phl0 (approved):
works fine.

HB9HIL (approved):

PR #3426, opened 2026-07-04
Accessibility (Start)

int2001:
Add aria-labels for screenreaders to:

- Dashboard
- Headers
- QSO-Form
- log_partial

Testing hints:
- test the above mentioned pages. nothing should've changed
- focus on elements which really changed (dividers in header // some links converted to buttons, because screenreaders don't like them)


phl0 (approved):

phl0 (approved):
Still works.

PR #3433, opened 2026-07-06
Fix annoying bug for GHz-QSOs on mobile

int2001:
These bugs are there, since we have the [switch](https://github.com/wavelog/wavelog/commit/abf1e431ca2c6367a3963578a47dc90713f0dbe8) for kHz, MHz, GHz in QSO-Form.

image

repro Bug 1:
- open logging page from mobile (iPhone)
- recall QO100 Fav or prepare logging for QO100 (e.g.: `2.400175` GHz)
try to log.
it fails with an Error, because stepsize is too small.

repro Bug 2:
- open logging page from mobile (iPhone) which is set to ENGLISH (OS-Language of mobile!)
- try to type in 2.400175 GHz // doesn't work, because English-fraction is a comma, not a dot. which isn't shown on android or apple.
Workaround: replace via regex to dot.

Testing:
- manual entering QRG with different units
- CAT
- Cluster check (follow QRG)

### DIFF
--- a/assets/js/sections/contesting/contest_engine/components/radio.js
+++ b/assets/js/sections/contesting/contest_engine/components/radio.js
@@ -401,20 +401,8 @@ class RadioComponent {
 	 * Setup QRG input type for small screens
 	 */
 	qrg_inputtype() {
-		// on small screens we change the input type of the frequency and frequency_rx fields to number to show the numeric keyboard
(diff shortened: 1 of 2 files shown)


phl0 (approved):
Both bugs could be confirmed on an iPhone with Safari browser. Patch fixes the problem(s).

PR #3434, opened 2026-07-06
autoload the user_model

HB9HIL:
The 'Hams of Note' update was broken, because the Quick Theme Switcher relies on the user_model in the header (which is loaded on every page). The user_model is necessary for everything in every model. Therefore it makes sense to autoload it. 

With an autoloaded user_model the HON Update is fixed aswell.


int2001 (approved):
since it is needed everywhere, go ahead

PR #3438, opened 2026-07-07
Hide station export button from unallowed users

kc9uhi:
Station location listing (stationsetup/list_locations) displayed "Export All Locations" to all users, regardless of their authorization level. Station location export is blocked to Clubmember and Clubmember ADIF (level 3,6) by the controller. Clicking the button by a level 3 or 6 user returned a generic failure message as a result.

This PR moves the user level check to also include the export button with the import button. Both buttons are hidden to users without permissions to export/import.

### DIFF
--- a/application/views/stationsetup/locationlist.php
+++ b/application/views/stationsetup/locationlist.php
@@ -128,6 +128,7 @@
 						$(node).removeClass('dt-button').addClass('btn btn-primary');
 					},
 				},
+				<?php if (!($cd_p_level == 3) && !($cd_p_level == 6)) { // ClubOfficer (9) and normal User can export/import, while ClubMember (ADIF) (3,6) can only see. ?>
 				{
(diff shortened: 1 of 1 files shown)


int2001:
<del>i am very sorry. but i am not a friend of changing the behaviour of the API at the moment.</del>

<del>Reasons:
<del>- There are already 3rd party services who use the API even as ClubMember. If someone uses an API the assumption is: He knows what he's doing.
<del>- Changing the behaviour would break 3rd party applications

<del>The UI is a different thing. Normally here are normal Users clicking around.

<del>This may be something relevant for APIv2 (which is planned for future). Yes UI is inconsistent with API here, but before it breaks already running applications, i suggest to leave it as it is.</del>

sorry. i misunderstood the PR - it's only about UI. forget what i am saying

int2001 (approved):

kc9uhi:
Yes, this is the alternate option to #3378

HB9HIL (approved):

PR #3449, opened 2026-07-08
Try to fix grid corner-activations

int2001:
Fixes an issue where QSOs are plotted at the wrong place, when working muilti-grid-activator.

issue:
- log QSO with someone on a gridcorner. e.g.: NN24,NN25,NN14,NN15 --> it is plotted southwest from corner.
- log QSO with someone on a gridline. e.g. JO41,JO40 --> it is plotted a bit south from gridline

this patch should fix it.

test and doublecheck all grid-combinations (even single grid)


phl0 (approved):
Tested and fixes the glitches for grid corners here.

PR #3450, opened 2026-07-08
Add "OK-OM SSB Liga" to contest list

tuxmartin:
Adds the Czech-Slovak SSB League (https://ssbliga.nagano.cz/) to the contest table via migration 287, using ADIF contest id OK-OM-SSB-LEAGUE.

### DIFF
--- a/application/config/migration.php
+++ b/application/config/migration.php
@@ -22,7 +22,7 @@ $config['migration_enabled'] = TRUE;
 |
 */
 
-$config['migration_version'] = 286;
+$config['migration_version'] = 287;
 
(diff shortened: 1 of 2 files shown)


HB9HIL:
Hello @tuxmartin 

Unfortunately this can't be accepted 

1) PR's should created against dev branch,
not master
2) We stick to the official adif contest id list https://www.adif.org/317/ADIF_317.htm#Contest_ID
If a contest is not on the list choose "Other" or add it to the contestlist on your instance in the admin menu by yourself.

But thx for your motivation

PR #3453, opened 2026-07-08
Drop htmx

HB9HIL:
We talked a lot about this in the past. Since I'm working on worker implementations it's a good moment to drop the htmx lib. it's used in 4 places which needs to be tested (I already did but for you I list them again):

- [x] Radio Display Component on the dashboard
- [x] QSO Table in QSO logging view (last 5)
- [x] DCL manual upload
- [x] LotW manual upload


int2001 (approved):
Tested:
- [x] DCL Upload
- [x] LoTW Download
- [x] qso / last-5
- [x] Radio-polling

phl0 (approved):
Works (tested all except radio - no CAT)

PR #3459, opened 2026-07-09
Only count LoTW confirmations for VUCC award calculations

phl0:
This removes the (paper) QSL cards from the VUCC count calculations. Even if cards are to be accepted those should be checked by VUCC card checker and then returned as confirmed via LoTW. So we shall not count locally available (and unchecked) paper QSLs. 
(This PR also refactors some whitespace in VUCC model)

### DIFF
--- a/application/models/Vucc.php
+++ b/application/models/Vucc.php
@@ -175,7 +175,7 @@ class VUCC extends CI_Model
     }
 
     /*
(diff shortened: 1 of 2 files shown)


phl0:
As even the grids checked by card checkers are not returned as confirmed by LoTW (to our knowledge) this PR does not make sense. So we will skip it (for now).

PR #3460, opened 2026-07-09
Qslqueue redesign

AndreasK79:
Instead of the button chaos at the bottom, add one button row, with dropdowns.
Also add menu to be able to print qslcards or label.
image
Todos:
- [x] Redesign button chaos
- [x] Added print QSL card menu item (with a dialog, instead of sending you to another page)
- [x] Added label print from menu item
- [x] Need to fix label print for all
- [x] Add check to see if anything is selected instead of showing dialog
- [x] Need to fix "Mark qsos as printed" for label and qslcard
- [x] Add title to dialog header
- [x] Add filters in the table (instead of function for marking in the table)


AndreasK79:
@DB4SCW I know you have done some work on this page previously. This is a redesign, much more clean. Having filters in the table, the buttons grouped, and being able to print qsl/label from the button menu (to improve the whole process of qsling).

Can you test it out and give some feedback here?

AndreasK79:
> @AndreasK79 Just from initial viewing: I can't seem to find the functionality to mark QSOs of a specific sending method anywhere

Use the filters, then click on the check box to mark all.

DB4SCW:
Okay... gotcha.

My remarks:

- Filter and then clicking the "select all" box works. Maybe a text indicating that selections "survive" filtering might be good to make that more clear?
- The options in the "mark" dropdown are really close together (physically). I can see some people (even me) by accident clicking "mark all requested" instead of "mark selected" with no easy way of rolling back those changes. Maybe splitting those options into 2 different buttons with maybe a yellow color for the "mark requested" option might be an idea to indicate this option does not respect selections?
- If we lean into the filters in the table directly, there is an argument to be had to remove the "station profile" selection box (which is still on top now) and add that filter into the corresponding column, too, just to make it more consistent.

As far as I can see the rest of the functionality works, especially the additional methods to filter callsigns or dates are really nice upgrades!

HB9HIL:
My 2cts:

>     * Filter and then clicking the "select all" box works. Maybe a text indicating that selections "survive" filtering might be good to make that more clear? 

mhh yeah I don't think this is a big issue. In this case I'd go with the current situation and see how $users react to it and if there are some more feedbacks

>     * The options in the "mark" dropdown are really close together (physically). I can see some people (even me) by accident clicking "mark all requested" instead of "mark selected" with no easy way of rolling back those changes. Maybe splitting those options into 2 different buttons with maybe a yellow color for the "mark requested" option might be an idea to indicate this option does not respect selections?

I absolutely double that. Both dropdown items are different in their consequence.. @AndreasK79 maybe add a model/alert before mark all as sent "Do you really want to mark ALL qsls as sent?"

>     * If we lean into the filters in the table directly, there is an argument to be had to remove the "station profile" selection box (which is still on top now) and add that filter into the corresponding column, too, just to make it more consistent.

Imho a good idea.

HB9HIL (commented):
did a code and test review to this point. really good job @AndreasK79 
I like this upgrade 💯

application/views/qslprint/index.php
```diff
@@ -20,26 +27,65 @@
...
+						<li><button type="button" class="dropdown-item markallprinted" onclick="markSelectedQsos();" title="<?= __("Mark selected QSOs as sent"); ?>"><?= __("Mark selected QSOs as sent"); ?></button></li>
+						<li><button type="button" class="dropdown-item" onclick="requestedQslAction('qsl_printed')" title="<?= __("Mark QSLs as printed"); ?>"><?= __("Mark requested QSLs as sent"); ?></button></li>
+					</ul>
+				</div>
+
+				<!-- Export -->
+				<div class="btn-group">
+					<button type="button" class="btn btn-sm btn-primary dropdown-toggle" data-bs-toggle="dropdown" aria-expanded="false"><i class="fas fa-file-export me-1"></i><?= __("Export"); ?></button>
+					<ul class="dropdown-menu">
+						<li><button type="button" class="dropdown-item exportselected" onclick="exportSelectedQsos();" title="<?= __("Export selected QSOs to ADIF-file"); ?>"><?= __("Export selected QSOs to ADIF-file"); ?></button></li>
```
HB9HIL: The only thing which catched me is that the dropdown items have a tooltip which contains the same text as the dropdown itself. 😁 

Image

Not sure if this is intended

AndreasK79: @HB9HIL the original view was like that. I just moved the buttons and turned them into a menu. Anyway, they are removed now.

AndreasK79:
@HB9HIL and @DB4SCW I have added menu dividers (like the top menu). Also I added checks for selected rows or not. And finally I have added a dialog to make sure you want to mark ALL.

Regarding station dropdown, I say keep it. The reason is that you can view the QSOs via a link from the labels. It will filter on the selected station from there, but you have no way of selecting another one if you visit via that route.

DB4SCW:
@AndreasK79 I like that compromise. The modal dialogues prevent misclicks effectively.

The point about the view via link from the labels is fair, keeping it that way also hurts no one.

AndreasK79:
@DB4SCW great, we all want a good (and intuitive) user experience. That keeps the support requests away. Also the point was to improve the work flow, and the layout of the page. Having all those buttons at the bottom didn't really have a good feel to it.

HB9HIL (approved):
I really like this one

PR #3463, opened 2026-07-10
Worker live stats

HB9HIL:
Changes the behaviour of the Wavelog Worker Status in the debug view. The status is now shown live via websocket (requires worker version >= 0.2.0). 

It also adds a generic worker.js which can be reused in other parts of Wavelog to make elements worker/ws ready. This does not affect the compatibility in the worker use in contesting.

Next step will be to make elements in Wavelog worker/ws ready

### DIFF
--- a/application/views/interface_assets/footer.php
+++ b/application/views/interface_assets/footer.php
@@ -170,6 +170,13 @@
 </script>
 
 
+<?php if ($worker_enabled ?? false) { ?>
+    <!-- Wavelog Worker JS -->
+    <script>window.WavelogWorker = <?php echo json_encode(['url' => $this->worker->client_url()]); ?>;</script>
+	<script type="text/javascript" src="<?php echo $this->paths->cache_buster('/assets/js/worker.js'); ?>"></script>
+<?php } ?>
+
+
 <!-- DATATABLES LANGUAGE -->
 <?php
 $local_code = $language['locale'];
--- a/application/controllers/Debug.php
+++ b/application/controllers/Debug.php
@@ -130,6 +130,18 @@ class Debug extends CI_Controller
 
 		$data['page_title'] = __("Debug");
 
+		$this->load->library('worker');
+		$data['worker_status_topic'] = '';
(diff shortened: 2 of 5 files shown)


application/views/interface_assets/footer.php
```diff
@@ -170,6 +170,13 @@
 </script>
 
 
+<?php if ($worker_enabled ?? false) { ?>
```
int2001, line 173: Where is `$worker_enabled` set?
i know we've a `$config['worker_enabled']` and a `$worker_enabled` in debug-controller furthermore a `$worker->enabled` (if activated).

or did i miss something here?

HB9HIL: If a page want's to make use of the wavelog worker (expect contesting, which is handled differently) you need this snippet in the controller

```php
$this->load->is_loaded('worker') ?: $this->load->library('worker');
$data['worker_enabled'] = $this->worker->is_enabled();
```

otherwise the worker.js (generic worker handler) is not loaded

https://github.com/HB9HIL/wavelog/blob/ad631d5df28b7c82893650193b704ff9d72a63bb/application/controllers/Debug.php#L133-L136

int2001: Ahhhh. Understand. Makes sense.

HB9HIL: maybe a short explaination:
IMHO its not a good idea (and it does not work properly) to add `$this->load->library('worker');` in the footer.php. The better way would be to add it to the autoload. But since we don't need this on every page it's better to handle that with a variable.. 

so next question: why not loading the script with the footer scripts: because we need to set the window variable in javascript and this can't be done in the controller. 

Maybe there is a better solution but this seemed the smartest one in this case

int2001 (approved):

application/views/interface_assets/footer.php
int2001: once i understood how it works, it's okay in the way you implemented it. ressource-saving ;)

PR #3464, opened 2026-07-10
Radio Display Component on Dashboard - Worker Ready

HB9HIL:
This PR makes the radio display component on the dashboard worker ready. Means if a wavelog worker is in place and configured the polling is disabled and happens only on demand (Cat Model Line 111 already publishes changes in radios). 

> [!IMPORTANT]
> This PR is based on #3463


int2001 (approved):
works.

just note / keep in mind:
if CAT Client pushes to frequently this causes also a "GET" flood (before patch 5s - whatever happened)

assets/js/sections/dashboard.js
```diff
@@ -24,6 +24,25 @@ $(document).ready(function () {
     const loadRadio = () => wlLoadInto(base_url + 'index.php/dashboard/radio_display_component', '#radio_display');
 	loadRadio();
 
+    var pending = null;
+	var missed = false;
+	function throttleLoadRadio() {
+		// Load at most once every 1 seconds. If more pushes arrive during the
+		// lockout, refresh once afterwards so no update is lost.
+		if (pending) {
+			missed = true;
```
int2001, line 33: clever. i ♥

PR #3466, opened 2026-07-10
[QSL Designer] Dialog when changing template

AndreasK79:
image

Fixes #3451.


int2001 (approved):
tested (realigned fields, etc.): works
code: lean.

PR #3475, opened 2026-07-11
QRG column for contesting

HB9HIL:
shows the qrg of the qso in the qso list
- respects the chosen qrg unit in the radio component (per band)
- if qrg is edited, band follows
- if band is edited, qrg follows to default qrg
- qrg has the higher prio if both is edited

### DIFF
--- a/application/controllers/Contesting.php
+++ b/application/controllers/Contesting.php
@@ -787,6 +787,7 @@ class Contesting extends CI_Controller {
 			}
 
 			$this->load->model('contesting_model');
+			$this->load->is_loaded('logbook_model') ?: $this->load->model('logbook_model');
 
 			// Session ownership check
 			if (!$this->contesting_model->check_user_contest($contest_session_id)) {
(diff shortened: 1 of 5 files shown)


int2001 (changes requested):
1. Open a 40m QSO for edit. The frequency field shows 7100 with unit label kHz (= 7,100,000 Hz).
2. Without saving, click the radio component's QRG unit toggle to switch 40m to MHz. radio.js writes qrgunit_40m = 'MHz' to localStorage and dispatches qrgUnitChanged.
4. rerenderVisibleRows() runs but skips the editing row (its dataset.editing === 'true'), so the field still shows 7100 / kHz.
5. Press Enter to save. _parseQrgInput("7100", _qrgUnit("40m")) now parses with the new unit MHz → 7,100,000,000 Hz. convertQrgToBand(7.1e9) returns '??', so the band is left at 40m, but data.frequency is set to "7100000000".
6. Result: the QSO is persisted with COL_FREQ = 7.1 GHz while COL_BAND = 40m.

tbh: imho overengineered. i'd only show the QRG. much easier.

int2001 (approved):

PR #3478, opened 2026-07-12
Fix Error at Useredit

int2001:
Try to change your emailadress (or one from a user) at userdit to one which is already existing.
useredit fails with a lot of (PHP/CI-)warnings.

this one fixes it. clean error to user.

PR is simple.
just moved code


AndreasK79 (approved):

AndreasK79:
Much better :)

PR #3480, opened 2026-07-12
Contest improvements

HB9HIL:
This PR fixes some stuff and adds some improvements. In detail:
- If callbook online is enabled we should log the grid even grid is not part of the exchange, but it was receieved from callbook so why not logging it then
- respect the default radio selected in hardware interfaces
- fixes a bug in preselected radios which was implemented last time
- adds a new option "serial per band" which calculates the serial number by band instead of global. Was requested in #3425 
- redesigned the qrgunit in edit mode of a qso so users don't think it's clickable (tnx @int2001)


int2001 (approved):
lgtm

PR #3482, opened 2026-07-12
Public Punchcard

int2001:
Adds a public punchcard to wavelog.
The punchcard from "Analyze -> Days with QSOs -> QSO of the Year" are now public available, if the User enabled public-slugs at his logbook.

it shows the actual year

it can be reached via `/[slug]/punchcard` in the same way as `/[slug]/mini`.
the punchcard is rendered "standalone" without JS, so one could embed it into - e.g. - qrz.com or other pages.

a theme param can be passed: example: `?theme=darkly`

Going to add documentation after merge

image

### DIFF
--- a/application/views/visitor/punchcard.php
+++ b/application/views/visitor/punchcard.php
@@ -104,7 +104,7 @@ body.dark td.day.empty { background-color: transparent; }
 <div class="pc-container">
   <div class="pc-header">
     <?= __("QSOs per Year"); ?> <?php echo htmlspecialchars($yr, ENT_QUOTES); ?>
-    <span class="pc-quantity"><?php echo (int) $total; ?> <?= __("QSOs"); ?></span>
+    <span class="pc-quantity"><?php echo (int) $total; ?> <?= _ngettext("QSO", "QSOs", $total); ?></span>
   </div>
   <div class="pc-body">
     <table class="pc">
--- a/application/views/visitor/punchcard.php
+++ b/application/views/visitor/punchcard.php
@@ -1,0 +1,173 @@
+<?php $numWeeks = count($weeks); ?>
+<!DOCTYPE html>
+<html lang="en">
+<head>
+<meta charset="UTF-8">
+<meta name="viewport" content="width=device-width, initial-scale=1.0">
+<title><?= __("QSOs per Year"); ?> <?= htmlspecialchars($yr, ENT_QUOTES); ?></title>
+<style>
+* { margin: 0; padding: 0; box-sizing: border-box; }
+body {
+  font-family: Menlo, Consolas, monospace;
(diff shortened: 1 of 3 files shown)


application/views/visitor/punchcard.php
```diff
@@ -0,0 +1,173 @@
...
+              <?php if ($cell === null): ?>
+                <td class="day empty"></td>
+              <?php else: ?>
+                <?php
+                $col = (int) $cell['col'];
+                $titleDate = htmlspecialchars($cell['date'], ENT_QUOTES);
+                $titleCount = (int) $cell['n'];
+                $titleText = $titleCount === 0
+                    ? $titleDate . ': ' . __("No QSOs")
+                    : $titleDate . ': ' . $titleCount . ' ' . ($titleCount === 1 ? __("QSO") : __("QSOs"));
```
HB9HIL: In some languages for example in some southeastern languages there is not just singular and plural but also a third form of describing an amount. Gettext offers a nice way here which we should use. In this case it might be not that important as QSOs is a shortened string, but consistency is also a good thing. 


```suggestion
                    : $titleDate . ': ' . $titleCount . ' ' . _ngettext("QSO", "QSOs", $titleCount);
```

Weblate and gettext then manages the the different kinds of plural forms per language itself ("one", "many" and in some languages also "some").

```php
_ngettext($singular_string,$plural_string,$number_for_decision)
```

application/views/visitor/punchcard.php
```diff
@@ -0,0 +1,173 @@
...
+.pc-legend .s3 { background-color: <?php echo $mode === 'dark' ? '#62a39c' : '#277672'; ?>; }
+.pc-legend .s4 { background-color: <?php echo $mode === 'dark' ? '#4b837e' : '#075652'; ?>; }
+.pc-legend .s5 { background-color: <?php echo $mode === 'dark' ? '#395b58' : '#004642'; ?>; }
+</style>
+</head>
+<body class="<?php echo $mode === 'dark' ? 'dark' : ''; ?>">
+<div class="pc-container">
+  <div class="pc-header">
+    <?= __("QSOs per Year"); ?> <?php echo htmlspecialchars($yr, ENT_QUOTES); ?>
+    <span class="pc-quantity"><?php echo (int) $total; ?> <?= __("QSOs"); ?></span>
```
HB9HIL: comment see above

```suggestion
    <span class="pc-quantity"><?php echo (int) $total; ?> <?= _ngettext("QSO", "QSOs", $total); ?></span>
```

HB9HIL (changes requested):
I love this one! will definitly land on my qrz page. There is just one minor thing regarding usage of gettext :-)

HB9HIL (approved):
🎉😊

Issue #1393, opened 2024-12-26
Run Now Button on Cron Manager Page


ethancedwards8:
**What Feature is Missing?**
A run now button should be placed on the Cron Manager page that allows you to run whatever cron job immediately, so that you don't have to wait for the next run.

**Why is This Feature Important?**
This could be helpful if your server went down for several weeks and it missed the cronjobs. You could force them to run, and everything would be back up to date. 

**Is This Feature Personal or Beneficial to Others as Well?**
I think it would be helpful, especially to test new cronjob functions and the like.

**Implementation**
This file seems like the correct place: https://github.com/wavelog/wavelog/blob/master/application/views/cron/index.php
Would implement a force_run() or similar function in this file. https://github.com/wavelog/wavelog/blob/c0fd01e0010fd63fde9efd4efd341df34e11fba1/application/controllers/Cron.php  Link it to the button.

PR #3483, opened 2026-07-12
Add a Run Now action to Cron Manager

slepp:
## What changed

I added a Run Now action for enabled cron jobs in Cron Manager. This is related to #1393 - it dispatches the configured job through its existing internal path instead of waiting for the next master-cron tick.

When master cron is not running, the existing warning and setup guidance remain visible above the job table so an administrator can recover individual jobs. A per-job lock prevents manual and scheduled runs from overlapping; an occupied job returns a warning. Manual actions report dispatch status without changing the scheduler health tracking.

## Testing

- `git diff --check`
- `node --check assets/js/sections/cron.js`

### DIFF
--- a/application/controllers/Cron.php
+++ b/application/controllers/Cron.php
@@ -166,6 +166,9 @@ class cron extends CI_Controller {
 			echo json_encode(['success' => false, 'messagecategory' => 'error', 'message' => __("You're not allowed to do that!")]);
 			return;
 		}
+		// Release the session lock: the job runs synchronously and would otherwise
+		// block every other request of this user (with the file session driver) until it finishes.
+		session_write_close();
 
 		$cron = $this->cron_model->cron($this->input->post('id', true))->row();
 		if ($cron === null) {
@@ -337,14 +340,20 @@ class cron extends CI_Controller {
 		}
 
 		// Prevent manual and scheduled executions of the same cronjob from overlapping.
-		$lock_handle = fopen(sys_get_temp_dir() . '/wavelog_cron_' . md5(FCPATH . ':' . $cron->id) . '.lock', 'c');
-		if ($lock_handle === false) {
-			return ['success' => false, 'locked' => false, 'response' => '', 'error' => 'Unable to acquire cronjob lock.', 'url' => $url];
-		}
-		if (!flock($lock_handle, LOCK_EX | LOCK_NB)) {
-			fclose($lock_handle);
+		// Load cache driver for locking mechanism.
+		$this->load->driver('cache', [
+			'adapter' => $this->config->item('cache_adapter') ?? 'file',
+			'backup' => $this->config->item('cache_backup') ?? 'file',
+			'key_prefix' => $this->config->item('cache_key_prefix') ?? ''
+		]);
+		$lock_key = 'cron_lock_' . md5($cron->id);
+		if ($this->cache->get($lock_key)) {
 			return ['success' => false, 'locked' => true, 'response' => '', 'error' => '', 'url' => $url];
 		}
+
+		// Lock for 12 hours to prevent overlapping executions of the same cronjob.
+		// some big instances have really really long cron runs
+		$this->cache->save($lock_key, time(), 43200);
 
 		$ch = curl_init();
 		curl_setopt($ch, CURLOPT_URL, $url);
@@ -357,8 +366,8 @@ class cron extends CI_Controller {
 		$response = curl_exec($ch);
 		$error = $response === false ? curl_error($ch) : '';
 
-		flock($lock_handle, LOCK_UN);
-		fclose($lock_handle);
+		// invalidate the cache so the job can run next time
+		$this->cache->delete($lock_key);
 		return [
 			'success' => $response !== false,
 			'locked' => false,
--- a/application/controllers/Cron.php
+++ b/application/controllers/Cron.php
@@ -109,36 +109,19 @@ class cron extends CI_Controller {
 							echo "CRON: " . $cron->id . " -> is due: " . $isdue_result . "\n";
 							echo "CRON: " . $cron->id . " -> RUNNING...\n";
 
-							if (ENVIRONMENT == "docker") {
-								// In Docker, we use the localhost[:80] to call the cron directly inside the container
-								$url = 'http://localhost/' . $cron->function;
-								log_message('debug', 'Docker Environment detected. Using URL: ' . $url);
-							} else {
-								// In other environments, we use the local_url() function to get the default url
-								// Even this local_url() helper created in https://github.com/wavelog/wavelog/pull/795 is not really necessary anymore
-								// we keep it in case of users do fancy things with it. It doesn't hurt to have it here as it usually returns the base_url
-								// from the config.php file.
-								$url = local_url() . $cron->function;
-							}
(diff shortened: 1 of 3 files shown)


application/controllers/Cron.php
```diff
@@ -302,6 +328,46 @@ private function cronEnabled2html($id, $enabled) {
...
+
+		flock($lock_handle, LOCK_UN);
+		fclose($lock_handle);
+		return [
+			'success' => $response !== false,
+			'locked' => false,
+			'response' => $response,
+			'error' => $error,
+			'url' => $url
+		];
```
HB9HIL, line 377: A lock mechanism is awesome. But in different environments it can be an issue to use fopen/lock/close. We made much better experience by using the internal cache mechanism for locking stuff down. Check my proposal


```suggestion
		// Prevent manual and scheduled executions of the same cronjob from overlapping.
		// Load cache driver for locking mechanism.
		$this->load->driver('cache', [
			'adapter' => $this->config->item('cache_adapter') ?? 'file',
			'backup' => $this->config->item('cache_backup')	 ?? 'file',
			'key_prefix' => $this->config->item('cache_key_prefix') ?? ''
		]);
		$lock_key = 'cron_lock_' . md5($cron->id);
		if ($this->cache->get($lock_key)) {
			return ['success' => false, 'locked' => true, 'response' => '', 'error' => '', 'url' => $url];
		}

		// Lock for 12 hours to prevent overlapping executions of the same cronjob.
		// some big instances have really really long cron runs
		$this->cache->save($lock_key, time(), 43200); 

		$ch = curl_init();
		curl_setopt($ch, CURLOPT_URL, $url);
		curl_setopt($ch, CURLOPT_HEADER, false);
		curl_setopt($ch, CURLOPT_USERAGENT, 'Wavelog Updater');
		curl_setopt($ch, CURLOPT_RETURNTRANSFER, true);
		if ($this->config->item('cron_allow_insecure') ?? false == true) {
			curl_setopt($ch, CURLOPT_SSL_VERIFYHOST, false);
		}
		$response = curl_exec($ch);
		$error = $response === false ? curl_error($ch) : '';

        // invalidate the cache so the job can run next time
		$this->cache->delete($lock_key);
		return [
			'success' => $response !== false,
			'locked' => false,
			'response' => $response,
			'error' => $error,
			'url' => $url
		];
```

application/controllers/Cron.php
```diff
@@ -177,6 +160,42 @@ public function run() {
...
+		$cron_result = $this->triggerCron($cron);
+		header("Content-type: application/json");
+
+		if ($cron_result['locked']) {
+			echo json_encode(['success' => false, 'messagecategory' => 'warning', 'message' => __('Cronjob is already running.')]);
+			return;
+		}
+
+		if ($cron_result['success']) {
+			echo json_encode(['success' => true, 'messagecategory' => 'success', 'message' => __('Cronjob triggered.')]);
```
HB9HIL, line 195: There seems to be an UI bug. The notification banner appears in green as it should be after triggering a job. When I hit the same button in a private browser window (another php session) and directly in the first window, the warning will be yellow that the job is already running. This is correct. But hitting it again it shows "Cronjob triggered" in orange. Seems like the class isn't properly reset.

HB9HIL (changes requested):
looks not bad so far :) check my comments after this first round.

application/controllers/Cron.php
slepp: This makes sense, thanks for the suggestion on that. I'll integrate that flow.

application/controllers/Cron.php
slepp: I'll see if I can get that to happen over here as well!

slepp:
@HB9HIL Thanks for the prior review, I made some changes here for the CSS colouring/updates and now using the cache to lock instead of `flock`.

HB9HIL (approved):
Looks good to me. :-) thx for your contribution!

application/controllers/Cron.php
```diff
@@ -302,6 +331,52 @@ private function cronEnabled2html($id, $enabled) {
...
+			'key_prefix' => $this->config->item('cache_key_prefix') ?? ''
+		]);
+		$lock_key = 'cron_lock_' . md5($cron->id);
+		if ($this->cache->get($lock_key)) {
+			return ['success' => false, 'locked' => true, 'response' => '', 'error' => '', 'url' => $url];
+		}
+
+		// Lock for 12 hours to prevent overlapping executions of the same cronjob.
+		// some big instances have really really long cron runs
+		$this->cache->save($lock_key, time(), 43200);
```
int2001, line 356: A killed/died cron process locks out that job for 12 hours

Unlike the previous flock(LOCK_EX | LOCK_NB) (which the OS releases the instant the holding process dies), the cache lock is only freed by the explicit $this->cache->delete() at the end. If PHP dies
mid-curl_exec — max_execution_time, OOM, a deploy, or a Docker container restart — delete never runs and the 12h TTL takes over, so every subsequent scheduled run and every "Run Now" sees the lock and skips until the TTL expires. This hits hardest on exactly the deployments this targets (long-running crons are the most likely to be interrupted), and it's worse on Docker.

Possible mitigations:
- wrap the curl in try { ... } finally { $this->cache->delete($lock_key); } to at least cover exceptions, and/or
- store a started-at timestamp and treat a stale lock as releasable, or
- keep flock (auto-released on process death, self-heals within one cron tick).

application/controllers/Cron.php
```diff
@@ -302,6 +331,52 @@ private function cronEnabled2html($id, $enabled) {
...
+		}
+
+		// Prevent manual and scheduled executions of the same cronjob from overlapping.
+		// Load cache driver for locking mechanism.
+		$this->load->driver('cache', [
+			'adapter' => $this->config->item('cache_adapter') ?? 'file',
+			'backup' => $this->config->item('cache_backup') ?? 'file',
+			'key_prefix' => $this->config->item('cache_key_prefix') ?? ''
+		]);
+		$lock_key = 'cron_lock_' . md5($cron->id);
```
int2001, line 349: The old flock key was `md5(FCPATH . ':' . $cron->id);`
while  the new cache key is `cron_lock_' . md5($cron->id)` with **no** install discriminator. 

It only relies on `$config['cache_key_prefix']`, which defaults to '' when unset. 

There's a (small) collision issue.

application/controllers/Cron.php
HB9HIL: @int2001 I'll fix that myself after this one is merged. There are some other things I want to add which are out of scope of this PR

application/controllers/Cron.php
HB9HIL: Solution: Set a `cache_key_prefix` to avoid collisions. This is exactly why this option exists

int2001: or some random generated static id per instance. URL, path or something like that.
just in case. doesn't hurt and makes it distinguishable

application/controllers/Cron.php
int2001: ok, but should happen immediatly after merge

int2001 (approved):
fine, but needs some afterwork as mentioned by @HB9HIL 
so merge if the afterworks happens soon.

PR #3497, opened 2026-07-15
Make grid in station location optional

phl0:
Gridsquare is currently mandatory in station location setup. There is no way to allow for example LoTW uploads without gridsquare (which is possible and not prohibited). So we remove the 'required' tag from the input field and allow '' as grid in check function.


int2001:
Till here reviewed. looks good so far. wasn't able to spot a problem

phl0:
Tnx fer testing. So lets see what Mr. Map @AndreasK79 has to say :D

AndreasK79 (approved):

application/views/satellite/flightpath.php
```diff
@@ -38,6 +38,9 @@
 			<?= __("View a map with satellite flightpaths"); ?>
 		</div>
 		<div class="card-body">
+			<?php if ($homegrid == '') { ?>
+				<p><span class="badge text-bg-warning"><?= __("Warning"); ?></span> <?= __("Active station location does not have a gridsquare configured. Please check"); ?>
```
int2001: close of `<p>` seems missing

application/views/satellite/pass.php
```diff
@@ -24,6 +24,9 @@
 
 	<!-- Station Cards Row -->
 	<div class="row gx-2">
+	<?php if ($activegrid == '') { ?>
+		<p id="nogridhint"><span class="badge text-bg-warning"><?= __("Warning"); ?></span> <?= __("Active station location does not have a gridsquare configured. Please check"); ?>
```
int2001: close of `<p>` is missing

int2001:
and: looks like LBA-Map doesn't draw QSOs without grid (but DXCC) any longer. intended?
even if own grid is set...

phl0:
> and: looks like LBA-Map doesn't draw QSOs without grid (but DXCC) any longer. intended? even if own grid is set...

Fixed.

int2001 (approved):

PR #3498, opened 2026-07-15
Remove files not in use anymore

AndreasK79:
These files seem pretty dead to me.


int2001 (approved):
can't spot any callers for the removed files.

PR #3499, opened 2026-07-15
Remove orphaned functions

AndreasK79:
This PR contains only removed functions from models, where a reference could not be found.


int2001 (approved):
cannot spot any callers for the removed functions

PR #3502, opened 2026-07-16
After several enhancements, several functions were orphaned

AndreasK79:
These were found not be in use anymore.


HB9HIL (approved):
I can confirm, this is dead code

PR #3506, opened 2026-07-16
Contest import enhancements

HB9HIL:
This PR does two things:

1. The legacy importer which imports historical contests from the log was built to import contests which happen once a year. This is bullsh**. This changes the logic a bit so the importer creates a new session if there was no qso for 72h after the last one.
2. some UI enhancements in the contest session management


int2001 (approved):
new logic is new...
good one to catch edge-cases

PR #3509, opened 2026-07-17
Fix no qrg

int2001:
Consistency with UI during ADIF Import or API-logging

- band + freq provided ==> pass
- freq only ==> derive band from freq ==> pass
- band only ==> derive freq (new, as in UI) ==> pass
- band only, but crappy band ==> reject import
- neither band, nor frequency ==> reject import

Bonus:
- band_rx softfail (derive freq_rx if possible) ==> pass
- some errormessages were placed as "dead-code". fixed

### DIFF
--- a/application/libraries/Frequency.php
+++ b/application/libraries/Frequency.php
@@ -160,6 +160,10 @@ class Frequency {
 
 		$result = $db->get()->row();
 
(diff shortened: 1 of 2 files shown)


phl0 (approved):
Tested and works as designed.

PR #3513, opened 2026-07-17
Api v2

HB9HIL:
This PR introduces the new **REST API v2**. It runs alongside the existing v1 API and is **not** a replacement, ensuring that existing integrations continue to work while new clients can take advantage of a more modern interface.

Documentation preview and to have it easier playing with this PR is here: https://hb9hil.github.io/wavelog_docs/developer/api-v2/

### Highlights

- RESTful resource-based API (`/api/v2/...`) instead of RPC-style endpoints
- Support for `GET`, `POST`, `PUT`, `PATCH`, and `DELETE`
- Bearer token authentication (`Authorization: Bearer wl2_...`)
- API v2 tokens are stored as SHA-256 hashes instead of plaintext
- Fine-grained permission model using scopes (e.g. `qso:read`, `qso:write`, `station:read`)
- Consistent response and error format across all endpoints
- Standardized HTTP status codes and error handling
- Pagination support for list endpoints
- Supports internal already existing rate limiter
- Public `/api/v2/status` endpoint for health check

### Available resources yet

- QSO
- Station
- Radio
- Statistic
- Lookup
- Club
- Token

### Key differences from API v1

- API v2 uses dedicated tokens with the `wl2_` prefix. Legacy v1 API keys are **not** accepted on the new endpoint.
- Authentication is performed via HTTP headers instead of the request body.
- Permissions are scope-based rather than simple read/write access.
- Tokens can (and should) expire
- Tokens are only stored with a hash and only shown once in the UI
- `PATCH` performs partial updates, while `PUT` replaces the entire resource.


application/views/api/index.php
```diff
@@ -3,12 +3,95 @@
...
+									} else {
+										echo $row->expires_at;
+									} ?>
+								</td>
+								<?php if ($clubmode) { ?>
+									<td><?php echo $row->user_callsign ?? ''; ?></td>
+								<?php } ?>
+								<td>
+									<?php
+										$cfnm_delete_token = sprintf(__("Are you sure you want delete the API Token %s?"), '&quot;' . htmlspecialchars($row->token_name, ENT_QUOTES, 'UTF-8') . '&quot;');
```
int2001: xss possible here, since quotes are not escaped. 
possible attack-vector: create token as clubmember, craft (clever) tokenname, clubofficer deletes token and JS is executed.

PoC (not complete exploit): ` test');` as token-name and see generated HTML/JS

application/models/Logbook_model.php
int2001:
```suggestion
			$this->notify_qso_change($user_id ?? $this->session->userdata('user_id'));
```
since here's a new parameter called `$user_id`, it should imho be taken for notification.

application/models/Logbook_model.php
```diff
@@ -3709,6 +3864,35 @@ function total_bands($dateFrom = null, $dateTo = null) {
...
+			$StationLocationsArray = $this->logbooks_model->list_logbook_relationships($this->session->userdata('active_station_logbook'));
+		}
+		$logbooks_locations_array = $StationLocationsArray;
+
+		if ($logbooks_locations_array[0] === -1) {
+			return null;
+		}
+
+		$this->db->select('COL_MODE AS mode, count( * ) AS count', FALSE);
+		$this->db->where_in('station_id', $logbooks_locations_array);
```
int2001, line 4040: tbh: would refactor this to proper binding and prepared SQLs (like we did before at so many places).
at the end up2u

int2001:
Checked code - lgtm
Checked endpoints - also looking good and working.

minor issue:

when calling endpoint with invalid token, php spits out a warning:

`Severity: Deprecated --> filter_var(): Passing null to parameter #3 ($options) of type array|int is deprecated /var/www/html/system/core/Input.php 515`

int2001 (approved):

int2001 (approved):

PR #3516, opened 2026-07-18
Call validation

HB9HIL:
soft input validation for callsigns. Used to catch malicious stuff rather then really validatin for proper amateur radio callsigns. It's still shit in/shit out, but with a fence for malicious injections


int2001 (approved):
good fence

PR #3523, opened 2026-07-19
Make sure oqrs belongs to user

HB9HIL:
Make sure the oqrs which gets marked as done belongs to user. 3yo untouched code


phl0 (approved):
Looks good from reading the code :)

PR #3531, opened 2026-07-20
[GeoJSON Map]  Updated page looks

AndreasK79:
User menu -> internal tools -> GeoJSON QSO Map

To activate, add this in the config.php:
> $config['internal_tools'] = true;

This PR does the following:

- [x] Updated looks to match other pages
- [x] Show all countries with geojson files in the dropdown, regardless if a qso exists in the log with a six char gridsquare
- [x] Show only gridsquare for the selected dxcc (like the gridsquare map)
- [x] Made infobar below map responsive
- [x] Show QSO info on mouseover instead of click
- [x] Autoselect active location in the dropdown
- [x] Add missing translation strings
- [x] Added worked/confirmed regions
- [x] Added band filter
- [x] Need to fix band when sat
- [x] Fix custom date
- [x] Add QSO lists

Should we move this page to the normal pages, so that it is not hidden behind a config switch? If so, where should we place it? It was more meant as an internal tool, however it can be used to look up which county/state/province etc you are in as well.

### DIFF
--- a/application/controllers/Map.php
+++ b/application/controllers/Map.php
@@ -30,12 +30,12 @@ class Map extends CI_Controller {
 		$this->load->model('Map_model');
 		$this->load->model('stations');
 
-		// Get supported DXCC countries with state data
+		// Get supported DXCC countries (those with GeoJSON boundary data)
 		$data['supported_dxccs'] = $this->geojson->getSupportedDxccs();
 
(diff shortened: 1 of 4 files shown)


phl0 (approved):
Looks and works fine.

PR #3534, opened 2026-07-22
simplify

HB9HIL:
The contesting token which is applied in the URL when starting the new contesting engine was initially implemented to possibly use it for other data or authentication. Since the development of the new contesting engine was more streamlined and closer on the core it was basically without any use and the controller checked session for authentication. I left it because it worked and didn't want to change a running system. 

Now it popped up, that depending on the payload the base64 encoding does use `=` and other characters for padding. I expected that and encoded them. But it turns out it does not work as expected and it may happen, that base64encode uses characters which are not included in CI3 allowed chars list. 

Therefore we just use the contesting sesssion ID for validation. Simple and easy

tnx to @DB4SCW for this heads up


int2001 (commented):
lgtm

PR #3537, opened 2026-07-24
Station accessibility

HB9HIL:
we should check if the station is accessible


HB9HIL:
redundant. model checks that

PR #3542, opened 2026-07-26
Layout update for calltester and zonechecker

AndreasK79:
Updated the layout to match other pages in Wavelog.
Added a few more columns in the result table.
Added a qso list dialog.


int2001 (approved):
lgtm

PR #3543, opened 2026-07-26
[Gridsquare Lookup] Enter a gridsquare and the map zooms

AndreasK79:
image
image

- Added zoom on click
- Clear button zooms out
- Added a second gridsquare input, now shows line, distance and bearing
- Show CQ/ITU Zone information for gridsquare
- Show distance in users selected unit


int2001:
so far checked and working.
great tool for planning `/p`

int2001 (approved):

PR #3553, opened 2026-07-29
Installer guard fix

HB9HIL:
Due to #3533 it can happen that the installer fails. This one fixes this and also does a bit more defense in depth for some edge cases :-)


int2001 (approved):
reviewed code and tested docker-install.
works

PR #3556, opened 2026-07-30
[Contesting] Serial Reservation in MultiOP Sessions

HB9HIL:
The new contesting lacks an important feature. In multi-operator sessions we need a serial reservation logik to block serials and prevent the same serial exchanged twice in qsos. This PR adds a sophisticated serial reservation logik. I'll also add some documentation in docs (once done, I'll comment the link here in this PR)

### DIFF
--- a/application/controllers/Contesting.php
+++ b/application/controllers/Contesting.php
@@ -460,6 +460,12 @@ class Contesting extends CI_Controller {
 			redirect('contesting');
 		}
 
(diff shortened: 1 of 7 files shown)


int2001 (approved):
works.

one info: as far as i tested: works with new contests, not old ones.

HB9HIL:
Documentation: https://docs.wavelog.org/user-guide/contesting/logging/?h=serial+re#serial-number-reservation

PR #3560, opened 2026-08-01
[API v2] Extensions of QSO and Statistic Ressource

HB9HIL:
This PR extends the QSO and Statistics ressource with:

- qso endpoint now accepts callsign filter
- qso endpoint now accepts qso_since and qso_until filter
- statistic endpoint now accepts qso_until filter (qso_since was already there)

Since the function had a lot of parameters I simplified that by using an array instead which results in an more easy to maintain code

### DIFF
--- a/application/models/Logbook_model.php
+++ b/application/models/Logbook_model.php
@@ -2300,13 +2300,14 @@ class Logbook_model extends CI_Model {
 	// appended in placeholder order.
 	//
 	// $station_ids  int[]   station-location ids (already ownership-checked)
+	// $callsign     string  '' for none, else an exact match on COL_CALL
 	// $band         string  '' for none, 'SAT', or a COL_BAND value
 	// $mode         string  '' for none, else a COL_MODE or COL_SUBMODE value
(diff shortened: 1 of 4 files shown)


application/models/Logbook_model.php
```diff
@@ -2299,55 +2299,73 @@ public function num_fields() {
...
 		// Compared uppercased, the same way the ADIF export filters it.
-		if ($operator !== null && $operator !== '') {
+		if (($filters['operator'] ?? '') !== '') {
 			$where .= " AND upper(qsos.`COL_OPERATOR`) = ?";
-			$bindings[] = strtoupper($operator);
+			$bindings[] = strtoupper($filters['operator']);
+		}
+
+		if (($filters['callsign'] ?? '') !== '') {
+			$where .= " AND upper(qsos.`COL_CALL`) = ?";
```
int2001: COL_CALL has index on Log-table.
and: COL_CALL is always already stored uppercase.

this one would bypass the index. recommendation: drop `upper(qso.COL_CALL)` to plain `COL_CALL`

int2001 (changes requested):
well done. only one little thing. see comment

int2001 (approved):

PR #3565, opened 2026-08-01
Apiv2 club permissions

HB9HIL:
Refactored the Club API v2. Documentation is already pushed since API v2 is not released yet https://docs.wavelog.org/developer/api-v2/club/
Short summary:
- allows now to manage member permissions
- admins can list all clubs and manage member permissions even without being a member (just like in the ui)
- club scope is unavailable for normal users (was a bug)

### DIFF
--- a/application/controllers/Api.php
+++ b/application/controllers/Api.php
@@ -38,7 +38,7 @@ class API extends CI_Controller {
 
 		$data['api_keys'] = $this->api_model->keys();
 		$data['api_tokens'] = $this->api_v2_model->get_tokens_for_user();
-		$data['token_scopes'] = Api_v2_model::scope_registry();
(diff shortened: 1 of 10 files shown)


int2001:
Currently testing. looks good so far:

but: since WL is used by "legal-entities" (like clubs), we need a switch for member-info / member-list to be GDPR-Compliant. Even it doesn't make sense from a logical-aspect. it's law and i don't like it as well.

so in detail a config-switch is needed which reduces the output (when true) to a minimum:

- only ID and call in `GET /api/v2/club` (even that one could be tricky from a legal-perspective. but it's needed otherwise its totally senseless)
- only ID, Call and Name in `GET /api/v2/club/{user_id}`

int2001 (approved):

PR #3567, opened 2026-08-01
Fix Azimuth (SAT) when call is entered

int2001:
Bug:

When working SAT, Az/Ele are calculated (and updated) every second based on "startQSO"-Time.
if user has enabled Start and End QSO-Time, the ticker stops when entering a call (because that's the startTime of the QSO).

this fix make the source for time independent of the start-time. So calculation moves on, even if call has been typed in/looked up.

No impact on:
- logging (SAT or HF)
- HF-Bearing
- Post QSO-Logging (for PostQSO the time of QSO counts)

### DIFF
--- a/assets/js/sections/qso.js
+++ b/assets/js/sections/qso.js
@@ -1059,16 +1059,8 @@ function start_az_ele_ticker(tle) {
 	};
 
 	function updateAzEl() {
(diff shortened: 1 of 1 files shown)


HB9HIL (approved):
lgtm, not tested but code looks good

PR #3574, opened 2026-08-03
Sota pota db map

AndreasK79:
Bases on #3572, so that needs merging first.

Changed updates for POTA and SOTA to use a db table, so that we can plot reference in the Gridsquare Lookup map.
- [x] Fix POTA lookup in QSO Entry
- [x] Fix SOTA lookup in QSO Entry

### DIFF
--- a/application/models/Sota.php
+++ b/application/models/Sota.php
@@ -23,6 +23,34 @@ class Sota extends CI_Model {
 
 		return $this->db->get($this->config->item('table_name'));
 	}
+
+	/*
+	 * The full SOTA reference directory with coordinates only — no QSO or
+	 * worked/confirmed status. Powers the optional clustered overlay on the
+	 * Gridsquare Lookup map, where it's used to look up where a reference sits.
+	 * Rows without coordinates are skipped because they can't be plotted.
+	 */
+	function get_directory() {
+		$sql = "SELECT reference, name, lat, lon, altitude FROM sota_directory
+		WHERE lat IS NOT null
+		AND lon IS NOT NULL
+		ORDER by reference";
+		
+		$query = $this->db->get('sota_directory');
+
+		$result = [];
+		foreach ($query->result() as $row) {
+			$result[] = [
+				'reference' => $row->reference,
+				'name'      => $row->name,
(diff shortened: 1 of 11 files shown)


application/models/Sota.php
```diff
@@ -23,6 +43,34 @@ function get_all() {
...
+	 * Gridsquare Lookup map, where it's used to look up where a reference sits.
+	 * Rows without coordinates are skipped because they can't be plotted.
+	 */
+	function get_directory() {
+		$sql = "SELECT reference, name, lat, lon, altitude FROM sota_directory
+		WHERE lat IS NOT null
+		AND lon IS NOT NULL
+		ORDER by reference";
+
+		$query = $this->db->get('sota_directory');
```
int2001: SQL above unused?!

application/models/Update_model.php
```diff
@@ -264,7 +329,82 @@ function hamqsl(){
...
+        }
+
+        if ($nCount > 0) {
+            return "DONE: " . number_format($nCount) . " POTA's saved";
+        } else {
+            return "FAILED: Empty file";
+        }
+    }
+
+    function pota2() {
```
int2001: dead code? can't spot caller

AndreasK79: Yep, forgot to remove before pr. Fixed now.

application/models/Sota.php
AndreasK79: No, look in gridlookup,php controller. It is used to fetch for the map.

int2001: Na, i meant: your query the table via `$query = $this->db->get('sota_directory');` instead of using the good SQL above.
This way the SQL is useless, and the query won't filter for `NOT null` (return everything)

AndreasK79: Yes, you are right, sorry. Fixed now.

int2001 (approved):
works good.
huge amount of data loaded for alle references. but i see no real way around it.

AndreasK79:
> works good. huge amount of data loaded for alle references. but i see no real way around it.

I know. Markercluster helps, but yes, a lot of data.

int2001 (approved):

Issue #2872, opened 2026-01-28
When adding a log,Satellite selection interface,The mobile terminal cannot display prompt words.


BH6AAP:
### I have checked the following

- [x] I am running the latest version of Wavelog
- [x] I have cleared my browser cache
- [x] This is a bug, not a setup question or feature request

### Bug Description

When I use the computer, I add the satellite selection in the log, and input fuzzy words to give a prompt.Satellite mode can also automatically give hints.But the mobile web page is not available now, even ifIt's not good for mobile phones to switch computer pages.This function,I remember it used to be normal and available.

tnx！73！
de bh6aap

### Steps to Reproduce

1.Add Log (Manual),2.click satellite 3.imput sat's
 name

### Expected Behavior

Prompts should be given under the satellite input box.

### Wavelog Version

2.2.2

### PHP Version

8.2.28

### MySQL/MariaDB Version

10.11.11-maraidb-0+deb12u1

### Server OS

debian 

### Browser

Android edge

### Device Type

Mobile

BH6AAP:
![Screenshot_2026-01-28-19-23-50-610_com.android.browser.jpg](https://github.com/user-attachments/assets/f53690a3-6fdf-4166-b498-ed71f60b7bdf)

BH6AAP:
Apple mobile browser can be used normally.

HB9HIL:
Hello @zhangyulu8023 

The mobile support for Wavelog is hard to realize and not our first priority (Mobile first - but not for Wavelog). There are some thirdparty mobile apps around which may solve your issues. Wavelog on Mobile is a special topic. 

We can try to fix the dropdown for sats on mobile but we can't promise.

BH6AAP:
ok rr ，thanks a lot，Pay tribute to the developers

PR #3578, opened 2026-08-03
Fix satellite name dropdown on mobile

MatykoBr:
Replaces the satellite name selection datalist on the logging page with a Bootstrap dropdown so it is also visible on mobile devices. (Requested in #2872)

### DIFF
--- a/application/views/qso/index.php
+++ b/application/views/qso/index.php
@@ -666,7 +666,7 @@ if (typeof window.DX_WATERFALL_FIELD_MAP === 'undefined') {
               <label for="sat_name"><?= __("Satellite Name"); ?></label>
 
               <input type="text" class="form-control" id="sat_name" name="sat_name" value="<?php echo $this->session->userdata('sat_name'); ?>" onblur="setTimeout(() => document.getElementById('satellite_names_list').innerHTML = '', 150)">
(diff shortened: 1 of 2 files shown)


int2001 (changes requested):
doesn't work.
throws instantly a JS-Error. Did you test it?

```
qso.js?v=1785819901:1246 Uncaught TypeError: Cannot read properties of undefined (reading 'options')
    at HTMLInputElement.<anonymous> (qso.js?v=1785819901:1246:49)
    ...
```

Edit: Seems to work on mobile (but even there's a JS-Error) but breaks several things:
- Shows always "Satellite nor found" (even if correct) in SAT-Tab
- Doesn't trigger the Az/Ele calculation
- Doesn't fill the Prop-mode to SAT
- ...

MatykoBr:
I have now rewritten the code and tested it extensively. The JS error is fixed and the things you mentioned work again. Sorry for the rushed PR.

int2001 (approved):
much better.
i tested:
- recalling FAV --> Works
- setting FAV --> Works
- Filling in SAT --> Works
- Az/Ele ticker --> Works

tbh: critical place in code. i'd like to see 2nd review from @phl0 here

phl0 (approved):
Tested and works for me (desktop and mobile). Going to merge here.

PR #3580, opened 2026-08-04
[Gridsquare Lookup] Use dropdown for refs

AndreasK79:
Small one this time:
image
image
image


int2001 (approved):
works. even on mobile.

PR #3595, opened 2026-08-06
[Activation Planner] Renamed from Gridsquare Lookup

AndreasK79:
After heavy convincing, @int2001 talked me into renaming the Gridsquare Lookup page.
It has grown into much more than just a gridsquare tool.


int2001 (approved):
create useful things and talk about/advertise them.
it's a full planner

PR #3596, opened 2026-08-07
[Activation Planner] share gridsquare, create location, show refs

AndreasK79:
Add sharing and local refs to gridsquare popup
- [x] Share to X
- ~~[ ] Share to hams.at~~
- [x] Create location
- [x] Show refs within 6 char gridsquare

image

image

Ref loading is slow. Location creating just brings gridsquare. Not tested X or hams.at.

This is highly a work in progress.


phl0 (approved):
This is fine with me. I shall setup a hams.at test server to check if and how sharing to hams.at makes sense. Lets just leave it out here (for now)?

phl0 (approved):

PR #3602, opened 2026-08-08
Dxcc award propagation filter

AndreasK79:
This PR adds a propagation dropdown to the filtering in the DXCC award.


phl0 (approved):
Works :)

PR #3603, opened 2026-08-08
[Activation Planner] Disable tooltip for WWFF, POTA and SOTA

AndreasK79:
- [x] Disable tooltip for WWFF, POTA and SOTA, as information is shown on click
- [x] Shows labels for overlays, doesn't catch click, as we want to focus on catching gridsquare click
- [x] Change button for refs and overlays if anything is active (same behavior as LBA)


phl0 (approved):
works

PR #3604, opened 2026-08-08
OnAir widget

HB9HIL:
Did some maintenance on the on-air widget by @filipmelik 

- Added translations show visitors see the widget in their native language (if translations are available, otherwise english)
- matched with code conventions (<?= only at translations, otherwise <?php echo)
- Added an user option to also display the radio name as requested in https://github.com/wavelog/wavelog/discussions/3550 (opt in)
- Widget shows now station callsign instead user callsign

### DIFF
--- a/application/controllers/Header_auth.php
+++ b/application/controllers/Header_auth.php
@@ -334,6 +334,7 @@ class Header_auth extends CI_Controller {
             $mapped['on_air_widget_enabled']                  ?? '',
             $mapped['on_air_widget_display_last_seen']        ?? '',
             $mapped['on_air_widget_show_only_most_recent_radio'] ?? '',
+            $mapped['on_air_widget_display_radio_name']       ?? '',
             $mapped['qso_widget_display_qso_time']            ?? '',
             $mapped['dashboard_banner']                       ?? '',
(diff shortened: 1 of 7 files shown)


phl0 (approved):
Works. I will think about a possible solution to pull the active callsign maybe from last log (not necessarily the action station location). But for now its ok.

PE1PQX:
> Works. I will think about a possible solution to pull the active callsign maybe from last log (not necessarily the action station location). But for now its ok.

After entering the V2 API URL and V2 API code, I get an 'Failed: Unknown resouce path'

WavelogGate 2.1.0 and WaveLog 3.1.0. Or will it be in the next update of Wavelog?

HB9HIL:
grafik



this has nothing to do with API v2!

HB9HIL:
@PE1PQX and it shows the call of the current **active** station location

PS: it's already in 3.1.0

PE1PQX:
Rig works now, thanks... But after entering the V2 API URL and V2 API code, I get an 'Failed: Unknown resouce path'

Change callsign (PE1PQX -> OE/PE1PQX for example) does not

HB9HIL:
mail me hb9hil@wavelog.org

PR #3611, opened 2026-08-09
[Activation Planner] Added IOTA refs

AndreasK79:


phl0 (approved):
Works fine for me. Just added a small style to make the hovering cursor look like its a (clickable) link.

PR #3612, opened 2026-08-09
[Activation Planner] Nearby ref list

AndreasK79:
This PR adds the following:

- [x] New button to show a dialog with a list of references within 20km from the gridsquare.
- [x] Clicking on map fills grid
- [x] Nearby refs should only load active refs
- [x] Same icons for markers when showing refs nearby while clicking on the map, as if they were activated from refs menu (same with popup menu)
- [x] Expanded limit to 20km for markers shown nearby clicked gridsquare


int2001:
waits for #3625

int2001:
open:
- [x] Clicking on map doesn't fill grid - so "nearby refs" show "invalid gridsquare"
- [x] Nearby refs should take care of active/inactive refs

int2001 (approved):
works. planner is getting better and better.

reminder for future: maybe add index to lat/lon

PR #3614, opened 2026-08-09
[Docker] Add missing installer prerequisites

dg0tm:
Add the PHP curl, mbstring, and xml extensions required by the
installer prechecks, and install curl for the container cron job.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,13 +8,16 @@ COPY --from=ghcr.io/mlocati/php-extension-installer /usr/bin/install-php-extensi
 
 # system packages
 RUN apt-get update \
-    && apt-get install -y --no-install-recommends cron \
+    && apt-get install -y --no-install-recommends cron curl \
     && apt-get clean \
     && rm -rf /var/lib/apt/lists/*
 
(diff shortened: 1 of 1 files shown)


HB9HIL:
current dev image has everything it needs

grafik


your mentioned extensions are already included in the default apache image.. 

why you think they are missing?

dg0tm:
Thanks for checking! My concern is about following Docker best practices
regarding explicit dependency management.

Even if the current default Apache image includes these extensions out of
the box, relying on that implicit behavior is a risk. Upstream base images
can change their defaults or strip out extensions in future releases
without warning.

It is a standard best practice to explicitly declare and install all required
extensions in our own Dockerfile. This ensures:

 * Deterministic builds: The image builds reliably, regardless of upstream changes.
 
 * Code as documentation: It makes it immediately clear to everyone which PHP extensions Wavelog strictly depends on.
 
It's basically a safeguard against unexpected breaking changes down the road.

HB9HIL (approved):

HB9HIL:
Ok makes sense

PR #3615, opened 2026-08-09
Dashboard DXCC Panel

dg0tm:
Count DXCC confirmed totals from paper QSL and LoTW only.
Derive needed from current DXCC entities minus confirmed.
Keep eQSL separate in dashboard and visitor breakdowns.
Document the updated DXCC semantics in API v2.


dg0tm:
DXCCPanel

phl0 (approved):

PR #3618, opened 2026-08-09
[Logbook] Fix DCL status handling

dg0tm:
Compare DCL sent and received against their status columns.
Keep the DCL save path from mixing date and status fields.

### DIFF
--- a/application/models/Logbook_model.php
+++ b/application/models/Logbook_model.php
@@ -1669,7 +1669,7 @@ class Logbook_model extends CI_Model {
 			$clublogrdate = $qso->COL_CLUBLOG_QSO_DOWNLOAD_DATE;
 		}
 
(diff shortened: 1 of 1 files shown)


application/models/Logbook_model.php
```diff
@@ -1671,15 +1671,15 @@ function edit() {
 
 		if ($dcl_sent == 'N' && $qso->COL_CLUBLOG_QSO_UPLOAD_STATUS != $dcl_sent) {
```
int2001: also wrong here...
```suggestion

		if ($dcl_sent == 'N' && $qso->COL_DCL_QSL_SENT != $dcl_sent) {
```

int2001 (changes requested):
good one.
found one more :)

int2001 (approved):

Issue #3621, opened 2026-08-10
Station Location State does not populate on edit


kc9uhi:
### I have checked the following

- [x] I am running the latest version of Wavelog
- [x] I have cleared my browser cache
- [x] This is a bug, not a setup question or feature request

### Bug Description

When editing a Station Location with a set Station State, it does not populate in the editor.

javascript set_state variable is set with stored station location state in line [127](https://github.com/wavelog/wavelog/blob/af3256140bd05403b7c4a421746c2ea653a4f04f/application/views/station_profile/edit.php#L127) of application/views/station_profile/edit.php, for later assignment of 'selected' in the dropdown; however line [40](https://github.com/wavelog/wavelog/blob/af3256140bd05403b7c4a421746c2ea653a4f04f/assets/js/sections/station_locations.js#L40) of assets/js/sections/station_locations.js grabs the current state of #stateDropdown (at this point, not set) and ignores set_state before calling updateStateDropdown(). This results in the stored state not getting filled in the dropdown.

### Steps to Reproduce

Create a station location with a Station State.
Edit the station location. 'Station State' dropdown will be blank.

### Expected Behavior

Station State dropdown is populated with stored state.

### Wavelog Version

3.1.0

### PHP Version

8.4.24

### MySQL/MariaDB Version

11.8.6-MariaDB

### Server OS

Debian

### Browser

Chrome

### Device Type

Desktop

int2001:
Working on it at #3622

PR #3622, opened 2026-08-10
Fix State dropdown/dep in station edit

int2001:
Fixes a bug mentioned in #3621:

State dropdown isn't populated when editing stationlocation (with a saved State)


phl0 (approved):
Works

PR #3625, opened 2026-08-10
Add POTA-Park-Boundaries (onClick)

int2001:
Add POTA Parkboundaries to WL / Activationplanner.

Test:
1. Run POTA-Boundaries updatejob (from debug or cron) once.
2. At Activationplanner: Enable Ref -> POTA.
3. Click on a park (or multiple ones) - boundaries should be visible.
4. Clear parks with clear-button.

Bonus: 
- Last-Activation (must be in any station-location of active logbook)
- Take care of inactive/outdated Refs (Need one cron-run) // see #3630 

~~This PR is intentionally drafted, because we need to clarify about licence for geoJSON-Files.~~
Use/Permission for geoJSON source (https://pota-map.info/geojson/) is granted by DK5UR
Works ONLY for parks located in `['DE','AT','CH','CZ','DK','LU','LI']`

image


phl0 (approved):
Tested and works fine here now. SOTA refs also loading fine.

AndreasK79 (approved):

PR #3626, opened 2026-08-10
Refactor DXCC stats on dashboards

phl0:
This PR implements minor changes in the dashbaords:

- Removed eQSL from countings (as this is not valid for the official DXCC award anyway)
- Aligned the layout of the visitor dashboard with the logged-in user dashboard (removed empty link on "Confirmed" and add tooltips to numbers)

### DIFF
--- a/application/models/Logbook_model.php
+++ b/application/models/Logbook_model.php
@@ -4284,7 +4284,7 @@ class Logbook_model extends CI_Model {
 				$row = $query->row();
 				return [
 					// Country stats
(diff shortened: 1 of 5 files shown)


application/models/Logbook_model.php
int2001: needed?

int2001 (approved):
finally :)

Nit: L4327 LB_Model: still returns 0 (in case of no data) for eQSL-Countries

application/models/Logbook_model.php
phl0: No

PR #3627, opened 2026-08-10
Widget showing last LoTW uploads

phl0:
This adds a widget that shows last LoTW uploads for a given slug. 2nd parameter controls whether results are sorted by date desc or call asc.
Example URL: 

https://<WAVELOG_URL>/index.php/widgets/lotw_upload/<SL_UG>?theme=default

which defaults to sorting by date desc or:

https://<WAVELOG_URL>/index.php/widgets/lotw_upload/<SL_UG>?theme=default&sort=call

which sorts the result by callsign ascending.


application/controllers/User.php
```diff
@@ -333,9 +333,10 @@ function add() {
 				$this->input->post('oqrs_grouped_search_show_station_name') ?? 'off',
 				$this->input->post('oqrs_auto_matching') ?? 'on',
 				$this->input->post('oqrs_direct_auto_matching') ?? 'on',
-        			$this->input->post('user_dxwaterfall_enable') ?? 'N',
+				$this->input->post('user_dxwaterfall_enable') ?? 'N',
 				$this->input->post('user_qso_show_map') ?? 1,
-				$this->input->post('clubstation') == '1' ? true : false)
+				$this->input->post('clubstation') == '1' ? true : false,
+				$this->input->post('last_lotw_upload_widget_enabled'))
```
int2001, line 339: doesn't align with `add` in User_model

application/models/User_model.php
```diff
@@ -343,6 +343,7 @@ function add($username, $password, $email, $type, $firstname, $lastname, $callsi
 				['widget',     'qso',                     'display_qso_time',              $qso_widget_display_qso_time                ?? 'false'],
 				['qso_db_search_priority', 'enable',      'boolean',                       $user_qso_db_search_priority ?? 'Y'],
 				['dxwaterfall', 'enable',                 'boolean',                       $user_dxwaterfall_enable     ?? 'N'],
+				['widget',     'lotw_upload_upload',      'enabled',                       $last_lotw_upload_widget_enabled            ?? 'false'],
```
int2001: `lotw_upload_upload` vs. `last_lotw_upload` in L435 ...

application/controllers/Widgets.php
```diff
@@ -235,6 +235,76 @@ public function on_air($user_slug = "") {
...
+			$this->load->view('widgets/lotw_upload', $data);
+			return;
+		}
+
+		try {
+			$user = $this->get_user_by_slug($user_slug);
+		} catch (\Exception $e) {
+			$data['text_size_class'] = $this->prepare_text_size_css_class($text_size);
+			$data['error'] = __("User slug not specified");
+			$data['error'] = $e->getMessage();
```
int2001: hmm. `$data['error']` gets overwritten at L273 - so L272 is dead?

int2001 (changes requested):

application/controllers/Widgets.php
phl0: Think that was there for debug puposes (copied from on_air widget). Removed both occurences.

application/models/User_model.php
phl0: Typo. Fixed

application/controllers/User.php
phl0: Fixed.

int2001 (approved):

PR #3633, opened 2026-08-10
Fix US Counties award undercounting confirmed QSOs (LoTW/QRZ.com/Clublog ignored)

Reid-n0rc:
## What changed and why

Fixes #3632.

**Branch tested:** [`Reid-n0rc:counties-lotw-qrz-clublog-confirmation-fix`](https://github.com/Reid-n0rc/wavelog/tree/counties-lotw-qrz-clublog-confirmation-fix) — this is exactly what was run through the Docker test described below, no changes since.

The "QSOs Confirmed" / "Confirmed Counties" numbers on **Awards > USA > US Counties** only counted a QSO as confirmed if it had `COL_QSL_RCVD='Y'` (paper QSL) or `COL_EQSL_QSL_RCVD='Y'` (eQSL). Confirmations received exclusively via LoTW, QRZ.com, or Clublog were ignored, so counties confirmed only through those (very common) methods showed up as worked but not confirmed.

This adds the three missing confirmation sources to the existing `(col_qsl_rcvd='Y' or col_eqsl_qsl_rcvd='Y')` predicate in all three places `application/models/Counties.php` checks it (`get_counties_summary()`, `get_counties()`, `get_county_counts()`):

- `col_lotw_qsl_rcvd`
- `COL_QRZCOM_QSO_DOWNLOAD_STATUS`
- `COL_CLUBLOG_QSO_DOWNLOAD_STATUS`

This brings the US Counties award in line with how every other award in the codebase (DXCC, WPX, WAC, ITU, POTA, SOTA, WWFF, WAE, …) already determines "confirmed" — those all check LoTW, QRZ.com, and Clublog alongside paper QSL/eQSL.

Single-purpose fix, no schema changes, no behavior change for anything outside this award.

## Testing

Ran the branch end-to-end against a real instance (not just reading the query):

1. Built a throwaway Docker stack from this repo's own `Dockerfile` + a MySQL 8 container, ran the standard web installer against it (`install/assets/install.sql`), and created a test user/station.
2. Seeded 5 QSOs in California, one per county, each with a **different single confirmation source** and nothing else set:

   | County | Confirmed via | Worked | Confirmed (before fix) | Confirmed (after fix) |
   |---|---|---|---|---|
   | Los Angeles | LoTW only | 1 | 0 | **1** |
   | Orange | QRZ.com only | 1 | 0 | **1** |
   | San Diego | Clublog only | 1 | 0 | **1** |
   | Ventura | Paper QSL only | 1 | 1 | 1 |
[...]

Issue #3632, opened 2026-08-10
US Counties (USA-CA) award undercounts confirmed QSOs — LoTW/QRZ.com/Clublog confirmations ignored


Reid-n0rc:
### I have checked the following

- [x] I am running the latest version of Wavelog
- [x] I have cleared my browser cache
- [x] This is a bug, not a setup question or feature request

### Bug Description

On the **Awards > USA > US Counties** page, the "QSOs Confirmed" / "Confirmed Counties" numbers do not reflect all confirmed contacts. A county that has a QSO confirmed in the log — sometimes via multiple confirmation methods — is frequently shown as worked but **not confirmed** (0 confirmed), even though the log clearly shows a valid confirmation.

Root cause: `Counties::get_counties_summary()`, `Counties::get_counties()`, and `Counties::get_county_counts()` in `application/models/Counties.php` only treat a QSO as "confirmed" if `COL_QSL_RCVD = 'Y'` (paper QSL) or `COL_EQSL_QSL_RCVD = 'Y'` (eQSL):

```php
and (col_qsl_rcvd='Y' or col_eqsl_qsl_rcvd='Y')
```

This ignores three other confirmation sources that Wavelog tracks and that every other award in the app (DXCC, WPX, WAC, ITU, POTA, SOTA, WWFF, WAE, …) already includes:

- `COL_LOTW_QSL_RCVD` (ARRL Logbook of The World) — the most common confirmation method for US counties
- `COL_QRZCOM_QSO_DOWNLOAD_STATUS` (QRZ.com Logbook)
- `COL_CLUBLOG_QSO_DOWNLOAD_STATUS` (Clublog)

Any county confirmed solely via LoTW, QRZ.com, or Clublog is therefore silently dropped from the "Confirmed" tally, which is why some users see partial undercounts and others (whose confirmations are exclusively via LoTW/QRZ/Clublog) see 0 confirmed counties across the board despite having genuinely confirmed contacts.

### Steps to Reproduce

1. Log a QSO with a US county (ADIF `CNTY`, e.g. `CA,Orange`) filled in.
2. Confirm that QSO via LoTW only (or QRZ.com Logbook only, or Clublog only) — do **not** mark it confirmed via paper QSL or eQSL.
3. Go to **Awards > USA > US Counties**.
4. Open the state the county belongs to (click the state code, or the "Worked"/"Confirmed" count).

### Expected Behavior

[...]

PR #3635, opened 2026-08-10
Add QSL confirmation-source filter to US Counties award page

Reid-n0rc:
## What changed and why

Implements #3634.

**Branch tested:** [`Reid-n0rc:counties-award-qsl-source-filter-v2`](https://github.com/Reid-n0rc/wavelog/tree/counties-award-qsl-source-filter-v2) — this is exactly what was run through the Docker test described below, no changes since.

**Depends on #3633** (the LoTW/QRZ.com/Clublog confirmed-count fix). This branch is built on top of that fix, so the diff currently shows both commits until #3633 merges — only the second commit ("Add QSL source filter...") is new here.

Once #3632 is fixed, "Confirmed" on **Awards > USA > US Counties** always means "confirmed via any source," with no way to narrow it down. Every other award page with a confirmation concept (DXCC, WPX, WAC, ITU, POTA, SOTA, WWFF, WAE, …) already has a "Show QSO with QSL Type" filter for this. This adds the same pattern to the US Counties page.

### Model (`application/models/Counties.php`)

- Added `$qsl_source_columns`, mapping the 5 selectable sources (`qsl`, `lotw`, `eqsl`, `qrz`, `clublog`) to their DB columns.
- Added `build_confirmed_condition($qsl_sources)`: builds the SQL `OR` condition from a caller-supplied list. `null` (not passed) means "default to all sources," so existing callers are unaffected. An **explicitly empty array** ("user unchecked every source") correctly returns `1=0` rather than silently falling back to all — this distinction matters and is covered below under Testing.
- `get_counties_summary()`, `get_counties()`, and `get_county_counts()` now accept an optional `$qsl_sources` param and use the shared helper instead of a hardcoded predicate.
- Every value that reaches the generated SQL is looked up through `isset($this->qsl_source_columns[$source])` first — user input never reaches the query string directly, only the fixed whitelisted column name does.

### Controller (`application/controllers/Awards.php`)

- `counties_qsl_sources_from_post()`: reads which checkboxes were submitted. A hidden `qslFilterSet` marker distinguishes "the filter form was actually submitted with these boxes checked/unchecked" from "no filter was submitted at all, use the default" — needed because unchecked checkboxes simply don't appear in the POST body, so presence-of-marker is the only reliable signal.
- `counties()`, `counties_list_ajax()`, and `counties_state_ajax()` all resolve and pass through `$qsl_sources`, so the main table and both drill-down views stay consistent.

### Views

- `application/views/awards/counties/index.php`: adds a "Filters" dropdown matching the DXCC page's pattern (same `dropdown-filters-responsive` CSS, same checkbox layout), defaulting to all 5 sources checked.
- `application/views/interface_assets/footer.php`: `countiesQslFilterData()` reads the current checkbox state and is merged into the existing `displayStateCounties()` / `displayStateCountiesList()` AJAX calls, so the state/county drill-downs match whatever was last submitted on the main page.

"Worked" and "Remaining" are intentionally unaffected — "Remaining" is explicitly `target - worked` (labeled "Based on current worked count" in the UI), and worked status doesn't depend on how/whether a QSO was confirmed.

## Testing

Tested end-to-end against a real running instance (Docker + MySQL 8, standard web installer), same methodology as #3633.

Seeded 6 QSOs in California, one confirmed via each of the 5 sources plus one unconfirmed:

| County | Confirmed via |
|---|---|
| Los Angeles | Paper QSL only |
| Orange | LoTW only |
| San Diego | eQSL only |
| Ventura | QRZ.com only |
| Kern | Clublog only |
| Fresno | none |

Submitted the filter form with different combinations and read back the rendered `awards/counties` page and the `counties_state_ajax` / `counties_list_ajax` drill-downs:

| Filter selection | Expected CA Confirmed | Actual |
|---|---|---|
| Default (fresh load, no filter submitted) | 5 | 5 ✅ |
| Only LoTW checked | 1 (Orange) | 1 ✅ |
| QSL + eQSL checked | 2 (Los Angeles, San Diego) | 2 ✅ |
| All 5 explicitly checked | 5 | 5 ✅ |
| **No sources checked** | **0** | 0 ✅ |

The last row caught a real bug during testing: the first version of `build_confirmed_condition()` used PHP's `empty()` to decide "no sources given → default to all," which also matched "user explicitly unchecked every box," so submitting the filter with nothing checked incorrectly showed Confirmed = 5 instead of 0. Fixed by switching the default check to `=== null` so a real (even empty) array from the controller is never re-defaulted.

Also verified the per-state and per-county drill-downs honor the filter — e.g. with only QRZ.com checked, `counties_list_ajax` (type=confirmed) correctly returns only `CA,Ventura`, and `counties_state_ajax` shows every other CA county at 0 confirmed with Ventura at 1.

## Screenshots / output

Per-county breakdown from `counties_state_ajax` with only LoTW checked:

```
#  County          QSOs Worked  QSOs Confirmed
1  CA,Fresno       1            0
2  CA,Kern         1            0
3  CA,Los Angeles  1            0
4  CA,Orange       1            1
5  CA,San Diego    1            0
6  CA,Ventura      1            0
   Total           6            1
```

### DIFF
--- a/application/controllers/Awards.php
+++ b/application/controllers/Awards.php
@@ -1357,7 +1357,9 @@ class Awards extends CI_Controller {
 
     public function counties()	{
         $this->load->model('counties');
-        $data['counties_progress'] = $this->counties->get_counties_progress();
+        $qsl_sources = $this->counties_qsl_sources_from_post();
+        $data['counties_progress'] = $this->counties->get_counties_progress($qsl_sources);
+        $data['qsl_sources'] = $qsl_sources;
 		$data['user_map_custom'] = $this->optionslib->get_map_custom();
 
         // Render Page
@@ -1367,11 +1369,37 @@ class Awards extends CI_Controller {
         $this->load->view('interface_assets/footer');
     }
 
+    /*
+     * Reads which QSL confirmation sources (qsl/lotw/eqsl/qrz/clublog) are
+     * selected in the counties award filter. The filter form (and the state
+     * / list AJAX calls it drives) always posts a "qslFilterSet" marker
+     * alongside the checkboxes, so an unchecked box can be told apart from
+     * a request that never included the filter (which defaults to all
+     * sources, e.g. a fresh page load).
+     */
+    private function counties_qsl_sources_from_post() {
+        $available = array('qsl', 'lotw', 'eqsl', 'qrz', 'clublog');
+
+        if (!$this->input->post('qslFilterSet')) {
+            return $available;
+        }
+
+        $qsl_sources = array();
+        foreach ($available as $source) {
+            if ($this->input->post($source)) {
+                $qsl_sources[] = $source;
+            }
+        }
+
+        return $qsl_sources;
+    }
+
     public function counties_list_ajax() {
         $this->load->model('counties');
         $state = str_replace('"', "", $this->security->xss_clean($this->input->post("State")));
         $type  = str_replace('"', "", $this->security->xss_clean($this->input->post("Type")));
-        $data['counties_array'] = $this->counties->counties_details($state, $type);
+        $qsl_sources = $this->counties_qsl_sources_from_post();
+        $data['counties_array'] = $this->counties->counties_details($state, $type, $qsl_sources);
         $data['type'] = $type;
         $this->load->view('awards/counties/details_ajax', $data);
     }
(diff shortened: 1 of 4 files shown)


int2001:
See https://github.com/wavelog/wavelog/pull/3637#issuecomment-5249252971 and close this one please

Reid-n0rc:
Will be fixed with PR #3637

PR #3636, opened 2026-08-10
[Logbook] Tighten Club Log confirmation handling

dg0tm:
This is a low-impact hardening pass around the existing Club Log
confirmation flow. It does not add new feature surface or change the
DXCC breakdown.

The main goal is to make the existing confirmation handlers use the
right data flags consistently:

  - `dxcc_confirmed`
  - `dxcc_confirmed_on_band`
  - `dxcc_confirmed_on_band_mode`
  - `confirmed` in the logbook / lookup / QSO UI paths

It also includes a small UI consistency fix so the Club Log column is
only shown when the user has Club Log configured.

Changes:
  - normalize confirmation input before building confirmation queries
  - guard missing QSO fields in confirmation checks
  - include Club Log in the shared confirmation filters
  - gate the Club Log column by user setup

This keeps Club Log aligned with the other confirmation sources and
fixes the existing handlers rather than adding new behavior.

### DIFF
--- a/application/controllers/Logbook.php
+++ b/application/controllers/Logbook.php
@@ -430,6 +430,12 @@ class Logbook extends CI_Controller {
 			}
 			$extrawhere.=" COL_QRZCOM_QSO_DOWNLOAD_STATUS='Y'";
 		}
+		if (isset($user_default_confirmation) && strpos($user_default_confirmation, 'C') !== false) {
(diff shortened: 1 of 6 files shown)


int2001 (approved):
LGTM

phl0 (approved):
Tested and works for me.

PR #3637, opened 2026-08-10
Add worked/confirmed county map to the US Counties award page

Reid-n0rc:
## Summary

Adds a "Show Counties Map" feature to Awards > USA > US Counties, showing worked and confirmed US counties on an interactive map - the same idea as the existing WAS award's "Show WAS Map", but at county granularity (~3,143 counties instead of 50 states). Also adds Mode (Phone/CW/Digital) and Band filters alongside the existing QSL-source filter, since the USA-CA award has separate endorsements for those.

## What's included

- **County boundary data** (`assets/json/geojson/counties_291/6/110.geojson`): built from the US Census Bureau's 2022 Cartographic Boundary Files at 1:500,000 scale (public domain), matched to the ARRL/MARAC county names already used by `assets/json/US_counties.csv` / `COL_CNTY`. All 3,143 mappable counties matched (one, Virginia's "Bedford City", has had no boundary of its own since a 2013 city/county merger, but is still counted toward the award target). Full sourcing/attribution and the name-matching notes are in `assets/json/geojson/COUNTIES_SOURCE.md`.
- **Backend**: `Counties::get_counties_map()` + `Awards::counties_map()`, mirroring the WAS award's `was_map()` AJAX endpoint, respecting the same QSL-source/Mode/Band filters as the rest of the page.
- **Frontend**: a Table/Map tab pair on the counties award page; `countiesmap.js` (adapted from `wasmap.js`) renders the Leaflet map with confirmed/worked/unworked coloring, a legend, hover info, and bolder state outlines drawn on top of the (thinner) county lines. The state outline is dissolved from the *same* county polygons used for shading (see below), not a separately-sourced file, so it stays pixel-aligned with the shading at any zoom level. Clicking a county reuses the existing QSO drill-down dialog.
- **GeoJSON caching**: all six boundary files (`counties_291/6/110.geojson` + `counties_state_outline_291/6/110.geojson`, ~30MB combined) are served through a new `Awards::counties_geojson()` endpoint instead of being fetched as static assets. It sets explicit `Cache-Control`/`ETag` headers and answers conditional requests with `304 Not Modified`, so the payload is only re-downloaded once per browser regardless of the host's static-file-serving defaults. Implementation mirrors `Activationplanner.php`'s directory endpoints (same ETag/304 pattern), with two deliberate differences: the ETag is computed from `filemtime()`+`filesize()` (a stat, no full read) rather than hashing the payload, since these come from disk instead of already being in memory from a DB query; and `max-age` is 30 days rather than Activationplanner's 1 hour, since this data only changes on a Wavelog upgrade, not per request. Thanks to @int2001 for flagging this and pointing at the right reference.
- **QSL-source filter**: QSL/LoTW/eQSL/QRZ/Clublog checkboxes, defaulting to QSL + LoTW only on a fresh page load - the two sources that actually count for the MARAC award (per int2001's point 3). The other three remain available but opt-in.
- **Mode filter**: Phone/CW/Digital checkboxes, all checked by default. Phone follows [ADIF 3.1.7](https://www.adif.org/317/ADIF_317.htm): `SSB`/`AM`/`FM` plus `DIGITALVOICE` and all of its submodes (`C4FM`/`DMR`/`DSTAR`/`FREEDV`/`M17` - the same grouping `assets/js/radiohelpers.js`'s `MODE_LISTS.DIGITAL_VOICE` already uses elsewhere in the app). CW is its own group. Digital is defined by *exclusion* (not Phone, not CW) rather than an enumerated list, so any current or future ADIF data mode is counted automatically without needing a code change. Checks both `COL_MODE` and `COL_SUBMODE`, since logging software varies in which one it populates for a given mode.
- **Band filter**: 14 bands, 160m through 70cm (`160m 80m 60m 40m 30m 20m 17m 15m 12m 10m 6m 2m 1.25m 70cm` - 4m intentionally excluded), all checked by default.
- Two small unrelated fixes to the counties award carried along from an earlier branch: undercounting of confirmed QSOs, and the QSL-source filter's underlying plumbing.

### Bug: state outline didn't align with county shading at high zoom

The bold state-outline overlay originally reused `states_291/6/110.geojson` - the WAS map's own, *separately* generalized/simplified state boundaries. Even though both files represent the same physical state edge, they're two independent Census-data simplification passes with different vertex positions, so at street-level zoom the bold outline visibly diverged from the county shading's own edge underneath (most noticeable along coastlines and other complex borders).

Fixed by generating `counties_state_outline_291/6/110.geojson`: the *same* county polygons already used for shading, dissolved per state (`shapely.ops.unary_union`, grouped by DXCC + state). This guarantees the outline and the shading share identical coordinates at every zoom level - it's the same data, not a separate approximation of it. Regeneration steps are documented in `COUNTIES_SOURCE.md`. Tradeoff: these files are necessarily heavier than the ones they replace (`counties_state_outline_291.geojson` is ~3.8MB vs the old `states_291.geojson`'s ~60KB) since they carry the same vertex density as the county data along every state edge - judged worth it for correctness, and it's cached the same 30-day way as the other boundary files.

### Bugs found and fixed while testing the Mode filter

Exhaustively testing the Mode filter against every distinct ADIF Mode/Submode value (see Test Methodology below) surfaced two real SQL bugs, both the same root cause on different columns:

`UPPER(COL_SUBMODE) in (...)` (and separately `UPPER(COL_MODE) in (...)`) evaluates to SQL `NULL`, not `false`, when the column itself is `NULL`. Since the Digital condition is built as `not phone_condition and not cw_condition`, a `NULL` on either side poisoned the whole expression to `NULL` (i.e. excluded from the `WHERE` clause) instead of `true` - meaning **every QSO logged without both a Mode and a Submode populated would have silently dropped out of the Digital bucket**, which is most real-world logging (most modes aren't logged with a submode at all). Fixed by wrapping both columns in `COALESCE(col, '')` before comparing.

## Screenshots

**Table view**
Table view

**Map overview**
Map overview

**Zoomed detail** (Los Angeles area - shows boundary accuracy and the bolder state outline)
Zoomed detail

**Click-through to QSO data**
Click dialog

**Street-level zoom - outline now pixel-aligned with shading** (Santa Barbara coastline, zoom 15 - the bold state outline traces exactly along the shading's edge, no gap)
Zoom outline alignment

## Test methodology

No existing project skill/fixture for running Wavelog locally, so a real environment was built for this PR:

- **Environment**: `docker build` from the repo's own `Dockerfile` + a `mysql:8.0` container, linked via `DB_HOST`. The web installer (a single-page JS wizard posting to `run.php`/`ajax.php`, not a plain multi-page form) was driven end-to-end with **Playwright** (chromium), since no chromium-cli was available in this environment.
- **Functional/visual testing** (map rendering, hover, click-through to the QSO dialog, legend counts, filter dropdown UI, screenshots): Playwright driving a real logged-in browser session against the containerized app.
- **Filter-logic correctness testing** (the QSL/Mode/Band filter combinations): a lighter-weight **Python (`requests`) harness** posting directly to the AJAX endpoints and asserting on the returned county sets - much faster to iterate than browser automation for pure logic sweeps over dozens/hundreds of cases, with no need to render anything.
- **Test data**: seeded directly via SQL into `TABLE_HRD_CONTACTS_V01`, using real county names from `assets/json/US_counties.csv` (including tricky-naming cases: VA independent cities, MO St. Louis city/county, CT's 9 planning regions, AK's post-2015 borough renames/splits) so the map/name-matching path was exercised too, not just the filter logic.
- **Combinatorial coverage for the new filters**, built as explicit matrices rather than a handful of spot checks:
  - **Band x Mode**: 14 bands x 3 mode groups = 42 cells, one dedicated county per cell (Kansas), each holding a single unconfirmed QSO. Lets every individual band and every individual mode be isolated and verified independently, plus every one of the 42 band+mode pair intersections.
  - **Every ADIF mode**: rather than testing a handful of representative modes (SSB/CW/FT8), pulled the **full, live set of Mode/Submode tokens from the app's own `adif_modes` DB table** (`SELECT mode ... UNION SELECT submode ...`, active and legacy/import-only rows both) - 245 distinct tokens - and seeded one QSO per token (Texas, 245 of 254 counties used) to confirm every single one lands in the correct Phone/CW/Digital bucket. This is what caught the two NULL-propagation bugs above; a smaller representative-mode test would have missed both, since `SSB`/`CW`/`FT8` all happen to populate `COL_MODE` without a submode - the failure only shows up for `NULL` on either column in various combinations.
  - **QSL sources**: one dedicated county per individual source (qsl/lotw/eqsl/qrz/clublog) plus one county confirmed via two sources simultaneously, to verify the OR-merge across sources doesn't double-count. (Not a full 2^5=32-combination sweep - the confirmation condition is a flat OR with no branching/NULL-coalescing logic like the mode filter had, so the risk profile is much lower and this was judged sufficient.)
  - **Edge cases**: selecting zero bands or zero modes correctly returns no counties rather than silently falling back to "all".
- **Caching verification**: `curl` with an explicit cookie jar and `If-None-Match` header, since browser `fetch()` transparently serves cached-and-revalidated (304) responses as status `200` to JS and won't show the real wire-level status code.

## Test results

- [x] Fresh install (Docker + MySQL) with seeded test QSOs across CA, TX, NY, FL, IL, LA, VA, MD, MO, CT, AK, HI, WA, CO, KS
- [x] Table totals match map legend counts
- [x] QSL-source filter toggles recolor the map correctly
- [x] Clicking a county opens the correct QSO list
- [x] Spot-checked Alaska (Chugach/Copper River split) and Connecticut (9 planning regions) render correctly
- [x] No browser console errors
- [x] `counties_geojson()`: `200` with `ETag`/`Cache-Control: public, max-age=2592000` on first load; `304 Not Modified` (empty body) on a repeat request with a matching `If-None-Match`; mismatched `If-None-Match` correctly falls back to `200` with the full payload; invalid/path-traversal scope values return `404`; unauthenticated requests are blocked (inherits the existing `Awards` controller auth check)
- [x] Band x Mode: all 14 bands and all 3 modes verified in isolation, plus all 42 band+mode pair intersections - unchecking any single band or mode drops exactly the counties whose sole QSO was in that cell, on both table and map
- [x] Mode filter accuracy against all 245 distinct ADIF Mode/Submode tokens from the app's own `adif_modes` table - every one lands in the correct bucket after the NULL-handling fixes above
- [x] QSL-source filter: each of the 5 sources independently drives confirmation; the two-source OR-merge case confirms without double-counting
- [x] Edge cases: zero bands / zero modes selected correctly show no counties
- [x] State outline/shading alignment: re-verified visually at street-level zoom (z15) along the Santa Barbara coastline after switching the outline source to the dissolved county data - no visible gap between the bold outline and the shading's edge; full filter-logic suites (band x mode matrix, all 245 ADIF tokens) re-run afterward with no regressions
- [x] All commits signed (SSH) and verified against the repo's key

Ready for review.


int2001:
Wait a sec please. That contribution is awesome. i really like. But there are a couple of issues, i'd like to discuss with you.

1. as far as i can see this PR include the commits from #3633 , #3635 - since you're working on one thing (improve Counties award) we strongly propose to close those ones, and work with this one. Please close the other two.
2. Counties geoJSON is >25MB - a lot. please add some caching to it. e.g.: https://github.com/wavelog/wavelog/blob/af3256140bd05403b7c4a421746c2ea653a4f04f/application/controllers/Activationplanner.php#L146
3. (Pre-)Checked QSL-Methods: Should take care of what really counts for the award at ARRL. imho its QSL and LoTW only. so we'd appreciate if those both are preselected. of course user can add additional eqsl, qrz or clublog...

if you need any help, let me know.

Reid-n0rc:
>  3. (Pre-)Checked QSL-Methods: Should take care of what really counts for the award at ARRL. imho its QSL and LoTW only. so we'd appreciate if those both are preselected. of course user can add additional eqsl, qrz or clublog...

ARRL does not administer this award, it is [MARAC](https://www.marac.org/). The requirement is actually just the honor system. Per the [rules](https://www.marac.org/documents/usa-ca-rules.pdf) "Confirmation by QSL or electronic confirmation is not required." While not required some operators (me) might want to try to get those confirmed and track that as well. 

I'm happy to make QSL and LoTW default, as that's how I'd like to do it. I just thought I should fill you in on the requirements for this award.

I've closed #3635 and #3633 and will focus on implementing this one. I will look at how Activationplanner.php implements caching and try to apply it here.

int2001:
great. go ahead.
i'll be here if you've questions regarding architecture or code

Reid-n0rc:
@int2001 caching's in - the six boundary GeoJSON files are now served through a new `Awards::counties_geojson()` endpoint with the same ETag/`304` pattern as `Activationplanner.php`'s directory endpoints, so the ~24MB combined payload is only downloaded once per browser rather than on every "Show Counties Map" click. Two deliberate tweaks from the reference: a stat-based ETag (`filemtime()`+`filesize()`) instead of hashing the payload, since these are read from disk rather than already in memory from a DB query; and a 30-day `max-age` instead of 1 hour, since this data only changes on a Wavelog upgrade. Details and test results are in the updated PR description. Also went ahead and made QSL/LoTW the default pre-checked sources per your point 3.

int2001:
> @int2001 caching's in - the six boundary GeoJSON files are now served through a new `Awards::counties_geojson()` endpoint with the same ETag/`304` pattern as `Activationplanner.php`'s directory endpoints, so the ~24MB combined payload is only downloaded once per browser rather than on every "Show Counties Map" click. Two deliberate tweaks from the reference: a stat-based ETag (`filemtime()`+`filesize()`) instead of hashing the payload, since these are read from disk rather than already in memory from a DB query; and a 30-day `max-age` instead of 1 hour, since this data only changes on a Wavelog upgrade. 

sounds good

> Details and test results are in the updated PR description. Also went ahead and made QSL/LoTW the default pre-checked sources per your point 3.

QSL. tell me when ready to check. take your time

Reid-n0rc:
@int2001 should be ready for review. Do you want me to add test data to the project? If so where should that be put?

int2001:
> @int2001 should be ready for review. Do you want me to add test data to the project? If so where should that be put?

i think i've enough data to test here :)
give me a bit time - currently trying to melt-down the Rest of the PR-PileUp with the team.
May take till tomorrow until i've reviewed your PR

Reid-n0rc:
No rush @int2001 take your time and thank you for reviewing this!

int2001:
Ok.
Starting from user-perspective without having a look at code. Things i noticed:
* Filter:
  * See how other awards filter work. especially those for SAT. We need to take care for "unified" look here. it's a dropdown at the other Award-pages. And: See how deriving of bands for awards works. it looks like you added an own list instead of using band- or bandxuser-table.
  * Digi/SSB/CW ... like that. but not unified to the other awards.
* Worked/Not worked/Confirmed
  * Matching seems to be case-sensitive. While a lot of counties are shown in table as confirmed here (e.g.: "PA,LANCASTER", the map is still showing that as unconfirmed. i think that's an easy one (compare `uc` to `uc`)
  * CNTY consist of [STATE],[CNTY] according to [adif.co.uk](https://www.adif.org.uk/317/ADIF_317.htm#Sponsor_Defined_Code_Format) - looks like nothing (which was confirmed in my log) is green

Stopping for now. since the PR is huge, i think we better review chunkwise

Reid-n0rc:
@int2001 Thanks for the review - all three points addressed in 2214ebe75:

* **Filters unified**: Band and Mode are now single-select dropdowns matching the WAS/RAC/Helvetia convention - Band sourced from `Bands::get_worked_bands('uscounties')` (the `bands`/`bandxuser` tables), Mode from `Modes::active()`, instead of the bespoke multi-select band list and Phone/CW/Digital grouping.
* **Case-sensitive matching**: fixed - county/state names are now matched case-insensitively when building the map's status keys.
* **CNTY format**: you were right that this was the real bug. `COL_CNTY` is stored as `"STATE,COUNTY"` (e.g. `"PA,LANCASTER"`) for US QSOs per the ADIF Secondary Administrative Subdivision format - the map's status keys weren't stripping that prefix before matching against the boundary GeoJSON, so almost nothing could ever show confirmed. Fixed by normalizing `COL_CNTY` in SQL before the match.

Tested locally by seeding your exact "PA,LANCASTER" case plus a lowercase-prefixed variant and confirming both now render correctly on the map.

One open question on the Mode filter now that it's unified with the other awards' per-mode dropdown: USA-CA's own endorsement structure groups modes into Phone/CW/Digital (not per individual mode), so a literal "every logged mode/submode" dropdown like WAS uses doesn't map as cleanly onto how this award is actually earned. Do you think we should keep the fully unified per-mode dropdown as it is now, or would a Phone/CW/Digital grouping (still as a single-select, just three options instead of one per mode) be a better fit for this award specifically?

int2001:
Hi @Reid-n0rc 
this is much better now. 

Re Modes/Bands: Fair point with multiple things. what about a dropdown with multiselect, like we did it in SAT Passes?
image

But there are still issues from user-perspective. 
Re Map/Confirmations and so on:
While map now works and aligns with table (content), the details for a state are not listed, when clicking on a county.

Map / Details for `ME,AROOSTOOK`
image

List for `ME`
Screenshot 2026-08-12 at 05 48 57

Besides the userperspective i had a short first look into the code. i noticed SQLs with literals in it. We're using prepared-statements (with bound variables) everywhere for security reasons.

AndreasK79:
@Reid-n0rc great work. I was discussing it with @int2001 about a map when I updated the counties award a few weeks back, but we never implemented it.

You are missing a statement for map colors:

> if (typeof(user_map_custom.qso) !== 'undefined') {
>       workedColor = user_map_custom.qso.color;
> }

Regarding modes, perhaps fetch just used modes from the locations instead.
See logbookadvanced_model -> get_modes().

Reid-n0rc:
Thanks both for the review - pushed commits addressing everything raised.

**@int2001**, from your second round:
- Band/Mode filters are now a multi-select dropdown (same widget as the Satellite Pass page's satellite picker), matching your suggestion - `936c4be64`.
- `Counties.php`'s query functions (`get_counties_summary`/`get_counties`/`get_county_counts`/`get_counties_map`) now bind the station/band/DXCC lists as parameters instead of concatenating them as literals - `62d0a92ad`.
- Clicking a county on the map now correctly returns its QSO details (repro'd and fixed your `ME,AROOSTOOK` case) - the click handler was passing the map's bare county name straight to the lookup, which never matched `COL_CNTY`'s stored `"STATE,COUNTY"` format - `f495ef0f5`.

**@AndreasK79**, from your comment:
- `countiesmap.js` now reads `user_map_custom.qso.color` for the "worked, not confirmed" state, same as `wasmap.js` does - `5a81689c4`.
- The Mode filter now lists modes/submodes actually present in your logbook rather than every ADIF-active mode on the install, using the same `Logbookadvanced_model::get_modes()` the Logbook Advanced Search and Statistics pages already use for their mode filters - `5a81689c4`.

All changes tested locally against a seeded logbook (multiple QSL sources, band/mode combinations, and the mixed-case/prefixed CNTY formats mentioned above).

Reid-n0rc:
Well my script to sign everything made a mess, sorry about that

AndreasK79:
@Reid-n0rc looks better for each commit :)

I managed to crash the browser with the zoom:
image

Add minZoom: 3 after line 107 in countiesmap.js. That should take care of it.

Reid-n0rc:
@AndreasK79 minZoom added

int2001:
From user-perspective it looks clean now. i try to dig into code during the next few days.
see my adjustment at #3662

phl0 (changes requested):
Filter for SAT does not work and produces bogus SQL code:

```
[...]
and COL_BAND != 'SAT'
and COL_BAND in ('60m','40m','30m','20m','17m','15m','12m','10m','6m','2m','70cm','23cm','13cm','SAT')
[...]
```

SAT is not a band in its sense but a filter on prop mode being 'SAT'. I will try to come up with a solution to this.

Reid-n0rc:
@phl0

`COL_BAND` is never actually `'SAT'` - satellite QSOs are flagged via `COL_PROP_MODE = 'SAT'`, with `COL_BAND` holding the real band worked. The Band filter's `'SAT'` option was getting folded into `COL_BAND in (...)`, and there was also a `COL_BAND != 'SAT'` guard elsewhere in the same query that never matched anything either way. Combined, that's the impossible SQL.

Fix: `band_condition()` splits `'SAT'` out of the selected bands and ORs in a `COL_PROP_MODE = 'SAT'` clause instead. Replaced the dead `COL_BAND != 'SAT'` literal with a `sat_exclusion()` helper - excludes satellite QSOs by default, backs off when SAT is explicitly selected.

Also fixed the same gap in the county QSO-detail dialog (map click / state list) - it ignored the Band/Mode filter entirely and had its own hardcoded SAT exclusion, so a county shown confirmed under a SAT-only filter listed zero QSOs when clicked.

AndreasK79:
@Reid-n0rc before we can do anything more here. Check this one https://github.com/Reid-n0rc/wavelog/pull/1

You also need to fix the merge conflict.

application/models/Logbook_model.php
```diff
@@ -6108,18 +6090,54 @@ function mark_lotw_ignore($station_id) {
...
+		// confirmed on the map but list zero QSOs when clicked.
+		$this->load->model('counties');
+		$binding = array_merge($logbooks_locations_array, [strtoupper($state), strtoupper($bare_county)]);
+		$band_condition = $this->counties->band_condition($band, $binding);
+		$mode_condition = $this->counties->mode_condition($mode, $binding);
+		$sat_exclusion = $this->counties->sat_exclusion($band);
+
+		$sql = "select * from $table thcv
+			join station_profile on station_profile.station_id = thcv.station_id
+			left outer join lotw_users on lotw_users.callsign = thcv.col_call
```
phl0, line 6132:
```suggestion
			left outer join lotw_users on lotw_users.callsign = thcv.col_call
			left outer join satellite on satellite.name = thcv.COL_SAT_NAME
```

application/models/Logbook_model.php
```diff
@@ -6108,18 +6090,54 @@ function mark_lotw_ignore($station_id) {
...
+		// showing - it previously ignored Band/Mode entirely and hardcoded
+		// SAT out, so e.g. filtering to just "SAT" would show a county as
+		// confirmed on the map but list zero QSOs when clicked.
+		$this->load->model('counties');
+		$binding = array_merge($logbooks_locations_array, [strtoupper($state), strtoupper($bare_county)]);
+		$band_condition = $this->counties->band_condition($band, $binding);
+		$mode_condition = $this->counties->mode_condition($mode, $binding);
+		$sat_exclusion = $this->counties->sat_exclusion($band);
+
+		$sql = "select * from $table thcv
```
phl0, line 6130:
```suggestion
		$sql = "select *, satellite.name as sat_name, satellite.displayname as sat_displayname from $table thcv
```

application/controllers/Awards.php
```diff
@@ -1381,19 +1547,28 @@ public function counties_details_ajax() {
 
         $state = str_replace('"', "", $this->security->xss_clean($this->input->post("State")));
         $county = str_replace('"', "", $this->security->xss_clean($this->input->post("County")));
-        $data['results'] = $this->logbook_model->county_qso_details($state, $county);
+        $band = $this->counties_band_from_post();
+        $mode = $this->counties_mode_from_post();
+        $data['results'] = $this->logbook_model->county_qso_details($state, $county, $band, $mode);
 		$data['adif_propmodes'] = $this->config->item('adif_propmodes');
```
phl0, line 1553:
```suggestion
        $data['adif_propmodes'] = $this->config->item('adif_propmodes');
```

phl0 (changes requested):
Needs a few more SQL for satellites. See below. Otherwise the QSO modal will run into errors (at least for satellite QSO):

Image

In addition the merge conflict in Logbook_model needs to be solved. Let me know if you need support here.

phl0:
An update here. I missed @int2001's PR against your feature branch. I just tested it and it works fine. So from my end it would make sense to merge https://github.com/Reid-n0rc/wavelog/pull/1 into this branch. And then solve the merge conflict in Logbook_mode.php. Then everything should be fine and good to go.

phl0:
Ping @Reid-n0rc ;-)

int2001:
@Reid-n0rc - if you don't mind, i'd take over. we still like the idea and want it in Wavelog. But it needs some afterwork.

Reid-n0rc:
Sorry for disappearing, everyone. It's been a busy few weeks at work and at home with the school year starting back up and dealing with that. 

@int2001 please take over, I don't mind at all. Thank you for taking that on.

PR #3641, opened 2026-08-11
Docker development

HB9HIL:
it was not possible to enable debug mode in docker containers. creating `.debug` in a container is kinda cumbersome so for docker there is a new option to set `DOCKER_DEVELOPMENT` as env in docker or somewhere else so you can enable the developer mode also in docker container without handling any files. dev mode in normal installations is untouched. Since I now saw some issue i removed some deprecated stuff.


int2001 (approved):
LGTM. tested it - works (even if debug within docker is ... rarely used?)

HB9HIL:
> LGTM. tested it - works (even if debug within docker is ... rarely used?)

I use it, that's enough :)

PR #3648, opened 2026-08-11
replaced star history with a gif

HB9HIL:
I removed the Start History from the README. I really really liked it... But Star History is struggling. It's not their fault. 

https://github.blog/changelog/2026-06-30-upcoming-access-restrictions-to-public-api-endpoints-and-ui-views/

By the end of June this year github restricted their API for metadata and stargazer data. It now requires a read token for metadata (which would not be that bad) but also the scope contents:write, which is potentially a big issue. If there appears any vulnerability there would be a token which full write access to the repo. This is not okay - and the Star History project suffers from that. 

I replaced it with a GIF for now. Maybe I find another solution but at the moment I think its a bad idea to use such a open scoped token for such a graph. 

This is the GIF: 

star_wavelog

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,13 +97,7 @@ When submitting PRs please make sure code is commented and includes one feature 
 
 Support Wavelog by starring the repository on GitHub!
 
-<a href="https://www.star-history.com/?repos=wavelog%2Fwavelog&type=date&legend=top-left">
- <picture>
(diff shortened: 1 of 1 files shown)


int2001 (approved):
even if i liked star-history as well, always good to reduce external dependencies

phl0 (approved):
Fine with me

PR #3649, opened 2026-08-11
[Logbook] Fix Club Log confirmation handling

dg0tm:
This PR fixes a few small issues around Club Log confirmation handling.

  - Mark Club Log entries as modified when paper QSL sent/received state changes.
  - Store Club Log upload/download timestamps as DATETIME values.
  - Replace the repeated "set status to M" update block with a shared helper.

This is a small correctness and cleanup change with limited user-facing impact.

VIA-only changes are intentionally ignored for Club Log modified detection.
QRZ VIA handling is not included in this PR.


int2001:
As far as i can see you set the clublog-thing to `M` if a Paper-Card has been either `S`ent or `I`gnored (Why ignore?).
so far so good - seems logical.

but:
- at LBA-bulkedit you update the clublog to `M`odified for every change which happens on the paperfield. intended? slightly different logic than at edit...
- Furthermore: We didn't implement this for a specific reason. Clublog seems to ignore Re-Uploads. They built in a non-adif-field called `QSLCALL` (see [here](https://clublog.freshdesk.com/support/solutions/articles/54908-using-adif-to-update-or-delete-qsos)). Documentation there is a bit thin. Do they need that field always? only for updating a call within a QSO? One doesn't know.

tbh: we never tested it, because it looked a bit fishy. At the very end Clublog itself isn't used for awards but for personal statistics or info about "am i in the log". so the real question is: is it worth the effort to pass the information of a received papercard (without any proof) to the service? not sure

dg0tm:
This PR adds two missing Club Log requeue hooks for paper QSL actions.

What changed:

  - paperqsl_update_sent() now also marks Club Log as modified.
  - paperqsl_requested() now also marks Club Log as modified.

Why:

  - These paper QSL actions should follow the same behavior as the existing QRZ flow for the same workflow.
  - The user should stay in control of the QSL state in the UI.
  - Club Log requeueing stays limited to real paper QSL state changes.

What this does not change:

  - Upload/download behavior itself.
  - Any Club Log API behavior or queue semantics.
  - LBA/bulk handling. I intentionally left that out here, because it would require a larger refactor and would also touch other sources.
  - A concrete upload proof from a real sync run. That still comes later.

Reviewer note:

  - This is a very small follow-up, intentionally limited to two lines in the single-edit paper QSL flow.

int2001:
Ok, just tested re-upload to clublog. works. that was my only concern. so go ahead.

int2001 (approved):
checked.
works. even if an update for QSL_VIA would also trigger update to clublog. i don't care - should be harmless

phl0 (approved):
Works as expected. IMHO it makes sense to also check LoTW for changes and set Clublog state to modified if so.

PR #3651, opened 2026-08-12
Various fixes concerning grids

phl0:
This PR aims to fix various issues:

Editing QSO allows to store bogus grid in gridsquare field resulting in errors on QSO view and shows distance to be 0km (default value on errors in distance calculation):

image

Filling grid and vucc grid in QSO edit results in bogus display of grids:

image

Storing more than 12 chars in gridsquare field on QSO edit resulted in bogus error message popup:

image

We changed:

- Return value of distance calculation in case of errors to false (was 0)
- Prefer display of VUCC_GRIDS over GRIDSQUARE and only show one of them
- Hide Distance row in case of errors on calculation
- Added separate checks for GRIDSQUARE and VUCC_GRIDS in QSO edit
- Fixed bogus error message in case of erroneous date in gridsquare or vucc_grids

### DIFF
--- a/application/controllers/Qso.php
+++ b/application/controllers/Qso.php
@@ -803,7 +803,9 @@ class QSO extends CI_Controller {
 		switch ($type) {
 		case 'grid':
 			$grid = $this->input->post('locator', TRUE);
-			$this->load->library('Qra');
+			if (!$this->load->is_loaded('Qra')) {
+				$this->load->library('Qra');
+			}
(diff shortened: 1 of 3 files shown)


int2001:
Findings so far:
- [x] every QSO MUST have a grid now when logging. Intended?
- [x] even if grid provided as told in hint: Error:

image

int2001 (approved):
works

PR #3653, opened 2026-08-12
Attempt to remove ALL occurences of the slashed zero...

int2001:
...and replace it with an opensourcefont.

I think here's still a lot of effort to be invested in it. but it's a foundation.
- [x] Most of pages where call is displayed
- [x] Searchfields
- [x] map
- [x] Globe (not feasible, therefore dropped)
- [x] Notes for contacts. needs migration if wanted.


phl0 (approved):
Fine with me so far.

int2001:
same ;)

HB9HIL (approved):
already run this on various test instances. seems okay so far

PR #3656, opened 2026-08-12
Count needed DXCC based on status and show deleted DXCCs separately

phl0:
The current DXCC query for dashboard counts worked DXCCs regardless of whether the DXCCs are (still) valid or not. So we add a criteria for the worked DXCCs to only count the not-deleted ones. Deleted ones are to be counted separately and shown in brackets on the dashboard.

Screenshot from 2026-08-12 14-52-23

### DIFF
--- a/application/controllers/Dashboard.php
+++ b/application/controllers/Dashboard.php
@@ -131,6 +131,7 @@ class Dashboard extends CI_Controller {
 
 		// Country stats
 		$data['total_countries'] = $stats['Countries_Worked'];
+		$data['total_deleted_countries'] = $stats['Countries_Deleted_Worked'];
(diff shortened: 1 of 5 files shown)


int2001:
Nice one. worked.
BUT: Since we've to take care of all multi-whistles/toys we've it'd be consequent to add that "deleted" one as well to APIv2. Somewhere here: https://github.com/wavelog/wavelog/blob/58e67ab00af646941363dcca8432eade32200cb7/application/libraries/api_v2/Statistic_resource.php#L153

AndreasK79 (approved):

AndreasK79:
Numbers looks better now :)

PR #3657, opened 2026-08-12
Session Locks and New Redis2 Session Driver

HB9HIL:
This adds `session_write_close()` to expensive ajax calls in QSO and Contesting to reduce blockings. 

While default session driver `files` is really fast and default (which is good), in some environments you need a shared session storage. For example in Kubernetes there is no way around that. As we use apache2 in our docker image real parallel requests are not possible here so reducing the blocking time between serialized ajax requests is crucial for a good performance. 

The legacy CI3 redis session driver uses polling against redis als sleeps for 1 second if the session lock was not released yet. In 2014 when codeigniter was designed, this probably was okay. Today it isn't as this causes a awesome (not) 1s staircase which massivle drops performance on redis session handlers in containerized environments. 

See here:
```php
# branch wavelog/dev in docker
$config['sess_driver'] = 'redis'; 
```

grafik

So I dropped the sleep in the legacy redis and memcached session driver to `100ms` (timeout still at 30s as it was before)

grafik

This causes 10 times more polling traffic between server and redis, which is nowadays not an issue. Also the 100ms polling shouldn't be a big deal against redis (if this causes issues, the infrastructure is just crap)

This reduction in sleep let's php picking up the session much much earlier instead waiting a full 1 second for the next request. 

So this is just a security net for those who stay with the legacy redis session driver and we all know polling is not that sexy. So I prepared a new session driver called `redis2` which lives in /application and not /system. This redis driver uses modern technology and doesn't poll redis to ask for the session lock. Instead it uses `BLPOP` which notifies php when the session is released from the lock. This reduces the waiting time between ajax calls to <1ms and doesn't require any polling. Unfortunately memcached doesn't support that. So we stick with polling here

```php
# this branch in docker
$config['sess_driver'] = 'redis2'; 
```

grafik

With `redis2` the performance is nearly as good as with the files driver while being fully compatible with the session locking. 

Cheers

### DIFF
--- a/application/config/config.sample.php
+++ b/application/config/config.sample.php
@@ -490,7 +490,13 @@ $config['encryption_key'] = 'flossie1234555541';
 |
 | 'sess_driver'
 |
-|	The storage driver to use: files, database, redis, memcached
+|	The storage driver to use: files, database, redis, redis2, memcached
+|
+|	'redis2' is Wavelog's own Redis driver. It behaves like 'redis' but waits
+|	for the session lock with BLPOP instead of polling, so parallel AJAX
+|	requests of the same user are no longer delayed by up to a second each.
+|	When redis is needed, use 'redis2' instead of 'redis' for better performance
+|   unless you have a really good reason to use the original 'redis' driver.
 |
 | 'sess_cookie_name'
 |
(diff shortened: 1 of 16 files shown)


int2001 (approved):
Tested:
- old redis driver
- new redis2 driver
- default file driver

untested: docker (but should work, since it uses defaults)

good one - speeds things up, if using redis2

PR #3659, opened 2026-08-12
escape stored data

HB9HIL:
escape all adif data which could be shown in the UI


HB9HIL:
I just went through the code and couldn't spot any obvious. I guess we can merge this one and see what happens. If some other places pop up we'll fix them accordingly

phl0 (approved):
Agree. No issues found so far

int2001 (approved):
ok for now.
but we should keep track whats still open:

- DOK, IOTA and at some pages the county

PR #3663, opened 2026-08-13
APIv2 should spit out deleted / confirmed as dashboard does

int2001:
Changes at API should follow same logic like in #3656 and #3615 

testing (caution: api takes ALL stations of user, if not explicity station_id requested):

`curl -s "https://[WL_URL]/api/v2/statistic" -H "Authorization: Bearer [v2_key]" |jq .`

Documentation adjusted as well: https://docs.wavelog.org/developer/api-v2/statistic/#qso-default


phl0 (approved):
Works and shows exactly the same numbers as on dashboard.

PR #3665, opened 2026-08-13
[Advanced Logbook] Add a fix all button to the state fixer in dbtools

AndreasK79:
Proposed here #3643.
image

One button to ~rule~ fix them all.


phl0 (commented):
Tested and works so far I can see :D

int2001 (approved):

PR #3613, opened 2026-08-09
feat(api): add API v2 contest resource

mcules:
This adds a `contest` resource to API v2, exposing contest sessions and
their QSO links. Motivation: external tools (offline mirrors / sync
clients such as field-day laptops, contest loggers) currently cannot
replicate contest sessions — QSOs travel fine via the QSO resource, but
the session objects have no API and must be rebuilt manually through
*Import Historical Contests* on every instance.

### Endpoints

| Verb | Path | Scope | Purpose |
|---|---|---|---|
| GET | `/api/v2/contest` | `contest:read` | List the token owner's sessions (`?station_id=`, `?since_id=` for incremental sync) |
| GET | `/api/v2/contest/{id}` | `contest:read` | Single session incl. linked QSO ids |
| POST | `/api/v2/contest` | `contest:write` | Create a session, optionally linking QSOs (`qso_ids`) |
| PATCH | `/api/v2/contest/{id}` | `contest:write` | Partial update; link/unlink QSOs via `link_qso_ids` / `unlink_qso_ids` |
| DELETE | `/api/v2/contest/{id}` | `contest:delete` | Delete a session; `?delete_qsos=true` also removes the QSOs (full teardown) |

### Design notes

- The contest is addressed by its **ADIF name** (`"contest": "DARC-WAG"`);
  responses carry both `contest` (adifname) and `contest_name`. The numeric
  catalog ids are instance-local, so they are unusable for cross-instance sync.
- QSO links travel in the request body (`link_qso_ids`/`unlink_qso_ids`)
  because the v2 URL space intentionally has no sub-paths. Linking is
  **idempotent** and batched; QSOs already linked to a *different* session
  are skipped and reported (`skipped`), so a sync client can re-send its
  full list.
- `settings` mirrors `Contesting_model::build_session_settings()`: create
  merges over the same defaults, PATCH merges over the stored settings and
  only validates the incoming keys — stored keys a future version added
  survive a partial update. Unknown keys in a request are a 400 so clients
  notice typos.
- QSO ownership is checked with the same join semantics as
  `Logbook_model::check_qso_is_accessible()`, batched into one query.
- Ownership rules follow the existing resources: foreign stations/QSOs are
  403, foreign sessions 404, empty PATCH bodies 400. Clubstations: create
  is open to members, update/delete require officer level 9 — exactly like
  the Contesting controller (`clubaccess_check(9)`).
- Scopes (`contest:read/write/delete`) join the token registry and the
[...]

PR #3063, opened 2026-03-14
New Wavelog Contesting and Basics for Wavelog Worker

HB9HIL:
> [!TIP]
> Contributors Welcome. Make a PR against [HB9HIL:sophisticated_contesting](https://github.com/HB9HIL/wavelog/tree/sophisticated_contesting)

> [!TIP]
> Test this feature in the public beta demo
> https://new-contesting-demo.wavelog.org/
> User: demo, Password: demo

> [!TIP]
> Documentation: https://docs.wavelog.org/user-guide/contesting/

# TODO List

## Priority 1
- [x] Radio component needs to be tested in any ways
- [x] "Worked Before" in QSO Form
- [x] Websocket Radio
- [x] Options for other exchange types are missing, and Serial Number is also missing (a toggleable counter)
- [x] ADIF/CBR import is completely missing for the new contesting module. The logic is also different here
- [x] We need to find a way how to "migrate" existing contest QSOs and create session from it - or not? Not sure here
- [x] Fix other themes then darkly. They are broken atm
- [x] Keybindings: Space to Rapport or Exchange, etc.
- [x] Exchange types behaviour should match the logic of the old contesting
- [x] Proper localization
- [x] Fix "please select a contest" in contest dropdown
- [x] Allow editing qso time
- [x] Make map more precise if grid was entered
- [x] If grid was entered calculate azimut

## Priority 2
- [x] Check inline edit for QSOs — not sure if it's easy to implement due to DataTables (it works, but need to check if it won't get too bloated)
- [x] Asynchronous callbook lookup would be nice for rotor control. DB-first approach!
- [x] UI/UX around contest session management needs another pass
- [x] Clean up code and search for leftover development artifacts
- [x] Stats component with points and statistics is still missing — implementation unclear, but it should be included. Stats only for the moment
- [x] Address #3056 
- [x] Address #3058 (may can @AndreasK79 help here)
- [x] Address #3245 

## Priority Maybe
- [x] Map component can be extended with features like night shadow, grid, or path lines. Must not affect performance!

# The New Contest Logger
[...]

PR #3666, opened 2026-08-13
Fix COL_CONTEST_ID

HB9HIL:
PR #3063 introduced a bug in set_contest() which set the COL_CONTEST_ID to the display name instead the ADIF name. This is wrong and this PR fixes this. To also fix existing QSOs it needs a migration. Since this is a heavy migration it set's an index before which helps a lot in this case and for later contest stuff performance. 

This popped up while reviewing #3613

### DIFF
--- a/application/migrations/294_fix_contest_id.php
+++ b/application/migrations/294_fix_contest_id.php
@@ -15,11 +15,13 @@ class Migration_fix_contest_id extends CI_Migration
 	{
 		$table_name = $this->config->item('table_name');
 		$ix_name = 'HRD_IDX_COL_CONTEST_ID';
+		$ix_created = false;
 
 		$ix_exist = $this->db->query("SHOW INDEX FROM {$table_name} WHERE Column_name = 'COL_CONTEST_ID' AND Seq_in_index = 1")->num_rows();
 
 		if ($ix_exist == 0) {
 			$this->dbtry("ALTER TABLE {$table_name} ADD INDEX `{$ix_name}` (COL_CONTEST_ID)");
+			$ix_created = true;
 		}
 
 		$this->dbtry("UPDATE {$table_name} t
@@ -33,6 +35,10 @@ class Migration_fix_contest_id extends CI_Migration
 							HAVING COUNT(*) = 1
 						) c ON t.COL_CONTEST_ID = c.name
 						SET t.COL_CONTEST_ID = c.adifname");
+		
+		if ($ix_created) {
+			$this->dbtry("ALTER TABLE {$table_name} DROP INDEX `{$ix_name}`");
+		}
 	}
 
 	public function down()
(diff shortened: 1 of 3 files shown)


phl0 (approved):
Code works. Not having any contests in the log though ...

HB9HIL:
thx @phl0. since this is a really heavy query I'd appreciate a second pair of eyes here.

phl0:
> thx @phl0. since this is a really heavy query I'd appreciate a second pair of eyes here.

Sure no worries. Might even generate some test data for a test tomo.

application/migrations/294_fix_contest_id.php
```diff
@@ -0,0 +1,50 @@
...
+{
+	public function up()
+	{
+		$table_name = $this->config->item('table_name');
+		$ix_name = 'HRD_IDX_COL_CONTEST_ID';
+
+		$ix_exist = $this->db->query("SHOW INDEX FROM {$table_name} WHERE Column_name = 'COL_CONTEST_ID' AND Seq_in_index = 1")->num_rows();
+
+		if ($ix_exist == 0) {
+			$this->dbtry("ALTER TABLE {$table_name} ADD INDEX `{$ix_name}` (COL_CONTEST_ID)");
```
int2001, line 23: this index is only for updating during migration?
must be, because station_id (which is in 99% of al selects) is missing.

if it's really only for mig, i'd propose to drop it afterwards.

alternative (but need to check careful how it impacts adif-insert on slow or huge machines):
add station_id to it. makes only sense if existing indexes are not covering it (assumption: existing station_id -idx is selective enough on a normal log)

int2001 (changes requested):
see comment regarding new idx on maintable.

this one can really become tricky on slow/huge instances.

consider bumping the `SET SESSION lock_wait_timeout` in maria/mysql.

further consideration: avoiding that inline-loop and creating a `update set t.logbook_contest=case when logbook_contest='xx' then 'xyz', when ...` etc (dynamically created in php by selecting possible contests first).

inline-loop / possible problems:
imagine instance with 27Mio QSOs (10% are contest). DB-Access with the nested-loop (marias optimizer always choses nested-loops - wasn't able to force hash-join) costs: 2.7Mio multiplied by amount of contests in contesting-table.

int2001 (approved):

HB9HIL:
> see comment regarding new idx on maintable.
> 
> this one can really become tricky on slow/huge instances.
> 
> consider bumping the `SET SESSION lock_wait_timeout` in maria/mysql.
> 
> further consideration: avoiding that inline-loop and creating a `update set t.logbook_contest=case when logbook_contest='xx' then 'xyz', when ...` etc (dynamically created in php by selecting possible contests first).
> 
> inline-loop / possible problems: imagine instance with 27Mio QSOs (10% are contest). DB-Access with the nested-loop (marias optimizer always choses nested-loops - wasn't able to force hash-join) costs: 2.7Mio multiplied by amount of contests in contesting-table.

such huge instances are the absolut edgecase. So I guess dropping the ix is okay. The rest is okay I guess for 99% of the users

PR #3668, opened 2026-08-14
Remove SAT timers from tools

phl0:
As currently (and for the past month) none of the observed birds was in a usable state and no point of central information about activations currently existing we remove SAT timers from tools. May be re-added at a later point if situation changes.


application/views/interface_assets/header.php
int2001: remove this also including the css-file?

int2001 (changes requested):
foother.php L 370 - also orphaned sattimers-stuff, as far as i can see.

phl0:
Fixed

HB9HIL (approved):
looks good to me :)

int2001 (approved):

PR #3671, opened 2026-08-14
[Activation Planner] Added search for places

AndreasK79:
image


int2001 (approved):
solid one. who needs google or a routeplanner, if you can have WL?

PR #3628, opened 2026-08-10
feat(api v2): logbook CRUD, DXCC list, station set-active

MaviKulubeliAdam:
## Summary

This PR adds full Wavelog Mobile app support to the API v2, making the mobile
patch unnecessary for most features.

### New token preset
A "Wavelog Mobile" one-click preset in the token creation dialog that
auto-selects all required scopes.

### New API v2 resources
- **Logbook** (`GET/POST/PATCH/DELETE /api/v2/logbook[/{id}]`)
  — CRUD, set active, link/unlink station locations
- **DXCC** (`GET /api/v2/dxcc[/{id}]`)
  — Read-only entity list; `?subdivisions=true` returns states/provinces
- **Contest** (`GET/POST/PATCH/DELETE /api/v2/contest[/{id}]`)
  — Session CRUD; `?templates=true` lists available contest definitions

### Station resource extensions
- `PATCH /api/v2/station/{id}` with `active: true` — switch active station
- `POST /api/v2/station` with `clone_from: {id}` — clone a station

### QSO resource extension
- `POST /api/v2/qso` accepts optional `contest_session_id` — links the new
  QSO to a contest session

## App
- Android (Google Play): https://play.google.com/store/apps/details?id=com.wavelog_mobile

PR #3674, opened 2026-08-15
Feat/api v2 logbook crud dxcc station set active

MaviKulubeliAdam:
Hi again, after talking with @HB9HIL I'm replacing #3628 with this one.

Main change: resource files were hitting the DB directly moved everything 
through the model layer instead. Also dropped Contest_resource since #3613 
already has a proper implementation.

- Logbooks_model: new user_id-based helpers for the logbook endpoints
- Dxcc: get_by_adif() and get_subdivisions()
- Stations: switch_active()
- Removed the wavelog_mobile preset from Api_v2_model


HB9HIL:
please update the branch and run `git pull` at your local machine 😄 

grafik

MaviKulubeliAdam:
I think I finally managed to create the PR properly this time  :smile: I hope I've implemented all the necessary corrections completely. Thank you so much for all the feedback and for the meeting we had. @HB9HIL

int2001:
Tried to create an "All-In" Token. since we've more and more scopes, VC 255 doesn't seem to be enough. When i enabled all, i receive a `Message: Data too long for column 'scopes' at row 1` from php.

[edit]: my `dev` had varchar(255) instead of text. don't know the reason. so ignore the thing with "all-in-token" and the error. sorry

Rest works.

Even if i've the open question why WL should deliver DXCC-Information and subdivisions for 3rd-party applications.
WL isn't master here, and that data is also public available. perhaps i miss something

HB9HIL:
@MaviKulubeliAdam your last commit is not needed. APIv2 token scope colum is already text. probably you checked out the apiv2 branch during development

https://github.com/wavelog/wavelog/blob/af3256140bd05403b7c4a421746c2ea653a4f04f/application/migrations/288_api_v2_tokens.php#L23

MaviKulubeliAdam:
@HB9HIL Thanks for the warning, I've sorted it out 🙏🏻

HB9HIL:
@MaviKulubeliAdam I removed the DXCC ressource and replaced it with a more generic catalog ressource which was implemented in #3700

HB9HIL (approved):
This is now ready 2 merge from my side. Needs at least one more approval by a wavelog team-member

int2001 (approved):

HB9HIL:
documentation added 
https://docs.wavelog.org/developer/api-v2/logbook/
https://docs.wavelog.org/developer/api-v2/catalog/
https://docs.wavelog.org/developer/api-v2/station/

PR #3675, opened 2026-08-15
[JCC Award] Revived and refactored

AndreasK79:
Based on works by @cebarobot here #3151.

Simplified and cleaned up. Also updated the page layout for existing pages.

image
image


HB9HIL (commented):
did a quick overview without testing. couldn't spot any big issues so far

phl0 (approved):
Tested and works.

PR #3679, opened 2026-08-15
OptionsLib Refactoring

HB9HIL:
This PR brings a major refactoring of the `OptionsLib` which was implemented many years ago and barely touched since then. Issue here was that the optionslib loaded every option from the database in memory even they don't are used. Back then this may was a cool idea but today this just eats ressources. 

Therefore I dropped the whole config option autoload thingy and dropped the autoload column from database as it's not needed anymore. The OptionsLib itself is still autoloaded by PHP on each request since this manages the auto migrate. 

Result is a more efficient way of loading options and reducing overhead (on my mac locally in docker each pageload is now 10ms faster)

### DIFF
--- a/application/config/migration.php
+++ b/application/config/migration.php
@@ -22,7 +22,7 @@ $config['migration_enabled'] = TRUE;
 |
 */
 
-$config['migration_version'] = 294;
+$config['migration_version'] = 295;
 
 /*
 |--------------------------------------------------------------------------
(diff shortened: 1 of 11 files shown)


int2001 (approved):
let's throw it into dev and see what happens.
there are still (out of scope in this PR) XSS-Flaws which would need admin-access - we adress them in our escape_all PRs

PR #3680, opened 2026-08-15
PHP fixes

HB9HIL:
imagedestroy has no function since php 8.0. So we can get rid of it and of the deprecation warning aswell. 

additionally the imagettftext complained about missing font since there was an FCPATH missing. (Was just a warning, worked anyway, so basically removed the warning).

### DIFF
--- a/application/controllers/Eqsl.php
+++ b/application/controllers/Eqsl.php
@@ -507,10 +507,6 @@ class eqsl extends CI_Controller {
 
 		// Output
 		imagejpeg($thumbnail, null, 90); // 90% quality
-
-		// Clean up
-		imagedestroy($original_image);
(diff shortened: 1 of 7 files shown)


int2001:
not sure, do we need to fix this as well? https://github.com/wavelog/wavelog/blob/2139ce6dba84ad848a784b910a5fc220df625ee8/src/Label/tfpdf.php#L1623

HB9HIL:
> not sure, do we need to fix this as well?
> 
> https://github.com/wavelog/wavelog/blob/2139ce6dba84ad848a784b910a5fc220df625ee8/src/Label/tfpdf.php#L1623

fixed all other occurence. :-)

int2001 (approved):

PR #3681, opened 2026-08-15
option "center" for staticmap

HB9HIL:
as you can see in #3678 people may have the need to center the staticmap to a custom grid instead the center of all station locations. This adds the option `center` which allows to set a custom map center. Works only if no continent is selected 

`?center=JN47RH` 

grid is validated

### DIFF
--- a/application/controllers/Staticmap.php
+++ b/application/controllers/Staticmap.php
@@ -127,6 +127,13 @@ class Staticmap extends CI_Controller {
             $ituzones = $r == 'true' ? true : false;
         }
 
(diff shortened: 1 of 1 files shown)


HB9HIL:
reminder for me to add this to the documentation once merged

int2001:
in general it works, but things noticed:
- centers only on longitude. intended?
- bug found: when calling staticmap on a non-existing slug, there are errors. (looks like it's a bug from before the PR)

int2001 (approved):
solid one.

@HB9HIL adrresses the mentioned bug (404) in another PR

PR #3683, opened 2026-08-15
decouple maintenance mode from CI_ENV

HB9HIL:
the maintenance mode which was implemented in #173 should not be handled by the defined CI_ENV but with a dedicated constant. This way it will be possible to enable the maintenance mode in docker environments which is not possible at the moment. Behaviour itself did not change. Create `.maintenance` in webroot.


int2001 (approved):
good one.

PR #3685, opened 2026-08-16
[Dashboard] Add more cards to menu

AndreasK79:
@int2001 added the right click menu to the dashboard, which works great. Map, dxpeditions and contest card also have the options, so these are now added.

image


int2001 (approved):
works, and makes dashboard more intuitive/interactive

PR #3688, opened 2026-08-16
Amsat pills

AndreasK79:
Use pills instead of table. Tweaked layout for mobile as well.


int2001 (approved):
cool UI/UX improvement. i like

PR #3691, opened 2026-08-16
[Advanced Logbook] Add slider to adjust time for duplicate QSO search

AndreasK79:
Default is still 30minutes, but now you have a slider you can adjust between 0 and 60 minutes:
image

Requested here #3571 by @iv3cvn.


application/models/Logbookadvanced_model.php
```diff
@@ -46,9 +46,16 @@ public function dupeSearchQuery($searchCriteria, $binding) {
...
+		$dupe_time = (int)($searchCriteria['dupedateval'] ?? 0);
+		if ($dupe_time <= 0) {
+			$dupe_time = 1800; // default 30 minutes
+		}
+		$dupe_time = min($dupe_time, 86400);
+
 		// Build the dupedate HAVING condition
 		$having_condition = isset($searchCriteria['dupedate']) && $searchCriteria['dupedate'] === 'Y'
-			? "AND TIMESTAMPDIFF(SECOND, prev_time_on, col_time_on) < 1800"
+			? "AND TIMESTAMPDIFF(SECOND, prev_time_on, col_time_on) < $dupe_time"
```
phl0:
```suggestion
			? "AND TIMESTAMPDIFF(SECOND, prev_time_on, col_time_on) <= $dupe_time"
```

IMHO `<=` would be more plausible here.

phl0 (commented):
The patch works. Just on minor addition and a question: Would it be good if the dupe dialog remembered the last time setting for the case you want to adjust?

AndreasK79:
> The patch works. Just on minor addition and a question: Would it be good if the dupe dialog remembered the last time setting for the case you want to adjust?

I can add it into localStorage, so it is remembered that way.

AndreasK79:
@phl0 it's done

phl0 (approved):
Nice. works :)

PR #3692, opened 2026-08-17
[QSL Print Queue] make qslprint queue length configurable

DB4SCW:
While processing the QSL queue of a special callsign with over 2000 entries, I realized the amount of QSOs that the new redesigned QSL queue loads is hard-capped at 1000.

This introduces an optional config item to raise that limit so that big qsl-jobs can be loaded completely so that all the filters and editing options can be used in one step.

If user does not add this config item to their config, nothing changes from the previous hard-coded limit of 1000.

Also added a few more lengths on the javascript side for bigger jobs.


AndreasK79:
@DB4SCW that sounds like way too many QSLs to handle :)

Anyway, just set the limit to 3000 and be done. No config needed.

DB4SCW:
And then some WWA station shows up with 10000 QSOs...
High numbers of QSLs are really nothing special for special event stations, which are now more likely to use wavelog due to the clubstation feature.
So... Does that config really hurt?
I know where to look and how to change the code temporarily just for me but should that be the solution?

AndreasK79:
@DB4SCW perhaps it doesn't hurt, but trying to avoid yet another parameter in a config file.

I know @HB9HIL started looking into pagination, so perhaps he can pitch in a bit here.

DB4SCW:
Pagination would only solve this problem if things like the filter propagate on all pages, as well as "select all", "mark as sent" or "export as ADIF". If not, pagination will only make it even worse.
I really liked the possibility to do those QSL jobs directly in wavelog without having to export everything to handle it outside of wavelog (like other log programs almost force you to do)

But I'm more than happy if another solution is found.

HB9HIL:
Give me a bit time.

PR #3702, opened 2026-08-18
Count (raw) QSOs by...

int2001:
This one introduces a new Analytics-method called "Count QSOs by".

Sometimes it's interesting (e.g. for non-hams, news, posing ... :) ) to see how many QSOs with Grids / Parks / DXCCs / ... were made during a period of time.

Or: You're part of dxpedition. vy interesting to see which DXCCs, Grids or whatever you reached max.

This page closes that gap:

image


phl0 (approved):
Works fine

PR #3708, opened 2026-08-20
[Antenna Analytics] Can show azimuthal map instead of radar chart

AndreasK79:


phl0 (approved):
Works fine on my dev.

PR #3710, opened 2026-08-21
Counties map clean

int2001:
Replaces #3637 

tnx [N0RC](https://github.com/Reid-n0rc) for the main part of the contribution here.

For content see #3637 - i need to make new PR, since the old one had some restrictions i wasn't able to fix with best effort. sorry for that.


phl0 (approved):
Tested and works fine. Even with SAT QSOs :)

PR #3712, opened 2026-08-21
Label designer

AndreasK79:
As the QSL designer was added a few months ago, I've always thought that we could do the same with the labels.

A new button now exists on the label page:
image

You can select between the default or the designed template from all places where you print labels.

I think I have added all that is necessary, but I might have missed something. The new template only shows in print if you have created one.


int2001 (approved):
nice one.

PR #3715, opened 2026-08-22
[DBTools] Relocate from the advanced logbook to separate page

AndreasK79:
The PR moves DBtools from inside the Advanced Logbook to its own menu item, located under Tools -> Database Tools.

I have several reasons for doing it:
- It is kinda hidden away
- I want to add more things and change it up a bit
- Since the Advanced Logbook code base is already big, it is better to split it, and the DBTools aren't dependent on the Advanced Logbook.
- I didn't like the looks using a dialog, and a separate page is better to achieve a better result
- Perhaps get rid of some internal tools and merge them together with the functionality in the DBTools


int2001 (approved):
code looks good now.
regarding logic: tested what i could test. still dangerous function (for someone who don't know what he's doing - but theres a warning).

PR #3722, opened 2026-08-23
Imp- Export for labels from labeldesigner

int2001:
This one adds (hopefully bulletproof) export and import for custom created labels from the label-designer to share designs.

Export contains:
- paperdefinition
- labeldefinition
- Design

Import checks:
- if valid file
- if a similar paperdefinition already exists (takes them / otherwise creates)
- various checks


AndreasK79 (approved):

AndreasK79:
Works :)

PR #3725, opened 2026-08-24
Check LoTW-Response B4 saving file

int2001:
LoTW-Downloader produces orphaned files, if user provided wrong credentials, lotw has hiccups, etc.

this one adds try/finally to the parser, checks content before saving it to disk and unlinks possible orphaned files b4 wrting new ones.


HB9HIL (approved):
lgtm

phl0:
The check for wrong credentials only checked the first 2k of the file. But the string indicating wrong creds is around 6k into the file:
image
So we adjusted that limit to be 6k instead of 2k.

phl0 (approved):
Works fine here and does not leave orphaned files.

PR #3729, opened 2026-08-25
Align QSL-Method-View to standard

int2001:
This PR contains:

- [x] optimize lookups (from 8 SQLs down to 2)
- [x] show Confirmationtype as "Q,L,E,C,Z" like in awards for `ALT-L` and Summary at the end of QSO-Form

Info: Even if there are now some "Green Confirms" for secondary services like eQSL, Clublog, etc.: At the very end the logic how to derive if a QSO has been confirmed for logging, DXCluster & Co. hasn't changed. it still takes what you selected as "default methods" at Userconfig. But this PR helps - e.g. - to see if you're already in someones log at those secondary services if you selected QSL and LoTW only as default-service

Before:
image

Now:
image

Things to test:
compare carefully including details/QSO-View

### DIFF
--- a/application/models/Lookup_model.php
+++ b/application/models/Lookup_model.php
@@ -7,17 +7,18 @@ class Lookup_model extends CI_Model{
 		foreach ($queryinfo['sats'] as $sat) {
 			$resultArray[$sat] = '-';
 		}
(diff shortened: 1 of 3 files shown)


phl0 (approved):
Checked (possibly) all cases based on my current log. Could not spot an issue.

AndreasK79 (approved):

AndreasK79:
Good idea :)

PR #3731, opened 2026-08-25
Fix Error at WAP when entries are logged with different case

int2001:
Small bug in WAP-Award.
it fails with errors/warnings if one logs the province in lowercase or mixed-case.

this PR mitigates it.

Error/Warning shown (example)

```
--> Severity: Warning --> Undefined array key "nb" /var/www/html/application/models/Wap.php 43
--> Severity: Warning --> Undefined array key "count" /var/www/html/application/models/Wap.php 43
--> Severity: Warning --> Undefined array key "Ut" /var/www/html/application/models/Wap.php 43
--> Severity: Warning --> Undefined array key "count" /var/www/html/application/models/Wap.php 43
```

### DIFF
--- a/application/models/Wap.php
+++ b/application/models/Wap.php
@@ -195,7 +195,7 @@ class wap extends CI_Model {
 	 */
 	function getwapWorked($location_list, $band, $postdata) {
 		$bindings=[];
-		$sql = "SELECT distinct col_state FROM " . $this->config->item('table_name') . " thcv
+		$sql = "SELECT distinct UPPER(col_state) as col_state FROM " . $this->config->item('table_name') . " thcv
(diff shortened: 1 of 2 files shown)


phl0 (approved):

phl0:
Maybe also change at QSO view?

image

Not sure but I think COL_STATE is always upper case (or numbers) as per ADIF: https://adif.org.uk/317/ADIF_317.htm#Primary_Administrative_Subdivision

phl0:
> Maybe also change at QSO view?

Pushed in https://github.com/wavelog/wavelog/pull/3731/commits/ea32d0c1b2a81fb2e5dc80a974399d075688f7c0

phl0 (approved):

PR #3732, opened 2026-08-25
Add Applybutton to each filter at awards

int2001:
Follow Mainstream UX-Guidelines. Add (duped) Showbutton as "Apply" to each Filtermodal.

Doesn't hurt, but looks like people are searching for it.


phl0 (approved):
Well, if they need it ...

PR #3735, opened 2026-08-26
Refactor passing QSO-IDs as json instead of single-vals

int2001:
At QSL Label-printing we pass the IDs of the QSOs as JSON-Array, which is safe.
at Postcard, we do it as old'n'busted HTML-ID's, which lead into problems. See #3730 

this one refactors Postcardprint to the same technology as labelprint works.

old:
image

new:
image

Testing:
- Print Cards from LBA
- Print Cards from QSL-Menu


AndreasK79 (approved):

AndreasK79:
Still works :)

PR #3738, opened 2026-08-26
[Awards] Makeover for the filter dropdown

AndreasK79:
I don't like the looks of the dropdown. Here is a better version:

Before:
image

After:
image


int2001 (approved):
LGTM, but it is only DXCC - all others will be different after merge...

AndreasK79:
The rest is coming.

int2001 (approved):
modern view

phl0 (approved):
Works for me and looks nice. Great work.

PR #3739, opened 2026-08-26
Fixes a bug introduced in 832aabf77b

int2001:
an accidently introduced bug in https://github.com/wavelog/wavelog/commit/832aabf77b causes:
- recenter of logging-map to London instead of HomeQTH.

this one fixes it


phl0 (approved):
Fixes the bug. Tested with and without station grid. In the latter case center of DXCC is shown.

PR #3741, opened 2026-08-26
Use ham friendly SAT name instead of TLE object name

phl0:
This fixes a regression from https://github.com/wavelog/wavelog/pull/3358. Sats which have OMM TLE format were using their object name for display instead of the ham-friendly and commonly used names. Also the link to the tpx info data was broken.
Screenshot:

image

The example shows AO-10 as "PHASE" and RS-15 as "RADIO". Fix is to hand the ham-friendly name to Predict's TLE construction and use that as name rather than the object name. Same behaviour was implemented for TLE in two line format back in https://github.com/wavelog/wavelog/pull/2828.


int2001 (approved):
LGTM

PR #3742, opened 2026-08-27
apiv2 station listing observes user option active_log_only

kc9uhi:
ref: #3362 #3373

Update APIv2 station location listing to follow active_log_only user option. Aligns behavior with APIv1.


HB9HIL (approved):
@kc9uhi added two more commits. thanks for your contribution

PR #3743, opened 2026-08-27
Modulo Az when it comes to plotting by 360

int2001:
Bug:
When Useradif or Log provides Azimuth > 360, the Antenna-Analytics-Map stops rendering.
This fixes it by taking Az and normalizing it to a base of 0-360


phl0 (approved):
Fences to the user 🫣

PR #3744, opened 2026-08-27
Make Callbooklookup during QSO configurable

int2001:
Adds Useroption to skip Callbook lookup.
At the moment every QSO(-Partner) is looked up and the results are prefilled. Sometimes one don't want that (and it seems to be to complicated for those OPs to simply overwrite the prefilled values). However, this adds:

image

Default is still as before. Rest should be self-explaining.


phl0 (approved):
Tested and works as requested / described.

PR #3745, opened 2026-08-27
Make static map deal with no grid locations

phl0:
Static map fails to render if logbook contains locations without grid:

```
DEBUG - 2026-08-27 10:55:57 --> Cached static map image file does not exist. Creating a new one...
INFO - 2026-08-27 10:55:57 --> Model "Modes" initialized
INFO - 2026-08-27 10:55:57 --> Model "Logbook_model" initialized
ERROR - 2026-08-27 10:55:57 --> Severity: Warning --> Trying to access array offset on false /var/www/wavelog/application/libraries/Qra.php 197
ERROR - 2026-08-27 10:55:57 --> Severity: Warning --> Trying to access array offset on false /var/www/wavelog/application/libraries/Qra.php 198
```

### DIFF
--- a/application/libraries/Qra.php
+++ b/application/libraries/Qra.php
@@ -190,11 +190,11 @@ class Qra {
 	 */
 
 	function getCenterLatLng($coordinates) {
-		$x = $y = $z = 0;
-		$n = count($coordinates);
+		$x = $y = $z = $n = 0;
 		foreach ($coordinates as $point)
 		{
 			if (isset($point[0]) && isset($point[1])) {
+				$n++;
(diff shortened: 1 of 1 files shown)


application/libraries/Qra.php
```diff
@@ -194,11 +194,13 @@ function getCenterLatLng($coordinates) {
 		$n = count($coordinates);
```
int2001: since it skips now things and calculates means, wouldn't it be good to put the count inside the loop?

phl0: Thats a very good idea. Implemented.

int2001 (approved):

PR #3747, opened 2026-08-27
Maptile-Cache

int2001:
This one adds a maptile-cache to Wavelog.
instead of using the users browser to request the tiles from OSM/Maptileserver, Wavelog itself downloads the maptiles and caches them.
~~atm the feature is optional (per default disabled). if one wants to enable it, add a config-switch to your config.php:~~

~~`$config['map_tile_cache'] = true;`~~

todo's:
- [x] remove config-switch and default it to true
- [ ] test caching-adapters (from my PoV fine - shouldn't make a difference if one caches huge JSON or a maptile - both are accessed by key)
  - [x] REDIS successfully tested in prod. Cache-mem between 200 and 700MB
  - [ ] Filecache successfully tested in prod. (maybe @phl0 or @AndreasK79 ?)
- [x] extend the cache to make it usable for static-mapfiles
~~- [ ] check if we need some central config for a key at maptile-provider~~ key not needed.
- [x] TTL for Tiles should be 30d at least instead of 7d
- [x] Remove extra-provider for Dark-Tiles. Use normal ones instead and handle them like leaflet is doing it
- [x] put everything into a central lib

This one may(!) be an alternative to setup an own maptileserver, which is really much effort


kc9uhi:
I totally get this for a LAN environment with a limited internet connection and several local users. However in the inverse situation of a cloud-hosted wavelog server and multiple remote users, this simply increases wavelog server traffic load. Or am I wrong about this? I'd really appreciate the config option to disable, but I can add back in on my own.

HB9HIL:
> I totally get this for a LAN environment with a limited internet connection and several local users. However in the inverse situation of a cloud-hosted wavelog server and multiple remote users, this simply increases wavelog server traffic load. Or am I wrong about this? I'd really appreciate the config option to disable, but I can add back in on my own.

Hello @kc9uhi 
this PR has multiple reasons. Our latest experience is that map providers start to close down their public endpoints and require an API. To reduce load to the OSM provider we decided to start caching the maptiles. This does indeed increase the load to the webserver but not that much as you might think. The tiles are normally cached in the browser, just like at OSM aswell. So a user who opens the same page again with the same map, nothing gets loaded. The tiles themselve only add a few MB to the cache, thats not a big deal. And just returning a file from cache is a cheap operation for PHP. So no deal aswell.

HB9HIL (approved):

PR #3748, opened 2026-08-27
use volume size ratio for free space checks

HB9HIL:
when using quotas on storage volumes (mounted zfs, kubernetes pvc, and so on) setting 1Gb quota on userdata does not work as the check requires 1 Gb on free storage. It would be better to use a relative check and check for at least 10% free storage of the whole disk capacity.


int2001 (approved):
agree. e.g.: on 10GB HDD it stays 1GB :)

PR #3749, opened 2026-08-27
Check QRG vs. Band

int2001:
Bug: QSO form frequency/band desync on save

- QSOs were saved with COL_FREQ inconsistent with COL_BAND (e.g. band 13cm with a 10m frequency, or frequency off by ×1000). Root cause: five async writers.

Fix:
- assets/js/dxwaterfall.js — setFrequency() now sets the band select alongside the frequency (derived via frequencyToBand), instead of leaving the previous band in place. Out-of-range frequencies keep the current band.
- assets/js/sections/qso.js — band-change handler now fetches the band default frequency whenever the current frequency falls outside the newly selected band
- assets/js/sections/qrg_handler.js — set_new_qrg() parses bare numbers using the unit label captured at input focus, closing the race where a CAT/bandmap update flipped the unit label between typing and parsing (source of ×1000 frequency errors). Also guards against a null derived band.
- assets/js/cat.js — CAT band auto-update guards against null derived band, so out-of-band CAT frequencies no longer clear the band selection.

Frontend only, no backend/API changes. Contest engine untouched (already treats QRG as authoritative).

Testing
- Spot click on other band without CAT; 
- band flip with selected/stale CAT; 
- bandmap handoff
- kHz-style QRG entry during active CAT polling
- out-of-range frequencies (manual + CAT)
- live CAT band switching (no regressions/loops).

### DIFF
--- a/assets/js/cat.js
+++ b/assets/js/cat.js
@@ -1034,7 +1034,7 @@ $(document).ready(function() {
                     $frequency.trigger('change'); // Trigger for other event handlers
                     const newBand = frequencyToBand(d);
                     // Auto-update band based on frequency
-                    if ($band.val() != newBand) {
+                    if (newBand && $band.val() != newBand) {
                         $band.val(newBand).trigger('change'); // Trigger band change
                         // Update callsign status when band changes via CAT
                         if ($('#callsign').val().length >= 3) {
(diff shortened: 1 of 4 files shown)


phl0 (approved):

phl0:
Tested and works ok

PR #3752, opened 2026-08-28
[WAB Award] Fixed layout and enhanced list

AndreasK79:
The page layout didn't match the other pages that well, so that is fixed.
Added a WAB search, which zoom in to the square.
Added coordinates to the bottom of the map, to find gridsquare and which WAB it is in.
Added more information to the WAB list.


phl0 (approved):

int2001 (approved):
huge one, but looks good to me

PR #3753, opened 2026-08-28
Refactor file adapter

HB9HIL:
> [!IMPORTANT]
> This PR is based on #3747 

The file adapter in the Codeigniter Core was pretty outdated as it was designed more than 10 years ago. So with the new Maptile Cache which produces a lot of files I refactored the file adapter

> [!NOTE]
> Normally it would be better to add a new file adapter as I did for example with redis2. In this case it's not as we don't want users to use the old file adapter as this could break their systems. An other option would be to use the MY_ prefix. But we don't have control of the config.php of the users and don't know if someone already implemented their own files with another prefix.

## What changed

- The file adapter now uses sharding by mapping the keys by their hash. This prevents one huge directory. (xx/yy/hash). The path is build from the xxh128 hash (builtin in php so no dependency and faster than md5).
- save() does not directly write to the cache. It writes to tmp and renames then. This makes sure values are never read half filled. 
- More defensive error handling. The @ operator on unlink(),file_get_contents() and unserialize() supresses warnings on failed commands. separatly the _get() function now deletes cache files whose content fails to unserialize into a valid array with time/ttl — cleaning up invalid or corrupt cache entries on read
- Updated garbage collector so it follows the shard subdirectories

See the diff between this PR and the new maptile cache: [diff](https://github.com/int2001/wavelog/compare/maptile_cache...HB9HIL:wavelog:refactor_file_adapter)


HB9HIL:
was merged into #3747

PR #3755, opened 2026-08-28
Use band (xuser-table) instead of frequency for DXCC at Dashboard

int2001:
Bug:

Dashboard used Frequency to derive HF (and more).
but ARRL uses Band. and our DXCC-Award page also uses Band.

so if anyone has - e.g. - 144.250 MHz but 20m as Band, the results were wrong.

Visitorpage is unaffected by this PR. tbh: unsure how to deal with it.


phl0 (approved):

phl0:
No issues (same numbers as before) :)

PR #3757, opened 2026-08-28
[DBTools] Rewrite how callbook gridsquare lookup works

AndreasK79:
A refresh of the gridsquare updater in the dbtools:
image


int2001 (approved):
can't spot security or logical issues.

only thing to keep in mind: if user updates a lot of QSOs with that function, a lot of requests to callbook-provider are happening.

for now ok, since user can cancel it

PR #3758, opened 2026-08-28
Make legend align to light themes

phl0:
Before:

image

After:

image


int2001 (approved):
can't spot an issue here

PR #3763, opened 2026-08-30
Show Toast when 3rd party fails at RT-logging

int2001:
Problem:
A lot of Users won't notice if they forget to configure - e.g. - Clublog or QRZ for the active Call.

Solution:
Show a toast to the User, if a Realtime-3rd-Party upload fails.
Includes Reason (Sanitized and truncated)


phl0 (approved):
Tested (with forged error messages) for qrz and clublog. No issues.

PR #3764, opened 2026-08-30
Fix bug, where propmode/sat was cleared on manual qrg-change

int2001:
Problem: Commit [021e24df](https://github.com/wavelog/wavelog/commit/021e24df) added .trigger('change') on band set after freq entry → band-change handler wiped propagation mode, sat name, sat mode, RX freq, antenna az/el.

Fixes:
- Sat/prop/RX wipe only on user-initiated change (e.originalEvent)
- Manual freq entry + CAT + bandmap: sat data preserved
- stop_az_ele_ticker() only when sat was active
- Non-sat fav now stops stale ticker from previous sat

Summary: sat/prop info survives now freq changes; ticker lifecycle tied to sat presence.

### DIFF
--- a/assets/js/sections/qso.js
+++ b/assets/js/sections/qso.js
@@ -726,10 +726,7 @@ $(document).on("click", "#fav_del", function (event) {
 });
 
 $(document).on("click", "#fav_recall", function (event) {
-	$('#sat_name').val(favs[this.innerText].sat_name);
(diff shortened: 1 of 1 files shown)


phl0 (approved):
Yeah. Works. Prevents loosing prop mode (and SAT details) on freq update

PR #3773, opened 2026-09-02
Fix handling of no coordinates for DXCC none

phl0:
Since we derive mapping coordinates for QSOs without grid from its DXCC and -NONE- as DXCC had 0/0 as coordinates we have some issues mapping those QSOs:

Though we have a grid the map does not center on grid and does not show a marker (if DXCC is -NONE-)

image

So we change structure of dxcc_entitites to allow null for lat/long and set it to null for DXCC -NONE- with id 0. We show maps centered and markers if we have a grid or coordinates (and not null as for DXCC -NONE- without grid)

### DIFF
--- a/application/migrations/298_dxcc_none_null_coordinates.php
+++ b/application/migrations/298_dxcc_none_null_coordinates.php
@@ -20,9 +20,9 @@ class Migration_dxcc_none_null_coordinates extends CI_Migration {
 
 	public function down()
 	{
+		$this->dbtry("UPDATE {$this->table_name} SET `long` = 0, `lat` = 0 WHERE {$this->table_name}.`adif` = 0;");
 		$this->dbtry("ALTER TABLE {$this->table_name} CHANGE `long` `long` FLOAT NOT NULL;");
 		$this->dbtry("ALTER TABLE {$this->table_name} CHANGE `lat` `lat` FLOAT NOT NULL;");
-		$this->dbtry("UPDATE {$this->table_name} SET `long` = 0, `lat` = 0 WHERE {$this->table_name}.`adif` = 0;");
 	}
 
 	function dbtry($what) {
--- a/application/migrations/298_dxcc_none_null_coordinates.php
+++ b/application/migrations/298_dxcc_none_null_coordinates.php
@@ -1,0 +1,35 @@
+<?php
+
+defined('BASEPATH') OR exit('No direct script access allowed');
+
+/*
+ * Lat/lon of 0/0 are valid coordinates though not for a DXCC.
+ * So we allow null as value here to allow differentiation
+ */
+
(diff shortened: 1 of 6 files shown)


application/views/view_log/qso.php
int2001: since both can be `null` now, this can produce at least warnings.
better:
```suggestion
var lat = <?php echo ($lat ?? 'null'); ?>;
var long = <?php echo ($lng ?? 'null'); ?>;
```

application/migrations/298_dxcc_none_null_coordinates.php
```diff
@@ -0,0 +1,35 @@
...
+		$this->dbtry("ALTER TABLE {$this->table_name} CHANGE `long` `long` FLOAT NULL;");
+		$this->dbtry("ALTER TABLE {$this->table_name} CHANGE `lat` `lat` FLOAT NULL;");
+		$this->dbtry("UPDATE {$this->table_name} SET `long` = NULL, `lat` = NULL WHERE {$this->table_name}.`adif` = 0;");
+	}
+
+	public function down()
+	{
+		$this->dbtry("ALTER TABLE {$this->table_name} CHANGE `long` `long` FLOAT NOT NULL;");
+		$this->dbtry("ALTER TABLE {$this->table_name} CHANGE `lat` `lat` FLOAT NOT NULL;");
+		$this->dbtry("UPDATE {$this->table_name} SET `long` = 0, `lat` = 0 WHERE {$this->table_name}.`adif` = 0;");
```
int2001: different order?
first set to `0` then alter table?

int2001 (changes requested):
see comments.
Furthermore:

since it can be `null` now, check Logbook_model.php function get_map_qsos loop. looks like the last real calculated loc isn't resetted, when one with `null` came around during iteration

application/views/view_log/qso.php
phl0: Solved by https://github.com/wavelog/wavelog/pull/3773/commits/f14fd276b3ca38dd120aa238f7188314dc128ef2.

application/migrations/298_dxcc_none_null_coordinates.php
phl0: Solved with https://github.com/wavelog/wavelog/pull/3773/commits/7247ae523d6f22d8f83af08e6b6da2e28af3718f.

int2001 (approved):

phl0:
> since it can be `null` now, check Logbook_model.php function get_map_qsos loop. looks like the last real calculated loc isn't resetted, when one with `null` came around during iteration

Fixed by https://github.com/wavelog/wavelog/pull/3773/commits/0637d2715a1eb2737127f22d71bf0d4dcd9ef7d6

PR #3776, opened 2026-09-02
Trophies fix

int2001:
#3772 had some issues like:
- double-encoding of text (leads to e.g.: `B&auml;nder`)
- Caching of rendered strings instead of values.

This PR fixes them


phl0 (approved):

phl0:
This PR merge was reverted in https://github.com/wavelog/wavelog/commit/725a00ea4902e6c67d788339e03183747ce3682b because it breaks HTML entities in other places.

PR #3777, opened 2026-09-02
RM Double encoding / cache values, not strings

int2001:
https://github.com/wavelog/wavelog/pull/3772 had some issues like:

double-encoding of text (leads to e.g.: B&auml;nder)
Caching of rendered strings instead of values.
This PR fixes them


phl0 (approved):
Tutet.

PR #3785, opened 2026-09-04
Refactored WAIP to safer binding

int2001:
Found minor QSL-Vulns. At the moment no real attack-vector (since data comes from DB).
but: since we use bindings whereever possible, why not at:
- Polska Award
- Italy Award
- query_db for partial log


HB9HIL (approved):

phl0 (approved):
Works for me

Issue #3789, opened 2026-09-04
DXCC needed is off by 10


zcrowman:
### I have checked the following

- [x] I am running the latest version of Wavelog
- [x] I have cleared my browser cache
- [x] This is a bug, not a setup question or feature request

### Bug Description

On the dashboard, right sidebar under DXCC's breakdown, the amount needed is too high by 10. I currently have 259 worked and would expect to see 81 needed but it is showing 91 needed. 

### Steps to Reproduce

1. Go to dashboard
2. Go to DXCCs Breakdown
3. compare DXCCs worked to DXCCs remaining out of the 340 total. 

### Expected Behavior

Should show 340 - DXCCs worked but is showing 10 additional DXCCs. 

### Wavelog Version

3.2.0

### PHP Version

8.3.30

### MySQL/MariaDB Version

8.0.41-0ubuntu0.24.04.1

### Server OS

Apache

### Browser

Firefox

### Device Type

Desktop

### Screenshots / Browser Console

<img width="1174" height="424" alt="Image" src="https://github.com/user-attachments/assets/9a02cdeb-fca3-4887-b91c-17db6d8585ea" />

phl0:
This will be hard to debug .... Would you be able to provide an ADIF of your log which I could use to investigate?

phl0:
Or can you send some screenshots of DXCC (map) one with confirmed by card only and another one with LoTW confirmations only?

zcrowman:
I just noticed that on the map, showing worked, not worked and confirmed by QSL and LOTW that the unworked DXCC is 81, which is what I am expecting to see. Seems that the incorrect number shows only on the dashboard.  I've attached my adif file. 

[VE7ZCZ-20260904-1756.adi.zip](https://github.com/user-attachments/files/31845943/VE7ZCZ-20260904-1756.adi.zip)

phl0:
The numbers are correct. Needed only counts the DXCCs which are not confirmed (by either card or LoTW). In your case you have 244 confirmations on LoTW and 113 on paper cards. Total sum of confirmed DXCCs is 249 (not shown on the dashboard) and thus 91 are remaining to be confirmed.

zcrowman:
Ah. I see. I find that distinction confusing. Might be good to define this a bit more clearly. So if I have this correct, on the dashboard DXCC, the "needed" means, the confirmations needed to complete 340, but on the DXCC map it is the number of DXCC still not worked. 

[...]

PR #3790, opened 2026-09-04
Refactoring DXCC numbers in dashboard

phl0:
Adresses https://github.com/wavelog/wavelog/issues/3789 and similar questions. Draft here for now as the layout / presentation is still to be negotiated (but the separate numbers are there).

Current state looks like this:

Screenshot from 2026-09-04 21-24-17


int2001 (approved):
LGTM. can't spot an issue now

Issue #3791, opened 2026-09-04
Callbook lookup using qrz: Username / password required


mkosmo:
### I have checked the following

- [x] I am running the latest version of Wavelog
- [x] I have cleared my browser cache
- [x] This is a bug, not a setup question or feature request

### Bug Description

When trying to use QRZ for callsign lookup, I'm getting an error that username and password are required.  The only special char in use is a hyphen, and I've confirmed that the credentials work at https://xml.qrz.com/xcheck (per a previous issue)

Note;  I added the additional debug statements in the screenshot to verify that the username and password were correctly loading.  They are correctly loaded to that point.

### Steps to Reproduce

1. Configure QRZ Callbook lookup
2. Look up callsign

### Expected Behavior

Retrieve callsign information.

### Wavelog Version

3.2.0

### PHP Version

8.3

### MySQL/MariaDB Version

10.11

### Server OS

Debian 12

### Browser

Firefox

### Device Type

Desktop

### Screenshots / Browser Console

<img width="701" height="269" alt="Image" src="https://github.com/user-attachments/assets/c0481b6f-9316-49bc-9449-548003ccf55b" />

<img width="936" height="78" alt="Image" src="https://github.com/user-attachments/assets/31d0f71e-d009-4343-b72e-33221a5b3570" />

My additional debug lines as they are:

<img width="845" height="191" alt="Image" src="https://github.com/user-attachments/assets/47c267d0-eeff-406d-b99f-51a3beaa3e32" />

int2001:
Sorry, i can't follow you.
Whats exactly the problem? 
- Specialchars in password, which are rejected / not properly passed?
- QRZ-Lookup in general not working? (Cannot reproduce it - working here and on many other instances)

mkosmo:
> Sorry, i can't follow you. Whats exactly the problem?
> 
>     * Specialchars in password, which are rejected / not properly passed?
> 
>     * QRZ-Lookup in general not working? (Cannot reproduce it - working here and on many other instances)

I only mentioned the hyphens to try to exclude that as an issue.  It's simply not working for me, and the debugging doesn't lend itself to easily understanding why it's not working for me.  I fully understand that it's working for many others and is specific to me, but I can use QRZ from a PC on the same network and access the XML feed from other logbooks.

int2001:
did you ever try to change the password at qrz?

mkosmo:
> did you ever try to change the password at qrz?

[...]

PR #3793, opened 2026-09-05
Use urlencoded credentials for qrz instead of raw-values

int2001:
Should fix possible problems with specialcharacters in QRZ-Credentials like mentioned in #3791 

Testing:
- QRZ Lookup should work as before
- Alter QRZ-Password to something weird, test again

### DIFF
--- a/application/libraries/Qrz.php
+++ b/application/libraries/Qrz.php
@@ -12,7 +12,7 @@ class Qrz {
 	public function session($username, $password) {
 		// URL to the XML Source
 		$ci = & get_instance();
-		$xml_feed_url = 'https://xmldata.qrz.com/xml/current/?username='.$username.';password='.urlencode($password).';agent=wavelog';
+		$xml_feed_url = 'https://xmldata.qrz.com/xml/current/?username='.urlencode($username).';password='.urlencode($password).';agent=wavelog';
 
 		// CURL Functions
 		$ch = curl_init();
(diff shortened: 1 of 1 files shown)


phl0:
@mkosmo Could you please outline which special chars were part of the password so I can test this patch?

phl0 (approved):
Ok, I see this patch has nothing to do with the password. Not feasible to change my username (and tbh I don't know why/how someone would have a non-printable/special char in username/callsign). But anyway this code (still works). So formal approval here.

phl0:
> @mkosmo Could you please outline which special chars were part of the password so I can test this patch?

Forget about this. This patch wouldn't have helped. I changed my password to include special chars like ";" and "/" but it still works. qrz.com does not list password requirements so we don't know which one would fail.

mkosmo:
> > @mkosmo Could you please outline which special chars were part of the password so I can test this patch?
> 
> Forget about this. This patch wouldn't have helped. I changed my password to include special chars like ";" and "/" but it still works. qrz.com does not list password requirements so we don't know which one would fail.

Even if you did, my current QRZ password only has a special char of a hypen (-).  I wish I was doing something that made the error make sense.

int2001:
> Ok, I see this patch has nothing to do with the password. Not feasible to change my username (and tbh I don't know why/how someone would have a non-printable/special char in username/callsign). But anyway this code (still works). So formal approval here.

It was in pre-advance for "special usernames". Tnx

PR #3795, opened 2026-09-05
Round degrees for visualization

int2001:
Some tools calculate Azimuth/Elevation with very high precision.
This PR rounds the values when it comes to display (LBA)


phl0 (approved):
Tested and works.

PR #3796, opened 2026-09-05
Added a fallback to DXCC reference coordinates for bearing when logging

Patrik-OM1EPT:
The live QSO-logging callsign lookup only computed a bearing/azimuth and distance when the callbook returned an actual gridsquare for the worked station. Most operators at our club know roughly where they should aim the antennas, but this should help in more precise beaming.

Added a fallback in Logbook::partial() lookup, when no gridsquare is available but the resolved DXCC entity has reference coordinates, use those to compute the bearing, antenna azimuth prefill, and distance, same as the grid-based path.


HB9HIL:
PR's always against `dev` branch please

Lopastudio:
Sorry, I'm pretty new to pushing PR's, I'll make a new one into the dev branch. Thanks for notifying me.
DE OM1EPT

PR #3797, opened 2026-09-05
Added a fallback to DXCC reference coordinates for bearing when logging

Patrik-OM1EPT:
The live QSO-logging callsign lookup only computed a bearing/azimuth and distance when the callbook returned an actual gridsquare for the worked station. Most operators at our club know roughly where they should aim the antennas, but this should help in more precise beaming.

Added a fallback in Logbook::partial() lookup, when no gridsquare is available but the resolved DXCC entity has reference coordinates, use those to compute the bearing, antenna azimuth prefill, and distance, same as the grid-based path.


int2001 (approved):
good one.
tested:
- live logging
- post logging
- logging while SAT is active
- looking up call without grid (for each case)
- looking up call without DXCC (e.g. `/MM`)

Pytlicek:
Great! many many thanks. I was missing this feature a lot 🙂

Issue #3781, opened 2026-09-03
US Counties award map does not actually show Connecticut counties because of an odd quirk of CT local government


kc3zyt:
### I have checked the following

- [x] I am running the latest version of Wavelog
- [x] I have cleared my browser cache
- [x] This is a bug, not a setup question or feature request

### Bug Description

Connecticut is a weird state. I'm not fully prepared to explain the background, as I'm not a resident and I wasn't aware of any of this until 20 minutes ago. My basic understanding is that [Connecticut has counties](https://en.wikipedia.org/wiki/List_of_counties_in_Connecticut), but those counties basically don't do anything anymore and local government is handled by [Councils of Government](https://en.wikipedia.org/wiki/Councils_of_governments_in_Connecticut), which have different boundaries.

The US counties table menu uses the counties. However, the US counties map shows the borders of the councils of government. 

As far as I am aware, every Connecticut contact I've ever had confirmed on LOTW, eQSL, or QRZ has listed their county, not their council of government.

Thus, the US counties map is completely red for me because (I assume) the map is expecting the council names.

### Steps to Reproduce

1. Hover over "Awards"
2. Hover over "USA"
3. Click "US Counties"
4. Click the "map" tab and look at Connecticut

### Expected Behavior

As every logging platform still seems to use Connecticut's counties, then the map should also use the counties instead of the councils of government.

### Wavelog Version

3.2.0

### PHP Version

8.5.10

### MySQL/MariaDB Version

11.3.2

### Server OS

TrueNAS 25.10.4

### Browser

Firefox

### Device Type

Desktop

### Screenshots / Browser Console

<img width="783" height="459" alt="Image" src="https://github.com/user-attachments/assets/e4da93d3-ff02-4b40-9d0b-1c1673fb95bd" />

<img width="1264" height="158" alt="Image" src="https://github.com/user-attachments/assets/deb840bc-18c9-459e-990e-53be25ff29f8" />

int2001:
see #3782 - imho it would be good to pin this issue to that one if you don't mind

Reid-n0rc:
@kc3zyt you're correct. The map needs to be updated with CT counties and not councils of government. I will try to look into this and hopefully not break everything in the process.

Reid-n0rc:
@int2001 

[...]

Issue #3798, opened 2026-09-05
USA-CA award: let users pick an adjoining county to credit unmatched independent-city QSOs


Reid-n0rc:
Follow-up to #3782.

Per MARAC's USA-CA award rules (rule C.5, https://marac.org/documents/usa-ca-rules.pdf): "In the case of independent cities or federal parks and reservations not located within a county, the applicant may claim any one of the adjoining counties for credit."

#3782 excludes these from the award's Worked/Confirmed/Target counts (they're not one of MARAC's 3,077 target counties - see `assets/json/US_COUNTIES_SOURCE.md`) and instead surfaces them under a new "Unmatched" column, since there's no single correct county to credit automatically - the rule leaves it up to the operator, and no canonical city-to-county mapping exists.

This covers: Virginia's independent cities, Carson City NV, and Washington DC's one row (not itself an independent city, but in the same "not part of any county" situation and not in MARAC's list either).

Idea for closing the gap: let the user manually assign one of the state's adjoining counties to an unmatched QSO (or a batch of them sharing the same logged county value), so it then counts toward that county's Worked/Confirmed total - matching what the rule actually allows, rather than leaving these permanently unscored.

Needs some design thought before implementation:
- Where the assignment lives - per-QSO override, or a reusable "credit X to county Y" mapping the user sets once per unmatched value?
- Which counties count as "adjoining" for a given independent city, for the picker's options - no published canonical list exists for this (checked during #3782's investigation).

Reid-n0rc:
The scoring/unmatched-surfacing work this depends on is now published: wavelog/wavelog#3799.

Reid-n0rc:
I cannot complete this without the use of AI. Someone else needs to pick this up or it needs to be closed.

Issue #3782, opened 2026-09-03
US Counties award target/progress totals don't match the actual USA-CA award (3,077 counties, no independent cities, AK judicial districts) — map display should stay as-is


Reid-n0rc:
The US Counties award (Awards > USA > US Counties) currently tracks progress against `assets/json/US_counties.csv`, which contains 3,149 entries — effectively the full Census Bureau county/county-equivalent list. This doesn't match the actual **USA Counties Award (USA-CA)**, which is defined against **3,077 counties**:

- **Independent cities are not separate targets.** Virginia's independent cities (e.g. Alexandria City, Bristol City, Bedford City) and similar entries elsewhere (Baltimore City MD, St. Louis City MO, Carson City NV) are listed as their own rows in `US_counties.csv`, but the award folds them into the county they're geographically part of / associated with rather than counting them separately.
- **Alaska counts as 4 judicial districts, not individual boroughs/census areas.** `US_counties.csv` lists all ~29 Alaska boroughs and census areas individually; the award instead uses the pre-2000s Alaska Judicial Districts grouping (1st–4th District).

Reference for the correct 3,077-entry list and Alaska grouping: https://www.qsl.net/kl7j/county.html

**What should stay the same — the map:**

The counties map should keep filling in at its current granularity — every individual county/independent city/Alaska borough or census area worked should still show on the map exactly as it does today. This is a display concern, not a scoring concern, and the existing 3,149-entry list is the right one for that. The map's per-county drill-down and QSO detail dialogs shouldn't change.

**What should change — target and progress totals:**

The numeric "Worked / Confirmed / Target" figures (the award page's summary counts, and `Counties.php::get_counties_progress()` which drives them) should be computed against the 3,077-entry USA-CA rule set, not the raw 3,149-entry list:
- An independent city worked/confirmed should credit its associated parent county for award-progress purposes (without removing the independent city as its own point on the map).
- Working any of an Alaska judicial district's underlying boroughs/census areas should credit that judicial district once toward the target (map still shows the individual borough worked).

**Suggested fix:**
[...]

PR #3799, opened 2026-09-05
Fix USA-CA county scoring: CT map, Alaska JD grouping, independent cities

Reid-n0rc:
## Summary

Fixes wavelog/wavelog#3781 and wavelog/wavelog#3782, and implements wavelog/wavelog#3798.

- **Connecticut counties map** (#3781): `counties_291.geojson`'s CT features were sourced from the 2022 Census vintage, which already reflects CT's 2022 switch from 8 counties to 9 "planning regions" (same boundaries as CT's Councils of Government). `US_counties.csv` (and all real QSO log data - LOTW/QRZ/eQSL/ARRL/MARAC) still uses CT's 8 traditional counties, so none of the 9 polygons could ever match a logged county and CT rendered permanently unworked. Fixed by sourcing CT's 8 counties from the pre-2022 `cb_2021_us_county_500k` vintage instead (scoped to CT only), and regenerating CT's dissolved outline to match. See `assets/json/geojson/COUNTIES_SOURCE.md` for the full writeup.

- **USA-CA scoring** (#3782): `Counties.php::get_counties_progress()` counted worked/confirmed/target against all 3,149 rows of `US_counties.csv` - the raw Census county/county-equivalent list. That overcounts against the actual USA Counties Award (USA-CA), which targets exactly **3,077** counties per MARAC's authoritative list ([county-abbrev-v4.doc](https://marac.org/pages/tools/county-abbrev-v4.doc)):
  - Alaska's boroughs/census areas collapse into 4 Judicial Districts, not counted individually.
  - Virginia's independent cities, Washington DC, and Carson City NV (Nevada's consolidated city-county) aren't in the 3,077-county list at all. Per the [award rules](https://marac.org/documents/usa-ca-rules.pdf), rule C.5, these may be credited to any adjoining county at the operator's choice - not a single fixed mapping.

  Added a `scoring_group` column to `US_counties.csv` (documented in the new `assets/json/US_COUNTIES_SOURCE.md`) naming the actual USA-CA target each row counts toward, or `UNSCORED` for the excluded rows above. `Counties.php`'s progress/target/needed logic now dedupes by `scoring_group`. Logged `UNSCORED` counties are no longer silently dropped - a new "Unmatched" column/click-through surfaces them. Alaska's judicial districts render as a labeled group with their member boroughs listed underneath in every dialog that lists counties for a state, rather than a bare district name with no context.

  The map and per-county QSO-detail views are untouched - they still show/link every individual county at full granularity; only the summary Worked/Confirmed/Target/Remaining counts and their click-through lists change.

- **Adjoining-county credit** (#3798): rule C.5 lets the operator credit an independent city/DC/Carson City NV QSO to any one of its real adjoining counties, so a user can now do that instead of the entity staying permanently unmatched. `assets/json/US_counties_adjoining.json` lists the real, geometrically-adjoining counties for each of the 41 `UNSCORED` rows, computed from the same polygon data as the counties map (not hand-picked) - full methodology in `US_COUNTIES_SOURCE.md`. The Unmatched dialog gets a per-row dropdown of those adjoining counties plus a Save button; the assignment is stored per-user in the existing `user_options` table (no migration needed) and is picked up by the Worked/Confirmed/Target/Needed counts and their click-through dialogs, crediting the correct *target* state even when it differs from the entity's own state (DC's neighbors span MD and VA).

  **This adds a new AJAX endpoint, `Awards::counties_assign_ajax()`.** Per the contributing guide ("Discuss new API endpoints before implementing them"), flagging it here explicitly for discussion before merge - it follows the same internal-AJAX-handler pattern as this file's existing `counties_map`/`counties_list_ajax`/`counties_details_ajax` endpoints (session-authenticated, POST-only, no new route needed), but it's new surface area and I want a maintainer's sign-off on the approach rather than assuming the existing pattern is sufficient cover.

## Sources

- 3,077-county target list: https://marac.org/pages/tools/county-abbrev-v4.doc
- Award rules (rule C.5): https://marac.org/documents/usa-ca-rules.pdf
- Alaska judicial district borough mapping: https://www.qsl.net/kl7j/county.html

## Test plan

Verified in a local Docker + MySQL instance built from this branch's `Dockerfile`, seeded with test QSOs covering:
- [x] All 8 Connecticut counties render and color correctly on the counties map
- [x] Two Alaska boroughs in the same judicial district (Juneau, Sitka - both First JD (SE)) collapse into one worked target instead of two; a third (Nome, Second JD (NW)) and a legacy pre-2019 name (Valdez-Cordova, mapping to Third JD (SC)) count correctly
- [x] Virginia's independent cities (including `Bristol`/`Salem`, whose bare CSV name has no "City" suffix) are excluded from VA's worked/confirmed/target and appear under Unmatched instead
- [x] Washington DC and Carson City, NV drop to a 0/0/0 target with 1 unmatched each
- [x] The counties map endpoint still returns individual boroughs/cities unchanged
- [x] Alaska's judicial districts list their member boroughs in the full per-county table and every worked/confirmed/target/needed/unmatched dialog
- [x] Ordinary states (CT, TX, etc.) are unaffected - no grouping, same counts as before
- [x] Assigning an independent city to an adjoining county (same-state, e.g. VA's Alexandria City -> Arlington, and cross-state, e.g. DC -> Montgomery, MD) moves it out of Unmatched and into the target's Worked/Confirmed count
- [x] The Needed dialog and the main table's Remaining count agree after a credit is applied
- [x] Clicking a credited county's Worked/Confirmed count shows the actual QSO (resolves via the credited entity's own real state/county, not the target's)
- [x] A non-adjoining target is rejected server-side
- [x] The Unmatched dialog stays reachable (link doesn't disappear) after a credit drops its state's unmatched count to 0, so an existing credit can still be reviewed or changed
- [x] The four VA independent cities with no real adjoining county (Chesapeake, Norfolk, Portsmouth, Virginia Beach - they only border other independent cities) show no assignment option

`po`/`mo` translation regeneration is intentionally not included here - opened as a separate PR per the project's convention of not mixing translation regen with feature work.


int2001:
Hey, @Reid-n0rc,
I'm really sorry—but I can't review your PR (which originally started as a simple issue to add a fourth column to the CSV file).

Reasons:
- It was generated entirely by AI. See our rules for PRs here: https://docs.wavelog.org/project/contributing/# code-guidelines (especially the last point regarding AI)
– The coding style isn’t followed.
– Unnecessary files are being added to Wavelog.
– This is a hobby project.

You should know:
We review every PR manually (of course, we also use AI for individual/simple, isolated tasks), just as we create our own PRs. It was already a major effort to clean up https://github.com/wavelog/wavelog/pull/3637 and get it ready for production. But this, in turn, is half of an (AI-generated) restructuring of the U.S. county content, again.

I’m truly sorry, but I hope you can understand.

Reid-n0rc:
I understand @int2001, unfortunately this is the limit of my ability. I'll close the PR.

PR #3804, opened 2026-09-07
Fix map when showing details for distancesworked

int2001:


phl0 (approved):
Looks good by reading the code

Issue #3803, opened 2026-09-07
Radio API does not allow decimal power levels


mattmelling:
### I have checked the following

- [x] I am running the latest version of Wavelog
- [x] I have cleared my browser cache
- [x] This is a bug, not a setup question or feature request

### Bug Description

In versions prior to 3.2.0 the Radio API would accept `float` power levels. For instance, my IC7300 when interrogated with `l RFPOWER` with the power level set to 50% will return `0.501961`.

In [WaveLogGoat](https://github.com/johnsonm/WaveLogGoat) we scale that value up to Watts and send it to update the rig info (see example request below in Steps to Reproduce). 

In 3.2.0 radio update requests with float power levels now fail with an error 400. [Cat.php L48](https://github.com/wavelog/wavelog/blob/824a7c2be815988799610abb2d0661e889ba119d/application/models/Cat.php#L48) specifically enforces that numeric values are ints, which makes sense for a lot of things but not power levels.

Other values such as 0.5W, 0.1W etc are similarly not be accepted.

### Steps to Reproduce

Post the following payload to `/api/radio` or `/api/v2/radio`:

```json
{
    "radio": "IC7300",
    "power": 50.1961,
    "frequency":7033000,
    "mode":"CW"
}
```

The response will have a 400 status and contain:

```json
{"error":{"code":"validation_error","message":"Invalid field(s): power","details":{"invalid":["power"]}}}
```

### Expected Behavior

Float power levels should be accepted. E.g. values such as 0.5W should work.

### Wavelog Version

3.2.0

### PHP Version

8.5.1

### MySQL/MariaDB Version

12.1.2-MariaDB-ubu2404

### Server OS

Debian 13

### Browser

Firefox

### Device Type

None

int2001:
Thats correct. Power at station_profile is defined as integer, same as at api/cat.
always has been.

And a value like `0.5W` is strictly(!) wrong. We can discuss about `0.5` (without the `W`) if its really that important.
ADIF says very clear: Number, without unit

int2001:
working on it at #3805 
the PR changes WL behaviour to accept float-values like defined at ADIF. Still without unit(!)

mattmelling:
Heh, thanks. I blame muscle memory for the addition of the `W` - I'm strictly talking about numeric values.

PR looks good I'll pull it into my working copy and try it out.

PR #3805, opened 2026-09-07
Fix Power to double

int2001:
Power allows float-values according to ADIF.
At Logtable the Field was already float.

But it was forced to be int at several places. This PR fixes it at:

- legacy API
- APIv2
- QSO-Edit / Add
- Station-Profile edit / add

Thanks for headsup at #3803


mattmelling:
Power is now a double and saved in the database correctly. The correct power value is pulled in when a rig is selected in the QSO create screen.

I ran a request to set the power for a dummy rig as `12.3456789`, however this QSO can not be saved as there is some validation on the `Transmit Power (W)` screen which only allows values up to 3 decimal places.

Personally I think 2 or 3 d.p. is enough and I verified that when limiting the precision at the source (Radio API sender) to 2 d.p. the power value is accepted without issue.

However this creates potential for a conflict between what WL inserts in the field and what it accepts so it might be worth truncating at 3 d.p. when the power field is populated from the current cat values.

int2001:
Should i be honest?
PWR field isn't interesting in the same way Antenna isn't. What counts are confirmations.

however: it's up to the user if he wants to fill no decimals or 3 or 20.

But yes: even if uninteresting for me, it should be safe. Going to add a guard for it

johnsonm:
I'm going to push back against your classification. Power is **very** interesting to people who are trying to see how little power they can get by with, who are exactly those who are affected by limiting to ints instead of allowing floats as allowed in the spec. You literally can't track your QRPp operations without this, so for anyone wanting to work QRPp, this bug would be disqualifying for being the log of record.

mattmelling:
Tested latest change locally and seeing `Transmit Power (W)` at 3 d.p. as expected. Thanks @int2001!

HB9HIL (approved):

PR #3806, opened 2026-09-08
[Search] add one-click option to search for all QSOs with specific DXCC

DB4SCW:
This one is a PR that is mostly a quality-of-life change that would improve my personal workflow. But maybe it's useful for others too. Feel free to say no if a feature like this is not wanted.

I relatively often find myself in the qso view wanting to know "what other QSOs have I had with this DXCC?" to check for used modes and confirmation status. Of course I can go into LBA and use the filters there.

This one makes the DXCC Flag clickable, using the same mechanism that the "More QSOs" button presents for searching for more QSOs with only that specific callsign.

At the moment only the DXCC flag is clickable, I decided against a dedicated "more from this DXCC"-Button for the first draft.

### DIFF
--- a/application/views/interface_assets/footer.php
+++ b/application/views/interface_assets/footer.php
@@ -1017,12 +1017,28 @@ function findlotwunconfirmed(){
     });
 }
 
-function searchButtonPress() {
+function searchButtonPress(searchDxcc) {
     if (event) { event.preventDefault(); }
     if ($('#callsign').val()) {
 		$('#btn-lba').removeAttr('hidden');
         let fixedcall = $('#callsign').val().trim();
         $('#partial_view').load("logbook/search_result/" + fixedcall, function() {
+            $('[data-bs-toggle="tooltip"]').tooltip();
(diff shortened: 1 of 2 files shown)


int2001 (approved):
i had to search a bit to find the function you added.
it's in log-view/partial log.

since you wrote "during QSO" (or "in the QSO") my 1st thought was:
Did he ever scrolled down in QSO-Form? There's a Matrix for DXCC, POTA, States and even DOKs, where each Slot it clickable.

However: Doesn't hurt to have a lookup-function as well at partial-log.

DB4SCW:
sorry my bad, could have worded that one better...

int2001:
> sorry my bad, could have worded that one better...

no problem. btw: Do you know `ALT-L`? Many ways lead to rome :)

DB4SCW:
Yes, I know Alt+L.
2 "problems" with that:
1) I have to enter the information again, while I just have to click what's already on screen for this one
2) on basic Firefox on Windows (which I use quite a bit with wavelog from a few stations), this one just opens the bookmark menu of firefox without doing anything in wavelog. Works on Edge, Safari and other combinations, but not everywhere.

PR #3808, opened 2026-09-08
Timeline conf

int2001:
Adding a different sorting and displaylogic to the Timeline-Analytics.
Once user choses a confirmation-method, the achievements are ordered by the oldest valid confirmation of the thing to reach (DXCC, VUCC, State, etc.)
Furthermore the confirmationdate is shown. if you hover over it (and select more than once cnf-method), the confirmation-dates of the other methods will also be shown as tooltip

image


phl0 (approved):
Tested and works fine on my system.

PR #3810, opened 2026-09-08
[DCL] Fix upload QSO carry-over across station profiles

dg0tm:
$qso_id_array was declared once outside the station loop, so IDs collected for one station remained in the array when that station's upload failed (HTTP != 200 or curl error) and were carried over into the next station iteration. If that next station then succeeded, mark_dcl_sent() was called for all accumulated IDs, flagging QSOs as COL_DCL_QSL_SENT = 'Y' even though they were never imported on the receiver side.

Because get_dcl_qsos_to_upload() filters out QSOs already marked 'Y', affected QSOs were permanently locked out of future uploads.

Move the declaration inside the loop so the array is per-station and carry-over between profiles is structurally impossible. Drop the now redundant reset at the end of the loop.

### DIFF
--- a/application/controllers/Dcl.php
+++ b/application/controllers/Dcl.php
@@ -105,13 +105,12 @@ class Dcl extends CI_Controller {
 			$sync_user_id=null;
 		}
 
-		// Array of QSO IDs being Uploaded
-
-		$qso_id_array = array();
-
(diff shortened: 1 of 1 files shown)


int2001 (approved):
vy good finding. tnx

PR #3811, opened 2026-09-08
Fix bad-path for LoTW-Uploads

int2001:
Bug:

1. Station A: QSOs 1-10 pushed into array. Upload fails (curl error / signing error) → continue. Array NOT cleared — IDs 1-10 stay                                                                                                           
2. Station B: QSOs 11-20 pushed. Array now 1-20. Upload succeeds → all 20 marked COL_LOTW_QSL_SENT='Y'                                         
3. But QSOs 1-10 never reached ARRL. Wavelog thinks sent

### DIFF
--- a/application/controllers/Lotw.php
+++ b/application/controllers/Lotw.php
@@ -211,14 +211,13 @@ class Lotw extends CI_Controller {
 			$sync_user_id=null;
 		}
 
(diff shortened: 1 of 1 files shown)


phl0 (approved):
Fixes the issue that also failed uploads are marked as QSL sent.

PR #3812, opened 2026-09-08
[Timeline] Fixes tooltip css, and layout css

AndreasK79:
The tooltip for confirmations was all blacked out in the dark themes (Cyborg, Darkly, Superhero).
I also noticed that the non-wide themes wrapped labels, even though we have space.
Labels also had the wrong classes.

The form layout now uses available space:
Before:
image

After:
image


phl0 (approved):
Could repro and this fixes the issue.

PR #3814, opened 2026-09-09
Parallelize QRZ-Upload (concurrency = 5)

int2001:
Besides the fact, that QRZ is only a callbook and has bugs handling SAT-QSOs, etc., there's another major gap in their interface:

User can set valid ranges (From / Till) in qrz itself. But when a client tries to upload a QSO which is out of that range, QRZ answers with `add_qso: outside date range`.

There's absolutely no way to check (via API) whats the valid range for a given Call.
A client needs to push QSO by QSO, to see if it's accepted or not.
The error itself (`add_qso: outside date range`) is already handled in WL. The QSO gets a mark "invalid" for QRZ-Upload.

But: If a user uploads - e.g. 100k QSOs - WL needs to iterate through that 100k QSOs.

The PR here addresses that. Parallelizing requests to QRZ, to milden it a bit. Without the PR it'd take multiple sync-cycles to process those QSOs. Side-Effect: All other stations/users have to wait till the broken QSOs are processed.

PS: Parallelizing is already implemented (and working) for about 1y with LoTW in the same way.

### DIFF
--- a/application/controllers/Qrz.php
+++ b/application/controllers/Qrz.php
@@ -116,7 +116,7 @@ class Qrz extends CI_Controller {
 
 	/*
 	 * Function gets all QSOs from given station_id, that are not previously uploaded to qrz.
-	 * Adif is build for each qso, and then uploaded, one at a time
+	 * Adif is build for each qso, and then uploaded in parallel (curl_multi, sliding window)
 	 */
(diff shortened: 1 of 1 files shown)


HB9HIL (approved):
lgtm

PR #3815, opened 2026-09-09
[Search] extend one-click option to search for all QSOs with specific DXCC

DB4SCW:
extends #3806 

Some DXCCs like Franz-Josef-Land do not render a DXCC flag and thus were not clickable. This makes the DXCC text itself clickable in addition to the flag.


int2001 (approved):
Still don't get real-usecase, since it's visible during logging and with ALT-L.
But doesn't hurt

PR #3819, opened 2026-09-10
[LotW] fix LotW cert status check when ARRL is not reachable

DB4SCW:
When a user has LotW certificates with serial numbers and at the same time ARRL is down (like at the moment, where it responds with HTTP 500 after a lot of waiting), the LotW certificate status check crashes wavelog when the user opens the LotW dashboard page.

This happens because the lotw_cert_status function never checks the HTTP status code and tries parsing the non-xml response.

This PR fixes that so (after a lot of waiting until HTTP 500 arrives), the dashboard opens again without crashing.

It might be worth a second PR to move this function to just right before the upload/download instead of checking during the opening of the page, potentially locking wavelog for that user for minutes at a time (wait time till failure multiplies with certificate count)

### DIFF
--- a/application/controllers/Lotw.php
+++ b/application/controllers/Lotw.php
@@ -1313,8 +1313,13 @@ class Lotw extends CI_Controller {
 			curl_setopt($ch, CURLOPT_RETURNTRANSFER, true);
 			curl_setopt($ch, CURLOPT_CONNECTTIMEOUT, 10);
 			$result = curl_exec($ch);
-			if(curl_errno($ch)){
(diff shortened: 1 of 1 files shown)


int2001 (approved):
good one. i like

phl0 (approved):
Works though having 10 or more certs still takes ages to load. Gonna look for another fix after merging this.

phl0:
> Works though having 10 or more certs still takes ages to load. Gonna look for another fix after merging this.

Follow-up on this: https://github.com/wavelog/wavelog/pull/3820

PR #3822, opened 2026-09-11
Fix Leakage for settings

int2001:
Bug / Repro for settings, which are leaked through sessions/impersonate:

Setup: Two accounts with different dashboard preferences, e.g. user A has "Active Expeditions" enabled, user B has it disabledn.

Impersonation:
1. Log in as A (or as admin who impersonates).                                                                                                                                                          2. Open dashboard — A's layout shows (expeditions visible).
3. Impersonate B (admin user list → impersonate).                                                                                                                                                        4. Bug: Dashboard still shows A's layout (expeditions visible), although B's stored preference is "off". B's DB rows in user_options were never read —  update_session() for the target user kept the source user's session values because of array_key_exists

Admin session pollution:
1. Log in as admin, note own dashboard prefs.
2. Edit another user via User edit form, change their dashboard prefs, save.
3. Bug: Admin's own dashboard now reflects the edited user's prefs until next login — the edit path mirrors posted values into the session unconditionally (User_model.php edit path), and "session wins" kept them.

After fix: update_session() always reads these prefs from user_options (DB), same pattern as user_stylesheet. Session is refreshed from DB on every request via validate_session(), so preferences always follow the currently active user.

For those who are afraid of "expensive SQLs": no additional SQL was added for this fix.


AndreasK79:
Could repro with different settings.

DEV: bleeds over,
PR: works.

AndreasK79 (approved):

PR #3823, opened 2026-09-11
Option to make hams.at posts from SAT passes page

phl0:
This aims to add a new option to post activations to hams.at from sat passes list. Stuff todo:

- [x] Handle setting of frequencies for sats with multiple tpx'es (e.g. ISS)
- [x] Verify hams.at API key on storage
- [x] Better placement of buttons in popup modal
- [ ] ~~Fix vertical alignment of text and input boxes~~

And maybe more. Sneak peak:

Screenshot from 2026-09-11 21-08-04

### DIFF
--- a/application/models/Stations.php
+++ b/application/models/Stations.php
@@ -656,6 +656,25 @@ class Stations extends CI_Model {
 		}
 		return false;
 	}
+
+	public function get_station_refs($stationid) {
+		$sql = "SELECT `station_iota`, `station_sota`, `station_sig`, `station_sig_info`, `station_wwff`, `station_pota` FROM `station_profile` WHERE station_id = ?;";
+		$query = $this->db->query($sql, $stationid);
+		$row = $query->row();
+		if($query->num_rows() >= 1) {
+			$row = $query->row(); // only one result expected
+			return [
+				'iota' => $row->station_iota,
+				'sota' => $row->station_sota,
+				'sig' => $row->station_sig,
+				'sig_info' => $row->station_sig_info,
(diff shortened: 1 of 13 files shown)


application/models/Stations.php
```diff
@@ -656,6 +656,25 @@ public function lookupProfileCoords($stationid) {
 		}
 		return false;
 	}
+
+	public function get_station_refs($stationid) {
+		$sql = "SELECT `station_iota`, `station_sota`, `station_sig`, `station_sig_info`, `station_wwff`, `station_pota` FROM `station_profile` WHERE station_id = ?;";
+		$query = $this->db->query($sql, $stationid);
+		$row = $query->row();
```
int2001: what is it doing?

assets/js/sections/satpasses.js
```diff
@@ -247,3 +286,12 @@ function loadPassSettingsList() {
         }
     });
 }
+
+function toggleRxTx(dir) {
+   console.log("toggleRxTx");
```
int2001: debug leftover

application/models/Stations.php
phl0: Fetches the references from the currently active station location and would set is as draft in comment field.

phl0: image

-> prefilled from station location

int2001: roger. but compare to L665

phl0: Fixed

int2001 (approved):
works. vy cool feature :)
Don't 4get to rm the testflag

PR #3826, opened 2026-09-12
[Contesting] fix clear call on qsy and border

AndreasK79:
This is a minor one. Didn't need to remove border, so it is added back.

Fixes "Clear call on QSY". This wasn't in my dev with the last PR, so missed it.


int2001 (approved):
Tnx!

PR #3830, opened 2026-09-12
always check if cache driver is loaded before loading

kc9uhi:
Cron Manager firing every minute triggers many `Cache class already loaded. Second attempt ignored.` DEBUG messages in application log.

This PR checks if cache driver is already loaded before loading.

(Whitespace auto-edits courtesy of VSCode and .editorconfig)


phl0 (approved):
Tested. Works 👍🏼

PR #3832, opened 2026-09-13
fix broken qso count

HB9HIL:
fixes the qso table in qso logging which was broken since commit https://github.com/wavelog/wavelog/pull/3822/changes/cfdf6ac1b229179ff46bd653775260e351961aa7

This bug only affects new users (fresh install oder newly added) which NEVER hit save in the user account settings which would fill in the missing values.


int2001 (approved):
Args! Sri

PR #3844, opened 2026-09-15
Add JCG (Japan Century Gun/County) award

JF9SOM:
Follow-up to #3839.

## Summary

Adds the JCG (Japan Century Gun/County) award, JCC's companion award covering Japan's *gun* (郡, rural county-equivalent) entities instead of cities. `assets/json/japan_award/jcg_list.json` has been in the tree since `2c0e4175` (2026-03-28) but was never wired up to anything — this does that, mirroring the JCC feature.

- `Jcg_model.php` (parallel to `Jcc_model.php`) — no designated-city-ward mapping needed (guns have no sub-entities), so entity matching is a direct `COL_CNTY` comparison. Kept as its own model per the discussion (my read is the two are similar enough to share logic in principle, but not so identical that I wanted to force a shared base uninvited — happy to revisit if a maintainer feels strongly).
- `Awards::jcg()` / `jcg_export()` / `jcg_map()`, mirroring the `jcc_*` actions.
- `awards/jcg/index.php` view + `jcg.js` / `jcgmap.js`, mirroring the JCC results grid / map / CSV export.
- A `JCG` case in `Logbook_model::qso_details()` for the QSO drill-down list.
- A `jcg` toggle column on the band settings page (`bandxuser.jcg`), alongside the existing `jcc` one — model/controller/view/JS, plus a migration. The band-settings master-checkbox JS already derives the award name from the row's CSS class, so no JS change was needed for that part.
- Menu entry next to JCC.

On my own log, this already surfaces 37 previously-untracked worked guns.

## Test plan

- Verified locally (results grid, map, CSV export, band-settings toggle) against a Docker dev instance with real QSO data.
- `php -l` on all touched PHP files, `node --check` on all touched JS files.
- Migration applies cleanly and is idempotent (`field_exists` guard), matching the existing `183_jcc_bandxuser.php` pattern.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


application/views/interface_assets/header.php
```diff
@@ -271,6 +271,7 @@
 										<li><a class="dropdown-item" href="<?php echo site_url('awards/waja'); ?>"><i class="fas fa-trophy"></i> <?= __("WAJA"); ?></a></li>
 										<li><hr class="dropdown-divider"></li>
 										<li><a class="dropdown-item" href="<?php echo site_url('awards/jcc'); ?>"><i class="fas fa-trophy"></i> <?= __("JCC"); ?></a></li>
+										<li><a class="dropdown-item" href="<?php echo site_url('awards/jcg'); ?>"><i class="fas fa-trophy"></i> <?= __("JCG"); ?></a></li>
```
phl0, line 275:
```suggestion
                                        <li><hr class="dropdown-divider"></li>
										<li><a class="dropdown-item" href="<?php echo site_url('awards/jcg'); ?>"><i class="fas fa-trophy"></i> <?= __("JCG"); ?></a></li>
```

phl0: Please add a divider line in the sub menu here. See suggestion

Image

assets/js/sections/jcgmap.js
```diff
@@ -0,0 +1,209 @@
+let osmUrl = tileUrl;
```
phl0:
```suggestion
var osmUrl = tileUrl;
```

phl0: JS error: `Uncaught SyntaxError: redeclaration of non-configurable global property osmUrl.` See suggestion

phl0 (changes requested):
Found two issues. Please see suggestions.

JF9SOM:
Thanks for the review! Both fixed in 83f33f02f5:

- Added the missing divider between JCC and JCG in the menu.
- `jcgmap.js` now uses `var osmUrl` to match `jccmap.js` (the `let` redeclaration was indeed clashing when both files are loaded on the same page).

phl0 (approved):
This works fine now. Tnx for fixing the issues.

PR #3845, opened 2026-09-15
Fix: Enter key not working in the clubstation personal-callsign modal

Patrik-OM1EPT:
The root cause was a timing (DOM-binding) issue. The old code looked for the input box the moment the page loaded, but the modal didn't yet exist, so it found nothing to attach to. The fix listens on the whole page instead, so it still works even if that box shows up later.


HB9HIL (changes requested):
Good find :) But can you please remove the AI comments? They are not necessary here 😀 such explainaings can live in the third line of the commit messages.

Patrik-OM1EPT:
yeaaa my bad sorry :>
This one I admit I used AI for, because my knowledge of JS isn't that deep to be honest.
I mostly code in C / C#.

Patrik-OM1EPT:
ok should be fine now. Thanks for the comment :)

phl0 (approved):

HB9HIL (approved):

PR #3846, opened 2026-09-15
Fix: QSO map locator jumping when clicking the gridsquares button in legend on the map.

Patrik-OM1EPT:
Two bugs existed and both were making the pin jump around.
1. Clicking the gridsquares checkbox: the map thought you clicked on the map itself, so it moved your pin to that spot instead of just toggling the checkbox.
2. Searching a gridsquare always reset the zoom to a fixed level, so if you'd zoomed in, it snapped back out every time.

hope this helps, because some members of our club complained about this interesting bug :D


phl0 (approved):
Thats a good find. Tnx

Patrik-OM1EPT:
> Thats a good find. Tnx

No problem, I am just happy to be able to help with such an awesome project and also learn PHP along the way. Like this is much better than anything they can teach you at schools or various courses.

Good work guys keep it up :D

PR #3847, opened 2026-09-15
Add per-club direct login toggle for clubstations.

Patrik-OM1EPT:
A simple feature in which I added a per-clubstation "Direct Login" toggle on the Club Permissions page, letting admins allow members to log in directly with the club's own credentials (with a configurable granted permission level) instead of switching from a personal account. This can be handy for clubs with too many people and it simplifies the process for operators.


HB9HIL:
Tbh. I'm not happy with this. Some things including the behaviour of the club station stuff were carefully handcrafted decisions by the team.

This PR throws a lot of them over board.

HB9HIL:
An example of my comment above:

It's on purpose that `club_direct` never was in the config.sample.php. It's just documented in the docs because people should only use it if they really need to.

Patrik-OM1EPT:
Well we mainly use the direct login, because at our club we have a sort of open policy. There is a lot of people moving around, and maintaining 50 accounts is a mess. Additionally, switching to a club with the menu is an extra step that some less IT knowledgeable operators may forget, and this would mean that they will start logging club contacts into their own logbook. My vision was to just click login, and then it will instantly prompt them for a callsign and then they can immediately start logging contacts. 
On the other hand, this kind of "advanced" feature should be hidden away and only admins that really need it should use it.  But if your, as the wavelog team, have a different vision, then I'm going to respect that and you can then close this PR. No problem there :)

int2001:
three options:
- either enable it in config.php
- or: let the people direct login to the account (shared credentials).
- or: use WL clubfeature as designed (Users which can switch to Club)

Hint: Via APIv2 you could even add and grant your 50 Users without hazzle.

tbh: Your PR would break more, than it'd help. Sorry

Patrik-OM1EPT:
Yeaa that's true. No problem, thanks for the comment :)

PR #3848, opened 2026-09-15
QSO edit/delete permissions between operators of a club station.

Patrik-OM1EPT:
I have added a per-club "Cross-Operator Access" toggle letting non-Officer members (Club Member / Club Member ADIF) edit or delete QSOs logged by other operators of the same club, instead of only their own. This is a big feature for our club, because QSOs made with within MSHV or WSJT-X are logged with the operator as OM3KFF (our club) not the operator. So if the program makes a mistake, or if a QSO is logged in by mistake from such programs, it could only be deleted by an admin, well not any more. Same applies for editing QSOs.


HB9HIL (commented):
I don't like this. We added security so Ops can't mess up QSOs of other operators. Why should we add a toggle now to disable this which needs to be maintained and secured in the future...

Patrik-OM1EPT:
I am going to answer your question with another question :D
Again, this is related to my other comment in another PR. Our club has a sort of open policy. And I think that in clubs, the logbook is a shared responsibility between all of the club's operators and I think that other operators wouldn't intentionally sabotage their own logbook by deleting QSOs.
Let's say, (as i alluded to in my original PR message), we sometimes do FT8, and the program that is doing and logging the QSOs sometimes makes a mistake. But because these programs don't support adding a operator (and if they would, it's another complicated step for the operator), so the QSO's are logged with the operator being our club's callsign. And the solution to delete or edit these QSOs is to simply login as an admin and do the changes that way. 
What is your opinion on this? Thanks.

int2001:
WL provides a way for that: Club-Officer. With Clubofficer everyone could edit everyones QSOs.
if you don't want to use the clubfeature and everyone should log everything and edit everything: Single User for clubstation with shared credentials.

what you want is less security/safety (Everyone can edit everyones QSOs) and at the same time more security/safety (Everyone should have a personal login - which is doable by the role "Clubofficer")

PR #3850, opened 2026-09-16
Don't reset zoom on click

int2001:
Activationplanner:
when zoomed in to a grid (via search) there's a zoomlevel (which is fine), but when zooming manual and clicking into map the manual zoom is lost.

this one fixes it


phl0 (approved):
Tested. Works.

PR #3852, opened 2026-09-17
Add button to copy share text to clipboard

phl0:
Not sure where but someone requested a "copy to clipboard" function for the "just worked..." text.

image


int2001 (commented):
Post-Approve :)

PR #3859, opened 2026-09-21
Catch error if session keys for callbook cannot be stored in cache

phl0:
If file adapter is used for cache and application/cache is not writable by the web server user qrz.com lookup fails with "invalid username/password" which is totally misleading:

image

The real problem is the missing write permissions. This at least fixes the error message to not mislead the user.

This addresses https://github.com/wavelog/wavelog/issues/3791


application/libraries/Callbook.php
```diff
@@ -164,7 +166,11 @@ private function _qrz($callsign, $fullname) {
 					$this->ci->cache->delete($this->qrz_session_cachekey);
 					return $callbook;
 				}
-				$this->ci->cache->save($this->qrz_session_cachekey, $qrz_session_key, self::QRZ_SESSION_DURATION);
+				if ($this->ci->cache->save($this->qrz_session_cachekey, $qrz_session_key, self::QRZ_SESSION_DURATION)) {
```
int2001: eh, wait. the whole `if`-block fires on `true` (which means: key could be saved).
sure that this is intended here? Same at L185

application/libraries/Callbook.php
```diff
@@ -69,6 +70,7 @@ public function __construct() {
 
 		$this->logbook_not_configured = __("Lookup not configured. Please review configuration.");
 		$this->error_obtaining_sessionkey = __("Error obtaining a session key for callbook. Error: %s");
+		$this->error_caching_sessionkey = __("Generic error storing the session key in cache.");
```
int2001: Nit: mention qrz.com here? would be good, otherwise someone may think it's about WL-Session-key

phl0: > Nit: mention qrz.com here? would be good, otherwise someone may think it's about WL-Session-key

Hmm no because its Callbook.php I added another commit generalizing the user error message and added more details to the log message.

application/libraries/Callbook.php
phl0: > eh, wait. the whole `if`-block fires on `true` (which means: key could be saved). sure that this is intended here? Same at L185

Right. Fixed in https://github.com/wavelog/wavelog/pull/3859/commits/eb461b68ef6349fc85ab4c570f0da44ff2d09f4e.

phl0:
Added a function in https://github.com/wavelog/wavelog/pull/3859/commits/322acb08b2603547f8ba9525f77a5d2dce9c98c2 to validate session key read from cache.

application/libraries/Callbook.php
```diff
@@ -257,11 +305,22 @@ private function _hamqth($callsign) {
...
+						log_message('error', 'Error storing hamqth.com session key in cache');
+						$callbook['error'] = $this->error_caching_sessionkey;
+						return $callbook;
+					}
 				}
 			}
 
-			$callbook = $this->ci->hamqth->search($callsign, $this->ci->cache->get($this->hamqth_session_cachekey));
+			$cached_hamqth_session_key = $this->ci->cache->get($this->hamqth_session_cachekey) ?? '';
+			if ($hamqth_session_cachekey != '') {
```
int2001: L316: `$cached_hamqth_session_key`
L317: `$hamqth_session_cachekey`

application/libraries/Callbook.php
```diff
@@ -271,8 +330,19 @@ private function _hamqth($callsign) {
...
 				}
-				$this->ci->cache->save($this->hamqth_session_cachekey, $hamqth_session_key, self::HAMQTH_SESSION_DURATION);
-				$callbook = $this->ci->hamqth->search($callsign, $this->ci->cache->get($this->hamqth_session_cachekey));
+				if (!$this->ci->cache->save($this->hamqth_session_cachekey, $hamqth_session_key, self::HAMQTH_SESSION_DURATION)) {
+					log_message('error', 'Error storing hamqth.com session key in cache');
+					$callbook['error'] = $this->error_caching_sessionkey;
+					return $callbook;
+				}
+				$cached_hamqth_session_key = $this->ci->cache->get($this->hamqth_session_cachekey) ?? '';
+				if ($hamqth_session_cachekey != '') {
```
int2001: L338: `$cached_hamqth_session_key`
L339: `$hamqth_session_cachekey`

application/libraries/Callbook.php
```diff
@@ -209,14 +235,25 @@ private function _qrzcq($callsign) {
...
 					return $callbook;
 				}
 			}
 
-			$callbook = $this->ci->qrzcq->search($callsign, $this->ci->cache->get($this->qrzcq_session_cachekey));
+			$cached_qrzcq_session_key = $this->ci->cache->get($this->qrzcq_session_cachekey) ?? '';
+			if ($cached_qrzcq_session_key != '') {
+				$callbook = $this->ci->qrzcq->search($callsign, $cached_qrzcq_session_key);
+			} else {
+				log_message('error', 'Error reading cached qrzru session key from cache');
```
int2001: qrzu? this is handler for qrzcq, isn't it?

application/libraries/Callbook.php
```diff
@@ -225,8 +262,19 @@ private function _qrzcq($callsign) {
...
+				if (!$this->ci->cache->save($this->qrzcq_session_cachekey, $qrzcq_session_key[1], self::QRZCQ_SESSION_DURATION)) {
+					log_message('error', 'Error storing qrzcq session key in cache');
+					$callbook['error'] = $this->error_caching_sessionkey;
+					return $callbook;
+				}
+				$cached_qrzcq_session_key = $this->ci->cache->get($this->qrzcq_session_cachekey) ?? '';
+				if ($cached_qrzcq_session_key != '') {
+					$callbook = $this->ci->qrzcq->search($callsign, $cached_qrzcq_session_key);
+				} else {
+					log_message('error', 'Error reading cached qrzru session key from cache');
```
int2001: qrzru? this is handler for qrzcq

application/libraries/Callbook.php
phl0: Fixed in https://github.com/wavelog/wavelog/pull/3859/commits/508890421a1488d3c870a7ac29dddf5cedfcf24c.

application/libraries/Callbook.php
phl0: Fixed in https://github.com/wavelog/wavelog/pull/3859/commits/508890421a1488d3c870a7ac29dddf5cedfcf24c.

application/libraries/Callbook.php
phl0: Fixed in https://github.com/wavelog/wavelog/pull/3859/commits/508890421a1488d3c870a7ac29dddf5cedfcf24c.

application/libraries/Callbook.php
phl0: Fixed in https://github.com/wavelog/wavelog/pull/3859/commits/508890421a1488d3c870a7ac29dddf5cedfcf24c.

int2001 (approved):

PR #3860, opened 2026-09-21
Test cache directory before using it

phl0:
This is a follow-up to https://github.com/wavelog/wavelog/pull/3859. If the cache path of the file adapter is not even readable by the web server user the debug page will throw an exception or even fail with HTTP 500. Test by withdraw all access permissions for the web server user:

```
An uncaught Exception was encountered

Type: UnexpectedValueException

Message: RecursiveDirectoryIterator::__construct(/var/www/wavelog/application/cache/70): Failed to open directory: Permission denied

Filename: /XXX/wavelog/application/models/Debug_model.php

Line Number: 306
```

So if the web server in case of https://github.com/wavelog/wavelog/issues/3791 could not even read the cache directory and the cache adapter is set to `file` the user could not even have shown the debug page.


int2001:
Hmm. it check only if the cache-dir is readable. can't spot a write-check here.
Proposal:
Add `if (!is_writable($cache_path)) {  return 0; }` to it

phl0:
> Hmm. it check only if the cache-dir is readable. can't spot a write-check here. Proposal: Add `if (!is_writable($cache_path)) { return 0; }` to it

Its not about writing but more about if the cache dir is really accessible. But closing here in favor of https://github.com/wavelog/wavelog/pull/3862 in combination with https://github.com/int2001/wavelog/pull/30

